PR #97, opened 2021-10-01
Add solution file for all projects

hishamco:


Piedone:
The UI Testing Toolbox is mean to be used from the [Lombiq's Open-Source Orchard Core Extensions solution](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/), especially because its samples need a web project too (and what you need to do for a web project to work along nicely with the UI Testing Toolbox is demonstrated there). The Readme here and other docs also make a note of this.

Thank you anyway!

hishamco:
I see that yesterday, thanks

hishamco:
One remark I need to point out, dependencies between more than one repo is not a good idea from developer perspective. Why we didn't add an OC sample to this repo

This is a common scenario that I had seen in ASP.NET Core repos (for integration & functional testing)

Piedone:
The reason is that that way we don't have one more solution to maintain :). Also, we use the sample tests of the UI Testing Toolbox to smoke test the other OSOCE projects. But I get your remark about convenience, and we'll solve this with NuGet packages.

PR #99, opened 2021-10-06
Basic Orchard feature tests

YevgeniyShunevych:


Lombiq.Tests.UI/Pages/OrchardCoreSetupParameters.cs
```diff
@@ -9,8 +11,8 @@ public class OrchardCoreSetupParameters
         public OrchardCoreSetupPage.DatabaseType DatabaseProvider { get; set; } = OrchardCoreSetupPage.DatabaseType.Sqlite;
         public string ConnectionString { get; set; }
         public string TablePrefix { get; set; }
-        public string UserName { get; set; } = "admin";
-        public string Email { get; set; } = "admin@admin.com";
-        public string Password { get; set; } = "Password1!";
+        public string UserName { get; set; } = DefaultUser.UserName;
+        public string Email { get; set; } = DefaultUser.UserName + "@admin.com";
+        public string Password { get; set; } = DefaultUser.Password;
```
Piedone: Good point. While we're at it, we could have a const for the Email in `DefaultUser` too.

Lombiq.Tests.UI/Attributes/Behaviors/SetsValueReliablyAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            element.ClearWithLogging();
+
+            if (text.Contains('@', StringComparison.Ordinal))
+            {
+                // On some platforms, probably due to keyboard settings, the @ character can be missing from the address
+                // when entered into a text field so we need to use Actions. The following solution doesn't work:
+                // https://stackoverflow.com/a/52202594/220230. This needs to be done in addition to the standard
+                // FillInWith() as without that some forms start to behave strange and not save values.
+                AtataContext.Current.Driver.Perform(x => x.SendKeys(element, text));
```
Piedone: Use a descriptive variable name instead of `x`.

Lombiq.Tests.UI/Attributes/Behaviors/SetsValueReliablyAttribute.cs
```diff
@@ -0,0 +1,38 @@
+using Atata;
+using Lombiq.Tests.UI.Helpers;
+using OpenQA.Selenium;
+using System;
+
+namespace Lombiq.Tests.UI.Attributes.Behaviors
+{
+    public sealed class SetsValueReliablyAttribute : ValueSetBehaviorAttribute
+    {
+        public override void Execute<TOwner>(IUIComponent<TOwner> component, string value)
```
Piedone, line 9: DRY this logic with the one in `FormUITestContextExtensions`.

Piedone (changes requested):

Lombiq.Tests.UI/Pages/OrchardCoreSetupParameters.cs
YevgeniyShunevych: Added `Email` property.

Lombiq.Tests.UI/Attributes/Behaviors/SetsValueReliablyAttribute.cs
YevgeniyShunevych: There is a problem here. In `SetsValueReliablyAttribute` I don't have an instance of `UITestContext` in order to invoke its extension method.

Lombiq.Tests.UI/Attributes/Behaviors/SetsValueReliablyAttribute.cs
YevgeniyShunevych: Renamed.

Lombiq.Tests.UI/Attributes/Behaviors/SetsValueReliablyAttribute.cs
Piedone: You can just put the logic here into a third method that accepts the driver, and then use that from both places: Here it would get the driver from `AtataContext.Current`, otherwise from `UITestContext`.

YevgeniyShunevych: OK. I was also thinking about that. Will do it.

YevgeniyShunevych: Refactored.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -11,15 +12,22 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
...
                 .TestLogin()
+                .TestTurningFeatureOnAndOff()
                 .TestLogout();
 
         public static UITestContext TestSetup(this UITestContext context, OrchardCoreSetupParameters parameters = null)
         {
-            parameters ??= new OrchardCoreSetupParameters();
+            parameters ??= new OrchardCoreSetupParameters
+            {
+                RecipeId = "Lombiq.OSOCE.BasicOrchardFeaturesTests",
```
Piedone: This can't really be a default as it's very specific to the OSOCE solution. It needs to be supplied from the sample code.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -63,6 +71,23 @@ public static UITestContext TestSetupNegatively(this UITestContext context, Orch
...
+            context.ExecuteTest(
+                "Test login with invalid data",
+                () =>
+                {
+                    context.GoToLoginPage()
+                        .LogInWith(userName, password)
+                        .ShouldStayOnLoginPage()
+                        .ValidationSummaryErrors.Should.Not.BeEmpty();
+
+                    context.GetCurrentUserName().ShouldNotBe(userName);
```
Piedone:
```suggestion
                    context.GetCurrentUserName().ShouldBeEmpty();
```

Lombiq.Tests.UI/Models/UserRegistrationModel.cs
```diff
@@ -0,0 +1,21 @@
+using Lombiq.Tests.UI.Constants;
+
+namespace Lombiq.Tests.UI.Models
+{
+    public class UserRegistrationModel
```
Piedone:
```suggestion
    public class UserRegistrationParameters
```

Lombiq.Tests.UI/Extensions/BasicWebDriverExtensions.cs
```diff
@@ -0,0 +1,29 @@
+using Atata;
+using OpenQA.Selenium;
+using System;
+
+namespace Lombiq.Tests.UI.Extensions
+{
+    public static class BasicWebDriverExtensions
```
Piedone:
```suggestion
    public static class FormWebDriverExtensions
```

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -76,21 +101,78 @@ public static UITestContext TestLogout(this UITestContext context)
...
+                "Test registration",
                 () =>
                 {
                     context.GoToLoginPage()
-                        .LogInWith(userName, password)
-                        .ShouldStayOnLoginPage();
+                        .RegisterAsNewUser.Should.BeVisible();
 
-                    context.GetCurrentUserName().ShouldNotBe(userName);
+                    context.GoToRegistrationPage()
```
Piedone: Let's click the previously checked link instead of going to the registration page directly.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -76,21 +101,78 @@ public static UITestContext TestLogout(this UITestContext context)
...
+        {
+            model ??= UserRegistrationModel.CreateDefault();
+
+            return context.ExecuteTest(
+                "Test registration with already registered email",
+                () =>
+                context.GoToRegistrationPage()
+                    .RegisterWith(model)
+                    .ShouldStayOnRegistrationPage()
+                    .ValidationMessages[x => x.Email].Should.BeVisible());
```
Piedone: Use a descriptive variable name instead of `x`.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -76,21 +101,78 @@ public static UITestContext TestLogout(this UITestContext context)
...
+        }
+
+        public static UITestContext TestTurningFeatureOnAndOff(this UITestContext context, string featureName = "Background Tasks")
+            =>
```
Piedone:
```suggestion
        public static UITestContext TestTurningFeatureOnAndOff(this UITestContext context, string featureName = "Background Tasks") =>
```

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -76,21 +101,78 @@ public static UITestContext TestLogout(this UITestContext context)
...
+            context.ExecuteTest(
+                "Test turning feature on and off",
+                () =>
+                context.GoToFeaturesPage()
+                    .SearchForFeature(featureName).IsEnabled.Get(out bool originalEnabledState)
+                    .Features[featureName].CheckBox.Check()
+                    .BulkActions.Toggle.Click()
+
+                    .AggregateAssert(x => x
```
Piedone: Use a descriptive variable name instead of `x`, also below.

Lombiq.Tests.UI/Pages/OrchardCoreAdminPage.cs
```diff
@@ -22,5 +23,8 @@ protected override void OnVerify()
             base.OnVerify();
             ShouldStayOnAdminPage();
         }
+
+        public TOwner ShouldContainSuccessAlertMessage(TermMatch expectedMatch, string expectedText) =>
+            AlertMessages.Should.Contain(x => x.IsSuccess && expectedMatch.IsMatch(x.Text.Value, expectedText));
```
Piedone: Use a descriptive variable name instead of `x`.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -76,21 +101,78 @@ public static UITestContext TestLogout(this UITestContext context)
...
-                        .LogInWith(userName, password)
-                        .ShouldStayOnLoginPage();
+                        .RegisterAsNewUser.Should.BeVisible();
 
-                    context.GetCurrentUserName().ShouldNotBe(userName);
+                    context.GoToRegistrationPage()
+                        .RegisterWith(model)
+                        .ShouldLeaveRegistrationPage();
+
+                    context.GetCurrentUserName().ShouldBe(model.UserName);
```
Piedone: After this, it would be useful to log out. I'm not actually sure why `TestRegistrationWithAlreadyRegisteredEmail()` works after this but perhaps it's just an oversight that as a logged in user you can fill out the registration page.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,188 @@
...
+            {
+                RecipeId = "Lombiq.OSOCE.BasicOrchardFeaturesTests",
+            };
+
+            return context.ExecuteTest(
+                "Test setup",
+                () =>
+                context.GoToSetupPage()
+                    .SetupOrchardCore(parameters)
+                    .ShouldLeaveSetupPage());
```
Piedone: Similarly in the other cases too:

```suggestion
                () => context
                    .GoToSetupPage()
                    .SetupOrchardCore(parameters)
                    .ShouldLeaveSetupPage());
```

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
YevgeniyShunevych: Currently, how I see Orchard works. Being already logged in, user can go to login page and try to login. If login failed, the user remains still be logged in with previously login data. Current implementation covers this scenario.

So if we change assertion to `ShouldBeEmpty()` and invoke `TestLoginWithInvalidData` when user is logged in, we'll get an assertion exception `"should be empty but was "TestUser""`. So I see few solutions here:
1. Remain current implementation.
1. Use `ShouldBeEmpty()`.
1. Use `ShouldBeEmpty()`, but add logout step to the test beginning.

@Piedone, which one to choose?

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
YevgeniyShunevych: Updated.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
YevgeniyShunevych: Updated the test.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
YevgeniyShunevych: OK. So first of all I will then use `new OrchardCoreSetupParameters()` by default in `TestSetup` method. 

What if we add optional `OrchardCoreSetupParameters` parameter to `TestBasicOrchardFeatures`? Then I can update the test method this way:

```cs
[Theory, Chrome]
public Task BasicOrchardFeaturesShouldWork(Browser browser) =>
    ExecuteTestAsync(
        context => context.TestBasicOrchardFeatures(
            new OrchardCoreSetupParameters
            {
                RecipeId = "Lombiq.OSOCE.BasicOrchardFeaturesTests",
            }),
        browser);
```

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
Piedone: A safety login step before with `SignOutDirectly()` is good.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
Piedone: Looks good!

YevgeniyShunevych: Done,

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
YevgeniyShunevych: Do you mean "A safety **logout** step"? Or something else?

Piedone: Yes, sorry.

YevgeniyShunevych: No problem. Updated the test.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -132,44 +137,65 @@ public static UITestContext TestRegistrationWithInvalidData(this UITestContext c
...
+                    .GoToRegistrationPage()
+                    .RegisterWith(parameters)
                     .ShouldStayOnRegistrationPage()
-                    .ValidationMessages[x => x.Email].Should.BeVisible());
+                    .ValidationMessages[page => page.Email].Should.BeVisible());
         }
 
-        public static UITestContext TestTurningFeatureOnAndOff(this UITestContext context, string featureName = "Background Tasks")
+        public static UITestContext TestContentOperations(this UITestContext context, string pageTitle = "Test page")
             =>
```
Piedone: There are a couple of other such examples in this class too.

```suggestion
        public static UITestContext TestContentOperations(this UITestContext context, string pageTitle = "Test page") =>
```

Lombiq.Tests.UI/Attributes/Behaviors/SetsValueReliablyAttribute.cs
```diff
@@ -0,0 +1,18 @@
...
+using Lombiq.Tests.UI.Helpers;
+
+namespace Lombiq.Tests.UI.Attributes.Behaviors
+{
+    public sealed class SetsValueReliablyAttribute : ValueSetBehaviorAttribute
+    {
+        public override void Execute<TOwner>(IUIComponent<TOwner> component, string value)
+        {
+            var element = component.Scope;
+            var driver = AtataContext.Current.Driver;
```
Piedone: There are a couple of such `AtataContext.Current` references. We've consciously avoided this until now due to the references possibly being lost when `async` code is in the tests, especially if multiple ones are executed in parallel. Hence `UITestContext.Scope.AtataContext` too. Please use `UITestContext` instead too.

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,214 @@
...
+using Lombiq.Tests.UI.Pages;
+using Lombiq.Tests.UI.Services;
+using Shouldly;
+using System;
+
+namespace Lombiq.Tests.UI.Extensions
+{
+    public static class BasicOrchardFeaturesTestingUITestContextExtensions
+    {
+        public static UITestContext TestBasicOrchardFeatures(this UITestContext context, OrchardCoreSetupParameters setupParameters = null) =>
```
Piedone, line 28: Add some docs on the public methods, please, especially outlining any assumptions they make, like the site not being set up.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
YevgeniyShunevych: Updated.

Lombiq.Tests.UI/Attributes/Behaviors/SetsValueReliablyAttribute.cs
YevgeniyShunevych: I thought that `AtataContext.Current` should not be such issue as before, as there was improvement on that. Anyway, `AtataContext.Current` is used in Atata itself in some places.

I would like to use `UITestContext` here but I can't get a reference to current instance in behavior attribute.

Either way, I have a suggestion. I can make [Driver property of UIComponent](https://github.com/atata-framework/atata/blob/bf4cbd0dd2974c2b5112f3ee6107e129d154bccb/src/Atata/Components/UIComponent.cs#L31) public and add it to `IUIComponent<TOwner>`. Then instead of using `AtataContext.Current.Driver` we'll be able to use `component.Driver`. I'll just need a couple of days to release Atata v1.13.0 with this change, as there are few other things that I want to add to upcoming Atata version release.

Piedone: Even after the improvements we had issues, hence the `GoToPage<T>()` extension to avoid using `AtataContext.Current`.

While using `component.Driver` would work here, we also have `AtataContext.Current.BaseUrl` references. Why not just add a `UITestContext` property onto the page classes that you can then access?

YevgeniyShunevych: I thought and reviewed Atata a bit. I am thinking to add `public AtataContext Context { get; }` property to `UIComponent`. This way we'll have and be able to use `AtataContext` instance in page objects, in attributes, etc.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
YevgeniyShunevych: I've added XML comments. Please check it out.

Lombiq.Tests.UI/Attributes/Behaviors/SetsValueReliablyAttribute.cs
Piedone: Would that mean that after `UITestContext.Scope.AtataContext.Go.To<MyPage>()` you'd be able to access the same `UITestContext.Scope.AtataContext` (and not `AtataContext.Current`) from inside `MyPage`? If yes then great!

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -8,8 +8,25 @@
...
+    /// Provides a set of extension methods for basic Orchard features testing.
+    /// </summary>
     public static class BasicOrchardFeaturesTestingUITestContextExtensions
     {
+        /// <summary>
+        /// <para>
+        /// Tests all the basic Orchard features.
+        /// At first sets up Orchard with optionally specified <paramref name="setupParameters"/>.
+        /// By default uses new <see cref="OrchardCoreSetupParameters"/> instance
+        /// with <c>"SaaS"</c> <see cref="OrchardCoreSetupParameters.RecipeId"/> value.
```
Piedone: If you want multiple paragraphs here (it seems to me based on the line breaks) then add multiple `<para>` tags, please. Otherwise, break comments at 120 characters (i.e. the first soft limit also added by the Analyzers). This extension can do this automatically: https://marketplace.visualstudio.com/items?itemName=stkb.rewrap

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -8,8 +8,25 @@
...
+        /// </para>
+        /// <para>
+        /// The test method assumes that site is not set up.
```
Piedone: Also below:

```suggestion
        /// The test method assumes that the site is not set up.
```

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -8,8 +8,25 @@
...
+        /// </para>
+        /// </summary>
+        /// <param name="context">The context.</param>
```
Piedone: No need to add a comment for the `this` parameter as you never have to explicitly provide it.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -35,6 +65,21 @@ public static UITestContext TestSetup(this UITestContext context, OrchardCoreSet
                     .ShouldLeaveSetupPage());
         }
 
+        /// <summary>
+        /// <para>
+        /// Tests negatively the site setup with optionally set <paramref name="parameters"/>.
```
Piedone:
```suggestion
        /// Tests the site setup negatively with optionally set <paramref name="parameters"/>.
```

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -101,6 +177,19 @@ public static UITestContext TestLogout(this UITestContext context)
                     context.GetCurrentUserName().ShouldBeNullOrEmpty();
                 });
 
+        /// <summary>
+        /// <para>
+        /// Tests the user registration with optionally specified <paramref name="parameters"/>.
+        /// After the user is registered, the test performs login with the user credentials, then logout.
+        /// </para>
+        /// <para>
+        /// The test method assumes that "Users Registration" Orchard feature is enabled
```
Piedone:
```suggestion
        /// The test method assumes that the "Users Registration" Orchard feature is enabled
```

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -101,6 +177,19 @@ public static UITestContext TestLogout(this UITestContext context)
...
+        /// and there is no registered user with the given <see cref="UserRegistrationParameters.Email"/> value.
```
Piedone: Should be email or username values, or isn't this how it works?

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -125,6 +214,19 @@ public static UITestContext TestRegistration(this UITestContext context, UserReg
                 });
         }
 
+        /// <summary>
+        /// <para>
+        /// Tests negatively the user registration with optionally specified invalid <paramref name="parameters"/>.
```
Piedone:
```suggestion
        /// Tests the user registration negatively with optionally specified invalid <paramref name="parameters"/>.
```

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -125,6 +214,19 @@ public static UITestContext TestRegistration(this UITestContext context, UserReg
...
+        /// Fills user registration fields with <paramref name="parameters"/> on registration page,
+        /// clicks "Register" button and verifies that there are validation messages on the page.
+        /// </para>
+        /// <para>
+        /// The test method assumes that "Users Registration" Orchard feature is enabled.
```
Piedone:
```suggestion
        /// The test method assumes that the "Users Registration" Orchard feature is enabled.
```

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -144,6 +246,21 @@ public static UITestContext TestRegistrationWithInvalidData(this UITestContext c
                     .ValidationMessages.Should.Not.BeEmpty());
         }
 
+        /// <summary>
+        /// <para>
+        /// Tests negatively the user registration with optionally specified <paramref name="parameters"/>
```
Piedone:
```suggestion
        /// Tests the user registration negatively with optionally specified <paramref name="parameters"/>
```

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -144,6 +246,21 @@ public static UITestContext TestRegistrationWithInvalidData(this UITestContext c
...
+        /// that uses email of the already registered user.
+        /// Fills user registration fields with <paramref name="parameters"/> on registration page,
+        /// clicks "Register" button and verifies that there is a validation message near "Email" field on the page.
+        /// </para>
+        /// <para>
+        /// The test method assumes that "Users Registration" Orchard feature is enabled
```
Piedone:
```suggestion
        /// The test method assumes that the "Users Registration" Orchard feature is enabled
```

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -144,6 +246,21 @@ public static UITestContext TestRegistrationWithInvalidData(this UITestContext c
...
+        /// and there is already registered user with the given <see cref="UserRegistrationParameters.Email"/> value.
```
Piedone:
```suggestion
        /// and there is an already registered user with the given <see cref="UserRegistrationParameters.Email"/> value.
```

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -159,8 +276,26 @@ public static UITestContext TestRegistrationWithInvalidData(this UITestContext c
...
         }
 
-        public static UITestContext TestContentOperations(this UITestContext context, string pageTitle = "Test page")
-            =>
+        /// <summary>
+        /// <para>
+        /// Tests content operations. The test executes the following steps:
+        /// </para>
+        /// <list type="number">
+        /// <item><description>Navigate to "Content / Content Items" page.</description></item>
```
Piedone:
```suggestion
        /// <item><description>Navigate to the "Content / Content Items" page.</description></item>
```

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -179,6 +314,26 @@ public static UITestContext TestContentOperations(this UITestContext context, st
                         .CloseWindow();
                 });
 
+        /// <summary>
+        /// <para>
+        /// Tests turning feature on and off. The test executes the following steps:
+        /// </para>
+        /// <list type="number">
+        /// <item><description>Navigate to "Configuration / Features" page.</description></item>
```
Piedone:
```suggestion
        /// <item><description>Navigate to the "Configuration / Features" page.</description></item>
```

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
YevgeniyShunevych: OK, reformatted.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
YevgeniyShunevych: OK, removed. I just get used to have a StyleCop warning in Atata for missing parameter comment.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
YevgeniyShunevych: You are right. I've updated the comment.

Lombiq.Tests.UI/Attributes/Behaviors/SetsValueReliablyAttribute.cs
YevgeniyShunevych: Basically, yes.

Piedone: Sounds good then.

YevgeniyShunevych: I've released Atata v1.13.0 and updated `AtataContext` related code here.

PR #100, opened 2021-10-08
FIN-440: Migrate AddPageMarker and add ClickPublishReliably.

sarahelsaig:
FIN-440

Relevant Commits:
- https://github.com/Lombiq/UI-Testing-Toolbox/pull/100/commits/91d38fbdb955674c3e7054b14803ed287a7a1e92
- [`af592c3` - `78f0052`](https://github.com/Lombiq/UI-Testing-Toolbox/pull/100/files/af592c3a8917bd28bd8397a8e476601cc08553cc..78f0052896f33de434f66070e41c78ac3b8cc738)


Lombiq.Tests.UI/Extensions/VerificationUITestContextExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace Lombiq.Tests.UI.Extensions
+{
+    public static class VerificationUITestContextExtensions
+    {
+        /// <summary>
+        /// Adds a unique empty element to the page. It can be used to indicate if there has been any navigation since
+        /// this method has been called.
+        /// </summary>
+        /// <returns>The query which indicates this unique element alone.</returns>
+        public static By AddPageMarker(this UITestContext context)
```
Piedone: The `<html>` element should serve as such too. See `EventsOrchardCoreUITestExecutorConfigurationExtensions` (in `dev`).

sarahelsaig: That would involve raising a stale elelment reference exception, right? With this you don't have to so it's a bit cleaner. (eg. using `context.Missing(byMarker.Safely()`)

Piedone: Yes, though I'm not sure if manipulating the DOM is universally safe. You can do a try though and see if using this in `EventsOrchardCoreUITestExecutorConfigurationExtensions` too works.

Piedone:
Any news? When will this be completed?

wAsnk (approved):

Lombiq.Tests.UI/Extensions/VerificationUITestContextExtensions.cs
Piedone: This wasn't resolved @wAsnk and still needs to be addressed.

sarahelsaig: How so? I have completely removed `AddPageMarker` and replaced it with the `PageNavigationState` class that uses the `<html>` element as you suggested. See this commit: https://github.com/Lombiq/UI-Testing-Toolbox/pull/100/commits/6385cdefc47029431999b890e14384c76b06413b That's the reason why I ported back `dev` branch into this branch.

On my end, wrapping the try-catch into a bool returning method (`PageNavigationState.CheckIfNavigationHasOccurred`) resolves my objections, as the framework complexity is abstracted away from the library user.

Piedone: Great then!

PR #102, opened 2021-10-21
OSOE-46: Use WebApplicationFactory directly instead of running the tested app with the dotnet CLI

### LINKED ISSUE #101: Use WebApplicationFactory directly instead of running the tested app with the dotnet CLI (OSOE-46)
You can run an ASP.NET Core app directly with `WebApplicationFactory`, see [this blog post](https://www.benday.com/2021/07/19/asp-net-core-integration-tests-with-selenium-webapplicationfactory/) and [the corresponding code](https://github.com/benday-inc/SeleniumDemo). Why?

- This can potentially make things faster.
- Reduce the app-level configuration you have to do, so no need to add the Shortcuts module, for example, or `ConfigureUITesting()`.
- Since this would require a hard reference from the UI test project and the web app, we'd get rid of the necessity to manually build the web app.
- This would allow us to debug the app being tested too, since it would be the same as the test process.
- After this, [this note](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/Docs/Troubleshooting.md#:~:text=Run%20tests%20with,running%20the%20app.) should be adjusted.

See the linked PR for the beginnings of this and comments.

Note that in the linked demo stylesheets didn't load so that may be something to resolve.

This needs to happen after a .NET 6 upgrade (https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/issues/30).

Piedone:
OSOE-46
Fixes #101


Piedone:
@dministro open a new PR, please.

PR #105, opened 2021-11-30
FINI-859, OSOE-50: Reliability extensions, setup screen size, orchard-core-preview

sarahelsaig:
[FINI-859](https://lombiq.atlassian.net/browse/FINI-859)
[OSOE-50](https://lombiq.atlassian.net/browse/OSOE-50)


Lombiq.Tests.UI/Docs/TestableOrchardCoreApps.md
```diff
@@ -47,13 +47,13 @@ Tips on making specific features testable are under the ["Creating tests" page](
     ]
     ```
     - For external web APIs you can implement mock API services in features only enabled in tests. Those features you can again enable in a test recipe. An alternative is to use a tool that provides fake APIs like [JSON Server](https://github.com/typicode/json-server) or [Fake JSON Server](https://github.com/ttu/dotnet-fake-json-server). You can run such tools from the command line in the test's code, e.g. with the excellent [CliWrap](https://github.com/Tyrrrz/CliWrap) that the UI Testing Toolbox uses too.
-- By default, the culture settings used when setting up an Orchard site depend on the host machine's culture. You want to make these settings consistent across all environments though so e.g. datetime and number formatting will be consistent. You can do this by enabling `OrchardCore.Localization` and configuring the culture in site settings from the setup recipe:
```
Piedone: Enabling `OrchardCore.Localization` is a necessary step.

Lombiq.Tests.UI/Exceptions/HtmlValidationAssertionException.cs
```diff
@@ -30,5 +35,17 @@ public HtmlValidationAssertionException(string message, Exception innerException
...
+        {
+            var url = context.Driver.Url;
+            var title = context.Driver.Title ?? url;
+
+            var message = $"Asserting the HTML validation result on page {url}({title}) failed.";
+
+            return createReportOnFailure
+                ? message + " Check the HTML validation report in the failure dump for details."
+                : message;
+        }
```
Piedone: This is quite cool but should be handled more generically at a higher level, so you get this contextual information for any exception that's thrown, including axe violations (which are asserted during page changes too) or else.

Lombiq.Tests.UI/Extensions/EmailUITestContextExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+using Lombiq.Tests.UI.Helpers;
+using Lombiq.Tests.UI.Services;
+using OpenQA.Selenium;
+using System;
+
+namespace Lombiq.Tests.UI.Extensions
+{
+    public static class EmailUITestContextExtensions
+    {
+        public static IWebElement FindSpecificEmailInInbox(
```
Piedone, line 34: Add documentation.

Lombiq.Tests.UI/Extensions/EmailUITestContextExtensions.cs
```diff
@@ -0,0 +1,33 @@
+using Atata;
+using Lombiq.Tests.UI.Helpers;
+using Lombiq.Tests.UI.Services;
+using OpenQA.Selenium;
+using System;
+
+namespace Lombiq.Tests.UI.Extensions
+{
+    public static class EmailUITestContextExtensions
```
Piedone, line 9: I'd now move the rest of the SMTP-related extensions here too.

Lombiq.Tests.UI/Extensions/EventsOrchardCoreUITestExecutorConfigurationExtensions.cs
```diff
@@ -27,38 +27,29 @@ static bool IsNoAlert(UITestContext context)
                 }
             }
 
-            IWebElement html = null;
+            PageNavigationState navigationState = null;
```
Piedone, line 16: So the logic here is the same just factored out, right?

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -210,6 +210,64 @@ public static void SetDropdownByText(this UITestContext context, By selectBy, st
         public static void ClickReliablyOnSubmit(this UITestContext context) =>
             context.ClickReliablyOn(By.CssSelector("button[type='submit']"));
 
+        /// <summary>
+        /// Finds the "Add New" button.
+        /// </summary>
+        public static IWebElement GetAddNewButton(this UITestContext context) =>
+            context.Get(By.XPath("//button[contains(.,'Add New')]"));
+
+        public static void SelectAddNewDropdown(this UITestContext context, By byLocalMenuItem = null) =>
```
Piedone, line 210: Docs, please.

Lombiq.Tests.UI/Extensions/OrchardCoreDashboardUITestContextExtensions.cs
```diff
@@ -15,26 +16,47 @@ public static void ClickPublish(this UITestContext context, bool withJavaScript
...
 
-        public static void GoToContentItemList(this UITestContext context)
+        public static void ClickModalOk(this UITestContext context) => context.ClickReliablyOn(By.Id("modalOkButton"));
+
+        /// <summary>
+        /// Sometimes the Publish button doesn't get clicked. This method retries pressing it up to 4 times with a 30
+        /// second interval between attempts. This should grant enough time to execute the publish action if the button
+        /// actually got pressed.
+        /// </summary>
+        public static void ClickPublishReliably(this UITestContext context, bool withJavaScript = false)
```
Piedone: Why would we not use this every time?

Lombiq.Tests.UI/Extensions/OrchardCoreDashboardUITestContextExtensions.cs
```diff
@@ -15,26 +16,47 @@ public static void ClickPublish(this UITestContext context, bool withJavaScript
...
+            var navigationState = context.AsPageNavigationState();
+
+            context.DoWithRetriesOrFail(
+                () =>
+                {
+                    ClickPublish(context, withJavaScript);
+                    return navigationState.CheckIfNavigationHasOccurred();
+                },
+                timeout: TimeSpan.FromSeconds(30),
+                interval: TimeSpan.FromMinutes(2));
```
Piedone: These should rather default to the values coming from the config (like `DoWithRetriesOrFail()` does) or at least be possible to override.

Lombiq.Tests.UI/Extensions/OrchardCoreDashboardUITestContextExtensions.cs
```diff
@@ -15,26 +16,47 @@ public static void ClickPublish(this UITestContext context, bool withJavaScript
...
+                context.ExecuteScript("document.querySelector('.publish-button, .publish.btn').click();");
             }
             else
             {
                 context.ClickReliablyOn(By.Name("submit.Publish"));
             }
         }
 
-        public static void GoToContentItemList(this UITestContext context)
+        public static void ClickModalOk(this UITestContext context) => context.ClickReliablyOn(By.Id("modalOkButton"));
```
Piedone, line 30: image

Lombiq.Tests.UI/Pages/OrchardCoreSetupPage.cs
```diff
@@ -68,20 +70,22 @@ public enum DatabaseType
...
         public _ SetupOrchardCore(UITestContext context, OrchardCoreSetupParameters parameters = null) =>
             SetupOrchardCore(parameters);
 
         public _ SetupOrchardCore(OrchardCoreSetupParameters parameters = null)
         {
             parameters ??= new OrchardCoreSetupParameters();
 
+            // Set the setup page window size to avoid the "OpenQA.Selenium.WebDriverException: move target out of
+            // bounds" that may happen in headless mode due to low default window size.
+            Driver.Manage().Window.Size = CommonDisplayResolutions.Fhd;
```
Piedone: The consumer might has set this already, like `Lombiq.Tests.UI.Samples.Helpers.SetupHelpers`. We shouldn't always override it.

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/EventsOrchardCoreUITestExecutorConfigurationExtensions.cs
sarahelsaig: Yes.

Lombiq.Tests.UI/Extensions/OrchardCoreDashboardUITestContextExtensions.cs
sarahelsaig: This will fail if no navigation has occurred. That's not applicable for instance if some client side validation is happening that we expect to fail.

Piedone: I see. Then the name should rather reflect this, because unlike the other *Reliably methods you'd only use this in specific scenarios.

Lombiq.Tests.UI/Pages/OrchardCoreSetupPage.cs
sarahelsaig: Can you think of a scenario where the consumer would want to manually set the setup page's screen size if they don't have to?

Lombiq.Tests.UI/Extensions/OrchardCoreDashboardUITestContextExtensions.cs
sarahelsaig: Is `ClickPublishUntilNavigation` ok?

Lombiq.Tests.UI/Pages/OrchardCoreSetupPage.cs
Piedone: It's not just for the setup page but also for the operations coming after it. For the setup page normally I wouldn't think so but keep in mind that you can override its styling (or completely) so people can have a custom setup.

Lombiq.Tests.UI/Extensions/EmailUITestContextExtensions.cs
sarahelsaig: I think that's only `GoToSmtpWebUI`.

Lombiq.Tests.UI/Extensions/OrchardCoreDashboardUITestContextExtensions.cs
Piedone: Yep.

Lombiq.Tests.UI/Pages/OrchardCoreSetupPage.cs
sarahelsaig: Do you mean there may be operations after setup inside the setup window?

Piedone: In the same browser window, yes.

sarahelsaig: ~Wouldn't it be better to have a config property for the window size?~
Never mind, that's not possible becuase we don't have access to the UITestContext inside SetupOrchardCore for some reason that I don't understand.

Piedone: We could have it as a parameter. That overload was obsoleted because we didn't need the context.

I wouldn't really put that into a config since we have statically typed ways of setting the resolution.

Lombiq.Tests.UI/Exceptions/HtmlValidationResultException.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Atata.HtmlValidation;
+using System;
+
+namespace Lombiq.Tests.UI.Exceptions
+{
+    public class HtmlValidationResultException : Exception
```
Piedone: The pair of this for accessibility is called `AccessibilityAssertionException` so I'd keep the naming pattern.

Lombiq.Tests.UI/Exceptions/HtmlValidationAssertionException.cs
```diff
@@ -1,51 +0,0 @@
...
-
-        private static string CreateErrorMessage(UITestContext context, bool createReportOnFailure)
-        {
-            var url = context.Driver.Url;
-            var title = context.Driver.Title ?? url;
-
-            var message = $"Asserting the HTML validation result on page {url}({title}) failed.";
-
-            return createReportOnFailure
-                ? message + " Check the HTML validation report in the failure dump for details."
```
Piedone: This useful info is now missing.

Lombiq.Tests.UI/Exceptions/PageChangeAssertionException.cs
```diff
@@ -0,0 +1,42 @@
...
+            : base(message, innerException)
+        {
+        }
+
+        private static string CreateErrorMessage(UITestContext context)
+        {
+            var url = context.Driver.Url;
+            var title = context.Driver.Title ?? url;
+
+            return $"Asserting the HTML validation result on page {url}({title}) failed.";
```
Piedone: This is not that exception.

Lombiq.Tests.UI/Extensions/ResponsivenessUITestContextExtensions.cs
```diff
@@ -1,8 +1,7 @@
 using Lombiq.Tests.UI.Constants;
-using Lombiq.Tests.UI.Extensions;
 using Lombiq.Tests.UI.Services;
 
-namespace Lombiq.Tests.UI.Samples.Extensions
+namespace Lombiq.Tests.UI.Extensions
```
Piedone, line 4: This was deliberately in the samples. What's a "standard" browser size depends on the project.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -98,6 +98,12 @@ public async Task<bool> ExecuteAsync(int retryCount, string dumpRootPath)
             }
             catch (Exception ex)
             {
+                if (ex is PageChangeAssertionException pageChangeAssertionException)
```
Piedone, line 101: The debug info coming from this exception would be helpful for every kind of exception we catch here, not just the ones happening during a page change.

Piedone (changes requested):
Also, while it originates from your other PR, please fix the strange analyzer violation in the build. Also see: https://teams.microsoft.com/l/message/19:525a1a3656c845ae92c5e2c0d3617cc7@thread.skype/1633442034056?tenantId=959ede5a-38e9-44e6-adc8-85553beb8220&groupId=d6a5d775-6fb9-41dc-a7bb-481d4f884046&parentMessageId=1633442034056&teamName=Lombiq%20Developers&channelName=%C3%81ltal%C3%A1nos&createdTime=1633442034056

Lombiq.Tests.UI/Exceptions/HtmlValidationAssertionException.cs
sarahelsaig: I don't see how I could maintain such an info yet keep the exception general. These seem like conflicting wants.

Piedone: The "Check the HTML validation report in the failure dump for details." message needs to be included in the exception currently named `HtmlValidationResultException`. I'm not talking about a more general exception.

Lombiq.Tests.UI/Extensions/ResponsivenessUITestContextExtensions.cs
sarahelsaig: If so, then we should rename it to something other than "standard".
Standards are decided by an issuer organization (in this case us) as a common baseline. If a project needs to test for phone, ultrawide, etc those are special scenarios and they can make their own custom extension method for those projects.

Piedone: Good point. Let's leave it like this.

Lombiq.Tests.UI/Extensions/LocationUITestContextExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+    public static class LocationUITestContextExtensions
+    {
+        public static string GetPageTitleAndAddress(this UITestContext context)
+        {
+            var url = context.Driver.Url;
+            var title = context.Driver.Title;
+
+            return string.IsNullOrEmpty(title)
+                ? url
+                : $"{url}({title})";
```
Piedone:
```suggestion
                : $"{url} ({title})";
```

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -140,6 +131,33 @@ public async Task<bool> ExecuteAsync(int retryCount, string dumpRootPath)
...
+            if (exception is PageChangeAssertionException pageChangeAssertionException)
+            {
+                _testOutputHelper.WriteLineTimestampedAndDebug(pageChangeAssertionException.Message);
+                exception = pageChangeAssertionException.InnerException;
+
+                if (exception is HtmlValidationAssertionException htmlValidationAssertionException &&
+                    htmlValidationAssertionException.HtmlValidationConfiguration.CreateReportOnFailure)
+                {
+                    _testOutputHelper.WriteLineTimestampedAndDebug(
+                        "Check the HTML validation report in the failure dump for details.");
```
Piedone: Why is this not good in the exception, as before? Then we don't need the whole block here.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -140,6 +131,33 @@ public async Task<bool> ExecuteAsync(int retryCount, string dumpRootPath)
...
+            }
+            else
+            {
+                _testOutputHelper.WriteLineTimestampedAndDebug(
+                    $"An exception has occurred while interacting with the page {_context.GetPageTitleAndAddress()}.");
+            }
+
+            _testOutputHelper.WriteLineTimestampedAndDebug($"The test failed with the following exception: {exception}");
+
+            if (exception is SetupFailedFastException) throw exception;
```
Piedone: `throw exception` will reset the stack trace: https://stackoverflow.com/a/730255/220230

Piedone (changes requested):

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
sarahelsaig: wtf!! That makes no sense. Anyway, I've reverted the auxiliary method.

PR #106, opened 2021-12-02
OSOE-52: Test should halt on error page

sarahelsaig:
[OSOE-52](https://lombiq.atlassian.net/browse/OSOE-52)

### DIFF
--- a/Lombiq.Tests.UI.Samples/Tests/ErrorHandlingTests.cs
+++ b/Lombiq.Tests.UI.Samples/Tests/ErrorHandlingTests.cs
@@ -17,8 +17,8 @@ namespace Lombiq.Tests.UI.Samples.Tests
         {
         }
 
-        // It's easier to diagnose a test failure if you know whether something is missing because there something is
-        // actually missing or there was a server side error. The below test visits a page where the action method
+        // It's easier to diagnose a test failure if you know whether an element is missing because there something is
+        // actually missing or there was a server-side error. The below test visits a page where the action method
         // throws an exception.
         [Theory, Chrome]
         public Task ErrorOnLoadedPageShouldHaltTest(Browser browser) =>
@@ -29,7 +29,7 @@ namespace Lombiq.Tests.UI.Samples.Tests
                     {
                         context.GoToErrorPageDirectly();
 
-                        // This point should be unreachable.
+                        // This point should be unreachable because logs are automatically asserted after a page load.
                         throw new InvalidOperationException("The log assertion didn't happen after page load!");
                     }
                     catch (ShouldAssertException assertException)
--- a/Lombiq.Tests.UI.Samples/Tests/ErrorHandlingTests.cs
+++ b/Lombiq.Tests.UI.Samples/Tests/ErrorHandlingTests.cs
@@ -1,0 +1,46 @@
+﻿using Lombiq.Tests.UI.Attributes;
+using Lombiq.Tests.UI.Extensions;
+using Lombiq.Tests.UI.Services;
+using Shouldly;
+using System;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Lombiq.Tests.UI.Samples.Tests
+{
+    // Sometimes errors are expected. Let's check out what can be done with them!
+    public class ErrorHandlingTests : UITestBase
+    {
+        public ErrorHandlingTests(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper)
+        {
+        }
+
+        // It's easier to diagnose a test failure if you know whether something is missing because there something is
+        // actually missing or there was a server side error. The below test visits a page where the action method
+        // throws an exception.
+        [Theory, Chrome]
+        public Task ErrorOnLoadedPageShouldHaltTest(Browser browser) =>
+            ExecuteTestAfterSetupAsync(
+                context =>
+                {
+                    try
+                    {
+                        context.GoToErrorPageDirectly();
+
+                        // This point should be unreachable.
+                        throw new InvalidOperationException("The log assertion didn't happen after page load!");
+                    }
+                    catch (ShouldAssertException assertException)
+                        when (assertException.Message.Contains("|ERROR|", StringComparison.Ordinal))
+                    {
+                        // Remove logs to have a clean slate.
+                        foreach (var log in context.Application.GetLogs()) log.Remove();
+                    }
+                },
+                browser);
+    }
+}
+
+// END OF TRAINING SECTION: Error Handling.
--- a/Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
@@ -163,7 +163,7 @@ namespace Lombiq.Tests.UI.Extensions
         /// Navigates to a page whose action method throws <see cref="InvalidOperationException"/>. This causes ASP.Net
         /// Core to display an error page.
         /// </summary>
-        public static void ThrowErrorDirectly(this UITestContext context) => GoTo<ErrorController>(context);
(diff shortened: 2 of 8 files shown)


Lombiq.Tests.UI.Samples/Tests/ErrorHandlingTests.cs
```diff
@@ -0,0 +1,46 @@
...
+    public class ErrorHandlingTests : UITestBase
+    {
+        public ErrorHandlingTests(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper)
+        {
+        }
+
+        // It's easier to diagnose a test failure if you know whether something is missing because there something is
+        // actually missing or there was a server side error. The below test visits a page where the action method
+        // throws an exception.
```
Piedone:
```suggestion
        // It's easier to diagnose a test failure if you know whether an element is missing because there something is
        // actually missing or there was a server-side error. The below test visits a page where the action method
        // throws an exception.
```

Lombiq.Tests.UI.Samples/Tests/ErrorHandlingTests.cs
```diff
@@ -0,0 +1,46 @@
...
+using Shouldly;
+using System;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Lombiq.Tests.UI.Samples.Tests
+{
+    // Sometimes errors are expected. Let's check out what can be done with them!
+    public class ErrorHandlingTests : UITestBase
```
Piedone, line 13: Nice!

Lombiq.Tests.UI.Samples/Tests/ErrorHandlingTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        [Theory, Chrome]
+        public Task ErrorOnLoadedPageShouldHaltTest(Browser browser) =>
+            ExecuteTestAfterSetupAsync(
+                context =>
+                {
+                    try
+                    {
+                        context.GoToErrorPageDirectly();
+
+                        // This point should be unreachable.
```
Piedone:
```suggestion
                        // This point should be unreachable because logs are automatically asserted after a page load.
```

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
```diff
@@ -143,12 +154,33 @@ public static void PurgeMediaCacheDirectly(this UITestContext context, bool togg
...
         /// </summary>
         public static void ExecuteRecipeDirectly(this UITestContext context, string recipeName) =>
-            context.GoToRelativeUrl("/Lombiq.Tests.UI.Shortcuts/Recipe/Execute?recipeName=" + recipeName);
+            GoTo<RecipeController>(
+                context,
+                nameof(RecipeController.Execute),
+                (nameof(recipeName), recipeName));
+
+        /// <summary>
+        /// Navigates to a page whose action method throws <see cref="InvalidOperationException"/>. This causes ASP.Net
```
Piedone:
```suggestion
        /// Navigates to a page whose action method throws <see cref="InvalidOperationException"/>. This causes ASP.NET
```

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
```diff
@@ -143,12 +154,33 @@ public static void PurgeMediaCacheDirectly(this UITestContext context, bool togg
...
+        /// Core to display an error page.
+        /// </summary>
+        public static void GoToErrorPageDirectly(this UITestContext context) => GoTo<ErrorController>(context);
+
+        private static void GoTo<TController>(
```
Piedone, line 168: Nice!

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
```diff
@@ -143,12 +154,33 @@ public static void PurgeMediaCacheDirectly(this UITestContext context, bool togg
...
+            UITestContext context,
```
0liver, line 169: Wouldn't it be desirable to add `this` here and invoke `context.GoTo<TController>()` as an extension method? That would read nicer in connection with e.g. `context.GoToRelativeUrl()`.

Piedone: I agree, indeed.

sarahelsaig: Hmm, I never use extension methods in the same class they are defined. It make it clearer that this is a local method. If you think that's not logical I'm not overly attached to that practice. Just want to share my rationale for not doing it that way.

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
sarahelsaig: Thanks. Are we interested in making a more general version for HL?

Piedone: We talked about having some kind of statically typed way of routing, I'd be happy to see it.

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
Piedone: I don't have a strong feeling either way but an extension does "look" nicer to me.

0liver: @DAud-IcI Why is it important for you to make it clear that something is a local method? I'm asking out of curiosity.

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
0liver: Me too!

0liver: I'd love to port it to O1 and use it in CARE.

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
sarahelsaig: After sleeping on it, I think I explained myself poorly. Its better to say that when I use extension method I expect it to not be local because it usually isn't. So when it is, that's confusing. Additionally when I see a private extension method, that's an instant WTF moment for me and have to do a double take to see if that's a mistake or what.

0liver: That last point convinces me.

I'm just wondering - why would you NOT make it public in the first place?

sarahelsaig: Because it's highly specific to the shortcuts module that we only invoke through these other extensions in this class. I plan to make a more general version and that can be public.

0liver: OK. Thank you for elaborating on this. I appreciate it.

Piedone: Let's leave it like then.

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
sarahelsaig: Will do, added it to my backlog: https://lombiq.atlassian.net/browse/OSOE-53

Issue #80, opened 2021-09-05
Support multi-tenancy (OSOE-364)


Piedone:
Currently, the Toolbox has some references to the Default site. If we want to test multi-tenancy, i.e. creating and launching tenants, then this needs to be extended.

- [x] Update `TenantTests` too. 
- [x] Needs prefixing in `GoToPageAsync()` and in `GoToRelativeUrlAsync()`.
- [x] Needs refactoring of `CreateAndEnterTenantAsync()`.
- [x] Support multi-tenancy with SQL Server. The logic changing the SQL Server connection string in the *appsettings.json* file can be substituted with using the [Database Shells Configuration Provider](https://docs.orchardcore.net/en/latest/docs/reference/core/Shells/#database-shells-configuration-provider and changing all connection strings in a query.
- [x] Support multi-tenancy with Azure Blob Storage. The logic in `AzureBlobStorageManager` can be changed to use a templated `{{ ShellSettings.Name }} BasePath`.
- [x] Support multi-tenancy for snapshots.
- [x] Refactor `Lombiq.Hosting.Tenants.FeaturesGuard.Tests.UI.Extensions.TestCaseUITestContextExtensions`.

sarahelsaig:
The usage of "Default" could be replaced with [`UITestContext.TenantName`](https://github.com/Lombiq/UI-Testing-Toolbox/blob/caac57ddb82db01bfb785f84e70de52d626ee1b7/Lombiq.Tests.UI/Services/UITestContext.cs#L59) which is already used by the [`TypedRoute`](https://github.com/Lombiq/Helpful-Libraries/blob/e619b8b04755345156218fd57f5f226abf09dbc9/Lombiq.HelpfulLibraries/Libraries/Mvc/TypedRoute.cs). It's currently always set to "Default" and the user must manually change it, but it could be initialized using the way you described.

PR #107, opened 2021-12-05
OSOE-53: RouteModel, unit test fixes, orchard-core-preview

sarahelsaig:
[OSOE-53](https://lombiq.atlassian.net/browse/OSOE-53)
Depends on OSOE-52.
Root PR: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/25


Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
```diff
@@ -35,7 +30,7 @@ public static class ShortcutsUITestContextExtensions
         /// enabled.
         /// </summary>
         public static void SignInDirectly(this UITestContext context, string userName = DefaultUser.UserName) =>
-                GoTo<AccountController>(context, nameof(AccountController.SignInDirectly), (nameof(userName), userName));
+            context.GoTo<AccountController>(controller => controller.SignInDirectly(userName));
```
Piedone, line 33: We get these:

image

Lombiq.Tests.UI/Services/UITestContext.cs
```diff
@@ -53,6 +53,11 @@ public class UITestContext
         /// </summary>
         public Dictionary<string, object> CustomContext { get; } = new();
 
+        /// <summary>
+        /// Gets or sets the current tenant name when testing multi-tenancy.
+        /// </summary>
+        public string TenantName { get; set; }
```
Piedone: This is never set.

Piedone (changes requested):

Lombiq.Tests.UI/Services/UITestContext.cs
sarahelsaig: Yes, it's just a placeholder. I'm not experienced with multitenancy but I assume we will want to be able to test such sites at some point in the future? At that point this prop could be set by the test.

Piedone: Set it to `"Default"` here then please, and make a note about why is it like that there and here: https://github.com/Lombiq/UI-Testing-Toolbox/issues/80

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
sarahelsaig: Any idea why it doesn't show up in CI?

Piedone: .NET SDK warnings have this issue for some reason, but not always...

PR #109, opened 2021-12-07
Monkey testing

YevgeniyShunevych:
Corresponding PR in OSOCE: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/29


Lombiq.Tests.UI.Samples/Tests/MonkeyTests.cs
```diff
@@ -0,0 +1,57 @@
...
+using Lombiq.Tests.UI.MonkeyTesting;
+using Lombiq.Tests.UI.Services;
+using System;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Lombiq.Tests.UI.Samples.Tests
+{
+    public class MonkeyTests : UITestBase
```
Piedone, line 15: Document this as the other sample tests here, and insert it into the training steps after `ErrorHandlingTests`.

Lombiq.Tests.UI/Extensions/MonkeyTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,41 @@
+using Lombiq.Tests.UI.MonkeyTesting;
+using Lombiq.Tests.UI.Services;
+
+namespace Lombiq.Tests.UI.Extensions
+{
+    /// <summary>
+    /// Provides a set of extension methods for monkey testing.
+    /// </summary>
+    public static class MonkeyTestingUITestContextExtensions
```
Piedone, line 10: Monkey testing availability needs a mention in the root Readme and _CreatingTests.md_, just as basic Orchard tests. As well as a mention of Gremlin would be necessary in _Tools.md_. Also some notes about how to reproduce and fix issues uncovered by such tests in _Troubleshooting.md_.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
```diff
@@ -0,0 +1,125 @@
...
+
+namespace Lombiq.Tests.UI.MonkeyTesting
+{
+    /// <summary>
+    /// Represents the options of monkey testing.
+    /// </summary>
+    public sealed class MonkeyTestingOptions
+    {
+        /// <summary>
+        /// Gets or sets the base random seed.
```
Piedone: While this technically documents the property, it doesn't explain what it's used for (what is a base random seed?). Same for the others.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// Gets or sets a value indicating whether to run application log assertion.
+        /// The default value is <see langword="true"/>.
+        /// </summary>
+        public bool RunAppLogAssertion { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to run browser log assertion.
+        /// The default value is <see langword="true"/>.
+        /// </summary>
+        public bool RunBrowserLogAssertion { get; set; } = true;
```
Piedone: Instead of being new options here, these checks should all be driven by `context.Configuration`.

Lombiq.Tests.UI/Extensions/MonkeyTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        /// <summary>
+        /// Tests the current page as monkey.
+        /// Test finishes by timeout or when the current page is left during testing.
+        /// </summary>
+        /// <param name="context">The context.</param>
```
Piedone: No need to add docs for the extended object.

Lombiq.Tests.UI.Samples/Tests/MonkeyTests.cs
```diff
@@ -0,0 +1,57 @@
...
+            RunHtmlValidationAssertion = false,
+        };
+
+        public MonkeyTests(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper)
+        {
+        }
+
+        [Theory, Chrome]
+        public Task TestCurrentPageAsMonkeyShouldWorkWithRandomSeed(Browser browser) =>
```
Piedone:
```suggestion
        public Task TestCurrentPageAsMonkeyShouldWorkWithConfiguredRandomSeed(Browser browser) =>
```

Lombiq.Tests.UI/Extensions/MonkeyTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <param name="options">The options.</param>
+        /// <param name="randomSeed">The random seed.</param>
```
Piedone: Explain these, also for the other method.

Lombiq.Tests.UI/Extensions/MonkeyTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <returns>The same <see cref="UITestContext"/> instance.</returns>
+        public static UITestContext TestCurrentPageAsMonkey(this UITestContext context, MonkeyTestingOptions options = null, int? randomSeed = null)
```
Piedone: Parameter lists are easy to break, so please break at 120 characters.

Lombiq.Tests.UI/Extensions/MonkeyTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            var monkeyTester = new MonkeyTester(context, options);
+            monkeyTester.TestOnePage(randomSeed);
+
+            return context;
+        }
+
+        /// <summary>
+        /// Tests the current page as monkey recursively.
```
Piedone: What does recursively mean? Add it to the comment.

Lombiq.Tests.UI/MonkeyTesting/GremlinsScriptBuilder.cs
```diff
@@ -0,0 +1,47 @@
...
+                gremlins.strategies.distribution({{
+                    nb: {NumberOfAttacks},
+                    delay: {AttackDelay}
+                }})
+            ],
+            randomizer: new gremlins.Chance({RandomSeed})
+        }}).unleash();
+    }}
+    var s = document.createElement('script');
+    s.src = 'https://unpkg.com/gremlins.js';
```
Piedone: We should avoid using resources out of our control during test execution. This could rather come from an NPM package and loaded locally.

Lombiq.Tests.UI/MonkeyTesting/IMonkeyTestingUrlCleaner.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// </summary>
+    public interface IMonkeyTestingUrlCleaner
+    {
+        /// <summary>
+        /// Cleans the specified URL.
+        /// </summary>
+        /// <param name="url">The URL.</param>
+        /// <param name="context">The context.</param>
+        /// <returns>A cleaned or original URL.</returns>
+        string Handle(string url, UITestContext context);
```
Piedone: `context` is usually the first parameter. Also, passing URLs is better as an `Uri` object. Same for `IMonkeyTestingUrlFilter`.

Lombiq.Tests.UI/MonkeyTesting/IMonkeyTestingUrlFilter.cs
```diff
@@ -0,0 +1,20 @@
...
+    public interface IMonkeyTestingUrlFilter
+    {
+        /// <summary>
+        /// Determines whether this filter allows the specified URL to be tested.
+        /// </summary>
+        /// <param name="url">The URL.</param>
+        /// <param name="context">The context.</param>
+        /// <returns>
+        ///   <see langword="true"/> if URL passes the filter; otherwise, <see langword="false"/>.
+        /// </returns>
```
Piedone: No need for indentations. Also can be a single line:
```suggestion
        /// <returns><see langword="true"/> if URL passes the filter; otherwise, <see langword="false"/>.</returns>
```

Lombiq.Tests.UI/MonkeyTesting/RemovesFragmentMonkeyTestingUrlCleaner.cs
```diff
@@ -0,0 +1,12 @@
+using Lombiq.Tests.UI.Services;
+using System;
+
+namespace Lombiq.Tests.UI.MonkeyTesting
+{
+    public sealed class RemovesFragmentMonkeyTestingUrlCleaner : IMonkeyTestingUrlCleaner
+    {
+        public string Handle(string url, UITestContext context) =>
+            new Uri(url, UriKind.RelativeOrAbsolute)
+                .GetComponents(UriComponents.HttpRequestUrl, UriFormat.Unescaped);
```
Piedone: What does this clean out?

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <para>By default contains:</para>
+        /// <list type="number">
+        /// <item><description><see cref="RemovesFragmentMonkeyTestingUrlCleaner"/> instance.</description></item>
+        /// <item><description><see cref="RemovesBaseUrlMonkeyTestingUrlCleaner"/> instance.</description></item>
+        /// <item><description><see cref="RemovesQueryParameterMonkeyTestingUrlCleaner"/> instance with <c>"admin"</c> argument.</description></item>
+        /// <item><description><see cref="RemovesQueryParameterMonkeyTestingUrlCleaner"/> instance with <c>"returnUrl"</c> argument.</description>
+        /// </item>
+        /// </list>
+        /// </summary>
+        public List<IMonkeyTestingUrlCleaner> UrlCleaners { get; } = new()
```
Piedone: "Cleaner" and "Filter" are similar terms and thus a bit confusing. I don't have a concrete better suggestion though.

Lombiq.Tests.UI/MonkeyTesting/RemovesBaseUrlMonkeyTestingUrlCleaner.cs
```diff
@@ -0,0 +1,23 @@
+using Lombiq.Tests.UI.Services;
+using System;
+
+namespace Lombiq.Tests.UI.MonkeyTesting
+{
+    public sealed class RemovesBaseUrlMonkeyTestingUrlCleaner : IMonkeyTestingUrlCleaner
```
Piedone: Why is this necessary?

Lombiq.Tests.UI/MonkeyTesting/RemovesByRegexMonkeyTestingUrlCleaner.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Text.RegularExpressions;
+
+namespace Lombiq.Tests.UI.MonkeyTesting
+{
+    public class RemovesByRegexMonkeyTestingUrlCleaner : IMonkeyTestingUrlCleaner
+    {
+        private readonly Regex _regex;
+
+        public RemovesByRegexMonkeyTestingUrlCleaner(string regexPattern)
+            : this(new Regex(regexPattern))
```
Piedone: Perhaps use `RegexOptions.Compiled`? Same for `MatchesRegexMonkeyTestingUrlFilter`.

Lombiq.Tests.UI/MonkeyTesting/RemovesQueryParameterMonkeyTestingUrlCleaner.cs
```diff
@@ -0,0 +1,10 @@
+namespace Lombiq.Tests.UI.MonkeyTesting
+{
+    public class RemovesQueryParameterMonkeyTestingUrlCleaner : RemovesByRegexMonkeyTestingUrlCleaner
+    {
+        public RemovesQueryParameterMonkeyTestingUrlCleaner(string parameterName)
+            : base(@$"(\b{parameterName}=[^&]*&|[\?&]{parameterName}=[^&]*$)")
```
Piedone: Why is `\b{parameterName}=[^&]*&` needed? Parameters should always start with a "?" or "&", what `[\?&]{parameterName}=[^&]*$` handles.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            new RemovesFragmentMonkeyTestingUrlCleaner(),
+            new RemovesBaseUrlMonkeyTestingUrlCleaner(),
+            new RemovesQueryParameterMonkeyTestingUrlCleaner("admin"),
+            new RemovesQueryParameterMonkeyTestingUrlCleaner("returnUrl"),
```
Piedone: A comment on why these are necessary would be useful.

Lombiq.Tests.UI/Services/UITestContext.cs
```diff
@@ -88,5 +88,14 @@ public async Task AssertBrowserLogAsync()
             var browserLog = await Scope.Driver.GetAndEmptyBrowserLogAsync();
             Configuration.AssertBrowserLog?.Invoke(browserLog);
         }
+
+        /// <summary>
+        /// Run an assertion on the application logs of the current tab with the delegate configured in <see
+        /// cref="Configuration"/>.
+        /// </summary>
+        public async Task AssertAppLogAsync()
```
Piedone: This should rather call `OrchardCoreUITestExecutorConfiguration.AssertAppLogsMaybeAsync()`. `AssertBrowserLogAsync()` was here before but it too should similarly call `AssertBrowserLogMaybe()` there.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+                () =>
+                {
+                    WriteOptionsToLog();
+
+                    var pageTestInfo = GetCurrentPageTestInfo();
+
+                    if (randomSeed is null)
+                        TestCurrentPage(pageTestInfo);
+                    else
+                        TestCurrentPageWithRandomSeed(pageTestInfo, randomSeed.Value);
```
Piedone:
```suggestion
                    if (randomSeed is null) TestCurrentPage(pageTestInfo);
                    else TestCurrentPageWithRandomSeed(pageTestInfo, randomSeed.Value);
```

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+        private readonly List<PageMonkeyTestInfo> _pageTestInfoList = new();
+
+        internal MonkeyTester(UITestContext context, MonkeyTestingOptions options = null)
+        {
+            _context = context ?? throw new ArgumentNullException(nameof(context));
+            _options = options ?? new MonkeyTestingOptions();
+            _random = new Random(_options.BaseRandomSeed);
+        }
+
+        private ILogManager Log => _context.Scope.AtataContext.Log;
```
Piedone:
```suggestion
        private const string SetIsMonkeyTestRunningScript = "window.isMonkeyTestRunning = true;";
        private const string GetIsMonkeyTestRunningScript = "return !!window.isMonkeyTestRunning;";

        private readonly UITestContext _context;
        private readonly MonkeyTestingOptions _options;
        private readonly Random _random;
        private readonly List<PageMonkeyTestInfo> _pageTestInfoList = new();

        private ILogManager Log => _context.Scope.AtataContext.Log;

        internal MonkeyTester(UITestContext context, MonkeyTestingOptions options = null)
        {
            _context = context ?? throw new ArgumentNullException(nameof(context));
            _options = options ?? new MonkeyTestingOptions();
            _random = new Random(_options.BaseRandomSeed);
        }
```

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+            Log.Trace(@$"Monkey testing options:
+- {nameof(MonkeyTestingOptions.BaseRandomSeed)}={_options.BaseRandomSeed}
+- {nameof(MonkeyTestingOptions.PageTestTime)}={_options.PageTestTime.ToShortIntervalString()}
+- {nameof(MonkeyTestingOptions.PageMarkerPollingInterval)}={_options.PageMarkerPollingInterval.ToShortIntervalString()}
+- {nameof(MonkeyTestingOptions.RunAccessibilityCheckingAssertion)}={_options.RunAccessibilityCheckingAssertion}
+- {nameof(MonkeyTestingOptions.RunHtmlValidationAssertion)}={_options.RunHtmlValidationAssertion}
+- {nameof(MonkeyTestingOptions.RunAppLogAssertion)}={_options.RunAppLogAssertion}
+- {nameof(MonkeyTestingOptions.RunBrowserLogAssertion)}={_options.RunBrowserLogAssertion}
+- {nameof(MonkeyTestingOptions.GremlinsSpecies)}={string.Join(',', _options.GremlinsSpecies)}
+- {nameof(MonkeyTestingOptions.GremlinsMogwais)}={string.Join(',', _options.GremlinsMogwais)}
```
Piedone:
```suggestion
- {nameof(MonkeyTestingOptions.GremlinsSpecies)}={string.Join(", ", _options.GremlinsSpecies)}
- {nameof(MonkeyTestingOptions.GremlinsMogwais)}={string.Join(", ", _options.GremlinsMogwais)}
```

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+            pageTestInfo = _pageTestInfoList.FirstOrDefault(x => x.HasTimeToTest);
+            return pageTestInfo != null;
+        }
+
+        private PageMonkeyTestInfo GetCurrentPageTestInfo()
+        {
+            var url = _context.Driver.Url;
+            var cleanUrl = CleanUrl(url);
+
+            var pageTestInfo = _pageTestInfoList.FirstOrDefault(x => x.CleanUrl == cleanUrl)
```
Piedone: Use proper variable names instead of `x`. Also in the other places.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+                       $"Monkey test \"{pageTestInfo.CleanUrl}\" within {pageTestInfo.TimeToTest.ToShortIntervalString()} with {randomSeed} random seed"),
+#pragma warning restore S103 // Lines should not be too long
+                    () =>
+                    {
+                        ExecutePreAssertions();
+
+                        var pageTestTimeLeft = TestCurrentPageAndMeasureTestTimeLeft(pageTestInfo.TimeToTest, randomSeed);
+                        pageTestInfo.TimeToTest = pageTestTimeLeft;
+                        if (!_pageTestInfoList.Contains(pageTestInfo))
+                            _pageTestInfoList.Add(pageTestInfo);
```
Piedone:
```suggestion
                        if (!_pageTestInfoList.Contains(pageTestInfo))  _pageTestInfoList.Add(pageTestInfo);
```

Lombiq.Tests.UI/MonkeyTesting/IMonkeyTestingUrlFilter.cs
```diff
@@ -0,0 +1,20 @@
...
+        bool CanHandle(string url, UITestContext context);
```
Piedone:
```suggestion
        bool AllowUrl(string url, UITestContext context);
```

Lombiq.Tests.UI/MonkeyTesting/IMonkeyTestingUrlCleaner.cs
Piedone:
```suggestion
        string Clean(string url, UITestContext context);
```

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+        private bool CanTestPage(PageMonkeyTestInfo pageTestInfo)
+        {
+            bool canTest = pageTestInfo.HasTimeToTest && ShouldTestPageUrl(pageTestInfo.Url);
+
+            if (!canTest)
+            {
+                Log.Info(
+                    !pageTestInfo.HasTimeToTest
+                    ? $"\"{pageTestInfo.CleanUrl}\" is tested completely"
+                    : $"Navigated to \"{pageTestInfo.Url}\" that should not be tested");
```
Piedone:
```suggestion
                    ? $"Available monkey testing time for \"{pageTestInfo.CleanUrl}\" is up and thus testing is complete."
                    : $"Navigated to \"{pageTestInfo.Url}\" that should not be tested.");
```

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+
+        private const string GetIsMonkeyTestRunningScript = "return !!window.isMonkeyTestRunning;";
+
+        private readonly UITestContext _context;
+
+        private readonly MonkeyTestingOptions _options;
+
+        private readonly Random _random;
+
+        private readonly List<PageMonkeyTestInfo> _pageTestInfoList = new();
```
Piedone:
```suggestion
        private readonly List<PageMonkeyTestInfo> _visitedPages = new();
```

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+
+            Log.Info($"Current page is \"{pageTestInfo.CleanUrl}\"");
+
+            return pageTestInfo;
+        }
+
+        private string CleanUrl(string url)
+        {
+            foreach (var cleaner in _options.UrlCleaners)
+                url = cleaner.Handle(url, _context);
```
Piedone:
```suggestion
            foreach (var cleaner in _options.UrlCleaners)  url = cleaner.Handle(url, _context);
```

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+
+        private PageMonkeyTestInfo GetCurrentPageTestInfo()
+        {
+            var url = _context.Driver.Url;
+            var cleanUrl = CleanUrl(url);
+
+            var pageTestInfo = _pageTestInfoList.FirstOrDefault(x => x.CleanUrl == cleanUrl)
+                ?? new PageMonkeyTestInfo(url, cleanUrl, _options.PageTestTime);
+
+            Log.Info($"Current page is \"{pageTestInfo.CleanUrl}\"");
```
Piedone:
```suggestion
            Log.Info($"Current page is \"{pageTestInfo.CleanUrl}\".");
```

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+                    : $"Navigated to \"{pageTestInfo.Url}\" that should not be tested");
+            }
+
+            return canTest;
+        }
+
+        private bool ShouldTestPageUrl(string url) =>
+            _options.UrlFilters.All(filter => filter.CanHandle(url, _context));
+
+        private bool TryGetLeftPageToTest(out PageMonkeyTestInfo pageTestInfo)
```
Piedone:
```suggestion
        private bool TryGetAvailablePageToTest(out PageMonkeyTestInfo pageTestInfo)
```

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+
+        private void ExecutePostAssertions()
+        {
+            if (_options.RunAppLogAssertion) _context.AssertAppLogAsync().GetAwaiter().GetResult();
+
+            if (_options.RunBrowserLogAssertion) _context.AssertBrowserLogAsync().GetAwaiter().GetResult();
+        }
+
+        [SuppressMessage("Security", "SCS0005:Weak random number generator.", Justification = "For current purpose it should not be secured.")]
+        [SuppressMessage("Security", "CA5394:Do not use insecure randomness", Justification = "For current purpose it should not be secured.")]
```
Piedone: Use `NonSecurityRandomizer` from Helpful Libraries instead.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+            if (_options.RunAccessibilityCheckingAssertion) _context.AssertAccessibility();
+
+            if (_options.RunHtmlValidationAssertion) _context.AssertHtmlValidityAsync().GetAwaiter().GetResult();
+        }
+
+        private void ExecutePostAssertions()
+        {
+            if (_options.RunAppLogAssertion) _context.AssertAppLogAsync().GetAwaiter().GetResult();
+
+            if (_options.RunBrowserLogAssertion) _context.AssertBrowserLogAsync().GetAwaiter().GetResult();
```
Piedone: Everything up to `TestCurrentPageAsMonkey()` and `TestCurrentPageAsMonkeyRecursively()` can be `async`, and for them there can be async counterparts. That way, you get the convenience of forced sync code but if the caller is actually async, then the advantage of async.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+                        ExecutePostAssertions();
+                    });
+            }
+            catch (Exception exception)
+            {
+                throw new AssertionException($"Failure on \"{pageTestInfo.CleanUrl}\" page", exception);
+            }
+        }
+
+        private void ExecutePreAssertions()
```
Piedone: Instead of this and `ExecutePostAssertions()` rather `context.Configuration.Events.AfterPageChange` should be invoked after the test (i.e. when `ExecutePostAssertions()` is called currently), which will run the centrally configured accessibility and HTML validation assertions, if necessary, as well as assert on the browser and app logs. `EventsOrchardCoreUITestExecutorConfigurationExtensions.OnEventsAfterNavigationAsync()` should be refactored to support external calls too to allow this.

Also, if you don't use `_context.Scope.AtataContext.Go.ToUrl()` on line 71 but `_context.GoToAbsoluteUrl()` then these assertions will also happen on page load since events are raised.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        private void TestCurrentPageWithRandomSeed(PageMonkeyTestInfo pageTestInfo, int randomSeed)
+        {
+            try
+            {
+                Log.ExecuteSection(
+                    new LogSection(
+#pragma warning disable S103 // Lines should not be too long
+                       $"Monkey test \"{pageTestInfo.CleanUrl}\" within {pageTestInfo.TimeToTest.ToShortIntervalString()} with {randomSeed} random seed"),
```
Piedone:
```suggestion
                       $"Monkey test \"{pageTestInfo.CleanUrl}\" within {pageTestInfo.TimeToTest.ToShortIntervalString()} with {randomSeed} random seed."),
```

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+                        pageTestInfo.TimeToTest = pageTestTimeLeft;
+                        if (!_pageTestInfoList.Contains(pageTestInfo))
+                            _pageTestInfoList.Add(pageTestInfo);
+
+                        ExecutePostAssertions();
+                    });
+            }
+            catch (Exception exception)
+            {
+                throw new AssertionException($"Failure on \"{pageTestInfo.CleanUrl}\" page", exception);
```
Piedone:
```suggestion
                throw new AssertionException($"Failure on \"{pageTestInfo.CleanUrl}\" page.", exception);
```

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,245 @@
...
+
+                        var pageTestTimeLeft = TestCurrentPageAndMeasureTestTimeLeft(pageTestInfo.TimeToTest, randomSeed);
+                        pageTestInfo.TimeToTest = pageTestTimeLeft;
+                        if (!_pageTestInfoList.Contains(pageTestInfo))
+                            _pageTestInfoList.Add(pageTestInfo);
+
+                        ExecutePostAssertions();
+                    });
+            }
+            catch (Exception exception)
```
Piedone:
```suggestion
            catch (Exception ex)
```

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
Piedone: Is this try-catch necessary? If an assertion fails, then there will be an exception and bubble up to the top, causing the test to fail. During that, the log entry will contain the URL too (like "An exception has occurred while interacting with the page https://localhost:9308/Login").

Piedone (changes requested):

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: Right, you can get that info in log. Exception wrapping just gives you failed page URL immediately in exception message, without a need to get to log. I also had doubts on need of this exception wrapping. Anyway, we always check logs. @Piedone, please decide, should I leave this try-catch or remove? I'm OK with both approaches.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: `NonSecurityRandomizer` doesn't have a constructor with seed parameter, which I need. Should I add it to `NonSecurityRandomizer`? It will also need another PR for Lombiq/Helpful-Libraries  repo.

Lombiq.Tests.UI/Extensions/MonkeyTestingUITestContextExtensions.cs
YevgeniyShunevych: Updated.

Lombiq.Tests.UI/Extensions/MonkeyTestingUITestContextExtensions.cs
YevgeniyShunevych: Done.

Lombiq.Tests.UI/Extensions/MonkeyTestingUITestContextExtensions.cs
YevgeniyShunevych: Updated.

Lombiq.Tests.UI/MonkeyTesting/RemovesFragmentMonkeyTestingUrlCleaner.cs
YevgeniyShunevych: A fragment part of URL, like "someurl#frag" -> "someurl".

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
YevgeniyShunevych: I have the same naming problem. As an option: `Cleaner` -> `Sanitizer`.

Lombiq.Tests.UI/MonkeyTesting/RemovesBaseUrlMonkeyTestingUrlCleaner.cs
YevgeniyShunevych: It isn't necessary. Just to keep URLs shorter (relative) in log. If you want, I can remove it completely.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
Piedone: I don't think we need it.

Lombiq.Tests.UI/MonkeyTesting/RemovesFragmentMonkeyTestingUrlCleaner.cs
Piedone: Ah I see, thanks.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
Piedone: Sounds better!

Lombiq.Tests.UI/MonkeyTesting/RemovesBaseUrlMonkeyTestingUrlCleaner.cs
Piedone: I see, that's good. Just please document its purpose on the class.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
Piedone: Right. For the sake of simplicity I added it.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: Removed.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: Resolving this, as try/catch was completely removed.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
YevgeniyShunevych: Renamed all cleaners to sanitizers.

Lombiq.Tests.UI/MonkeyTesting/RemovesBaseUrlMonkeyTestingUrlCleaner.cs
YevgeniyShunevych: Added a comment.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
YevgeniyShunevych: Sorry, a comment in code? Or you don't understand why they are needed?

You wrote in requirements that "admin" parameter should be disregarded. I decided that "returnUrl", which occurs on Login page, should also be disregarded, as it doesn't change the page logic very much, and to avoid extra monkey cycles on Login page with different sources.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: Renamed.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
Piedone: A comment in code, yes. I understand the reason for these and I remember I asked for the admin one but for someone else looking at this piece of code it's not obvious.

YevgeniyShunevych: Added to XML comments.

Lombiq.Tests.UI/MonkeyTesting/IMonkeyTestingUrlCleaner.cs
YevgeniyShunevych: Updated.

Lombiq.Tests.UI/MonkeyTesting/RemovesByRegexMonkeyTestingUrlCleaner.cs
YevgeniyShunevych: Updated.

Lombiq.Tests.UI/MonkeyTesting/IMonkeyTestingUrlFilter.cs
YevgeniyShunevych: Updated.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: Renamed.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: Renamed.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: Applied `NonSecurityRandomizer`.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: OK. I can move `EventsOrchardCoreUITestExecutorConfigurationExtensions.OnEventsAfterNavigationAsync()` to `UITestContext` and rename it to `TriggerAfterPageChangeEventAsync()` and make it public or internal. What do think?

Lombiq.Tests.UI/MonkeyTesting/RemovesQueryParameterMonkeyTestingUrlCleaner.cs
YevgeniyShunevych: Here is where I got this regex: https://stackoverflow.com/a/8793662/8063294. Explanation is there.

Lombiq.Tests.UI.Samples/Tests/MonkeyTests.cs
YevgeniyShunevych: Done.

Lombiq.Tests.UI/Extensions/MonkeyTestingUITestContextExtensions.cs
YevgeniyShunevych: Done. If you want, @Piedone, you can correct that monkey testing docs additions, as you have better overall vision of the project docs.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
Piedone: Sounds good.

Lombiq.Tests.UI/Extensions/MonkeyTestingUITestContextExtensions.cs
Piedone: Please attempt something.

Lombiq.Tests.UI/MonkeyTesting/RemovesQueryParameterMonkeyTestingUrlCleaner.cs
Piedone: I see, looks OK.

Lombiq.Tests.UI/Extensions/MonkeyTestingUITestContextExtensions.cs
YevgeniyShunevych: What do you mean? I've added info to docs `md` files and `MonkeyTests.cs`.

Piedone: Ah I see, great then.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: I'm OK regarding inner methods. But `TestCurrentPageAsMonkey()` and `TestCurrentPageAsMonkeyRecursively()` should remain non-async, right? Otherwise it will be uncomfortable to use these methods in tests inside `ExecuteTestAfterSetupAsync`, as `Action<UITestContext> test` parameter is non-async.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
YevgeniyShunevych: Added comments.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
YevgeniyShunevych: Done. Please check.

Lombiq.Tests.UI/Services/UITestContext.cs
YevgeniyShunevych: I removed this method as I don't use and need it anymore.

Lombiq.Tests.UI/MonkeyTesting/GremlinsScriptBuilder.cs
YevgeniyShunevych: @Piedone, could you please describe how you see such implementation?

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
Piedone: What I'm saying is that they should be async by default but have sync overloads that make sync calls to them. We mostly need sync code in tests but if someone can use async then let's optimize for that.

Lombiq.Tests.UI/MonkeyTesting/GremlinsScriptBuilder.cs
Piedone: I'm just thinking aloud but e.g. you can add an NPM dependency for Gremlin to `Lombiq.Tests.UI`. NPM package restore can be integrated into MSBuild with our [NPM Targets](https://github.com/Lombiq/NPM-Targets) project which is used by multiple other projects in the OSOCE solution too.

Then, you can load the file as an inline JS here.

As an alternative, you can add the NPM package to `Lombiq.Tests.UI.AppExtensions` instead (which is loaded by the web app under test) and register the file to be accessible under a local URL in the web app from `ConfigureUITesting()`. You can then load the script from that URL.

Also, [Selenium.Axe for .NET](https://github.com/TroyWalshProf/SeleniumAxeDotnet) loads the axe JS from NPM during UI testing somehow, what we talked about before, that's also a viable approach.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
```diff
@@ -10,47 +10,21 @@ namespace Lombiq.Tests.UI.MonkeyTesting
...
-        /// Gets or sets the page test time.
-        /// The default value is 60 seconds.
+        /// Gets or sets the page test time. The default value is 60 seconds.
         /// </summary>
         public TimeSpan PageTestTime { get; set; } = TimeSpan.FromSeconds(60);
 
         /// <summary>
-        /// Gets or sets the page marker polling interval.
-        /// The default value is 200 milliseconds.
+        /// Gets or sets the page marker polling interval. The default value is 200 milliseconds.
```
Piedone: It's not clear from the docs what this is doing.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
```diff
@@ -10,47 +10,21 @@ namespace Lombiq.Tests.UI.MonkeyTesting
...
-        /// Gets or sets a value indicating whether to run application log assertion.
-        /// The default value is <see langword="true"/>.
-        /// </summary>
-        public bool RunAppLogAssertion { get; set; } = true;
-
-        /// <summary>
-        /// Gets or sets a value indicating whether to run browser log assertion.
-        /// The default value is <see langword="true"/>.
-        /// </summary>
-        public bool RunBrowserLogAssertion { get; set; } = true;
```
Piedone: When running `TestContensAdminAsMonkeyRecursivelyShouldWorkWithAdminUser` 404 pages (like https://localhost:9312/Admin/Contents/ContentItems&Options.SelectedContentType=Article#! that it visits) which show up in the browser log are not caught. This is due to the log coming from the cache:

image

Probably this cache should still always retrieve log entries but add them to the existing ones (this additive behavior wouldn't matter during monkey testing but for other tests it can).

Once this is fixed, `TestContensAdminAsMonkeyRecursivelyShouldWorkWithAdminUser` will (rightfully, due to an Orchard bug) fail. At that point, please rename and change the test to be restricted to `/Admin/BackgroundTasks` instead.

Lombiq.Tests.UI/MonkeyTesting/IMonkeyTestingUrlSanitizer.cs
```diff
@@ -1,18 +1,19 @@
...
     {
         /// <summary>
         /// Cleans the specified URL.
         /// </summary>
-        /// <param name="url">The URL.</param>
         /// <param name="context">The context.</param>
+        /// <param name="url">The URL.</param>
         /// <returns>A cleaned or original URL.</returns>
-        string Handle(string url, UITestContext context);
+        Uri Clean(UITestContext context, Uri url);
```
Piedone: Now with the type name being Sanitizer, this and the comment should be adapted too.

Lombiq.Tests.UI/Docs/Troubleshooting.md
```diff
@@ -53,3 +53,9 @@
 
     EXEC sp_executesql @sql 
     ```
+
+
+## Monkey testing
```
Piedone, line 59: How would you troubleshoot which interaction exactly causes the 404 around 4s

https://user-images.githubusercontent.com/1976647/146612427-8766f3a9-621f-4c2b-8c7b-cc7d28b89921.mp4

?

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
```diff
@@ -0,0 +1,213 @@
...
+                    $"Monkey test \"{pageTestInfo.SanitizedUrl}\" within {pageTestInfo.TimeToTest.ToShortIntervalString()} with {randomSeed} random seed."),
+#pragma warning restore S103 // Lines should not be too long
+                () =>
+                {
+                    var pageTestTimeLeft = TestCurrentPageAndMeasureTestTimeLeft(pageTestInfo.TimeToTest, randomSeed);
+                    pageTestInfo.TimeToTest = pageTestTimeLeft;
+                    if (!_visitedPages.Contains(pageTestInfo)) _visitedPages.Add(pageTestInfo);
+                });
+
+            return _context.TriggerAfterPageChangeEventAsync();
```
Piedone, line 161: Control can arrive here before Gremlin finishes. I've seen this only on the frontend though (with 60s Gremlin time) but probably that's only the reason because accessibility checks happen just on the frontend (if turned on) and accessibility failing on the Gremlin overlay surfaced this.

image

Piedone (changes requested):

Lombiq.Tests.UI/MonkeyTesting/IMonkeyTestingUrlSanitizer.cs
YevgeniyShunevych: Renamed.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
YevgeniyShunevych: Updated XML comment.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: OK. I've added async method versions.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
YevgeniyShunevych: Changed and renamed.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: I updated monkey testing functionality to stop Gremlins script explicitly. It should solve that issue,

Lombiq.Tests.UI/Docs/Troubleshooting.md
YevgeniyShunevych: Try to view in slow motion? I don't know actually what we can do here, as it what Gremlins does. Any ideas?

Piedone: I don't know. Slowing down Gremlin so we can see what causes it, or logging the last interaction in this case if possible would work.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: Slowing down is possible by changing the value of `GremlinsAttackDelay` options property, which by default is 10ms.

Lombiq.Tests.UI/Docs/Troubleshooting.md
YevgeniyShunevych: Slowing down is possible by changing the value of `GremlinsAttackDelay` options property, which by default is 10ms.

Piedone: Document this then, please. And any way to log the interaction?

Lombiq.Tests.UI/MonkeyTesting/GremlinsScriptBuilder.cs
YevgeniyShunevych: I implemented in a way similar to Axe.Net by embedding `gremlins.min.js` script (which is loaded thru npm `package.json`) into `Lombiq.Tests.UI` assembly. Looks nice, as for me. Please check it out.

Lombiq.Tests.UI/EmbeddedResourceProvider.cs
```diff
@@ -0,0 +1,18 @@
+using System.IO;
+using System.Text;
+
+namespace Lombiq.Tests.UI
+{
+    internal static class EmbeddedResourceProvider
```
Piedone, line 6: Can't `ResourceHelper` from Helpful Libraries be used?

Lombiq.Tests.UI/MonkeyTesting/GremlinsScriptBuilder.cs
Piedone: Looks good, thanks.

Lombiq.Tests.UI/Lombiq.Tests.UI.csproj
```diff
@@ -79,4 +79,16 @@
     <ProjectReference Include="..\Lombiq.Tests.UI.Shortcuts\Lombiq.Tests.UI.Shortcuts.csproj" />
   </ItemGroup>
 
+  <Target Name="DownloadNpmPackages" Inputs="package.json" Outputs="node_modules\package.json" BeforeTargets="CollectPackageReferences">
```
Piedone: Great idea but please use [NPM Targets](https://github.com/Lombiq/NPM-Targets) for this instead.

YevgeniyShunevych: Changed to use `Lombiq.Npm.Targets`.

Lombiq.Tests.UI/EmbeddedResourceProvider.cs
YevgeniyShunevych: Unfortunately, `ResourceHelper` doesn't have such method that is needed.

Lombiq.Tests.UI/Docs/Troubleshooting.md
YevgeniyShunevych: I added a note on `GremlinsAttackDelay` property.

Regarding a way to log the interaction, I don't know. I don't see a simple solution here as we need to extent Gremlins.js probably. I would stop on the current point. Anyway we made too many things in scope of this PR.

Lombiq.Tests.UI/EmbeddedResourceProvider.cs
Piedone: Are you sure that `GetEmbeddedFileAsync<T>()` doesn't do the same in the end?

Lombiq.Tests.UI/EmbeddedResourceProvider.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        internal static string ReadEmbeddedFile(string fileName)
+        {
+            var assembly = typeof(EmbeddedResourceProvider).Assembly;
+            var resourceStream = assembly.GetManifestResourceStream($"{assembly.GetName().Name}.Resources.{fileName}");
+
+            using var reader = new StreamReader(resourceStream, Encoding.UTF8);
+
+            return reader.ReadToEnd();
```
Piedone, line 15: This can be async.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
Piedone: I don't see the log issue being fixed. The `TestContensAdminAsMonkeyRecursivelyShouldWorkWithAdminUser` I've temporarily re-added still passes though it should fail due to the 404.

Lombiq.Tests.UI/Docs/Troubleshooting.md
Piedone: Gremin supposedly logs to the browser console, which is what I'm looking for. However, this seems to be disabled for us somehow. See: https://github.com/marmelab/gremlins.js/issues/173

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
Piedone: This is still happening. I've added `TestEmptyPageAsMonkey` that demonstrates the issue and also fails because accessibility checking trips on the gremlin markers. Please delete it and `TestContensAdminAsMonkeyRecursivelyShouldWorkWithAdminUser` once you're done.

image

If you run the test, you'll get an accessibility report like this (it's HTML but GitHub doesn't let you upload that).
[AccessibilityReport.txt](https://github.com/Lombiq/UI-Testing-Toolbox/files/7850983/AccessibilityReport.txt)

Piedone (changes requested):
See comments.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: I've fixed that by cleaning Gremlins visual indicator elements. Check it out. Works for me fine.

Lombiq.Tests.UI/EmbeddedResourceProvider.cs
YevgeniyShunevych: Absolutely. Just check that method and its generic and regular argument.

Lombiq.Tests.UI/EmbeddedResourceProvider.cs
YevgeniyShunevych: Any reason for that? Its an internal class method that is used in one place. Check where it is used (in `Lazy`). I can make it async, but in its usage there will be typical `.GetAwaiter().GetResult()` added. What for?

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
YevgeniyShunevych: Sorry, I didn't get that you mean you want me to fix that log issue. I honestly don't understand description of that issue. But it seems like one outside of this PR task scope.

Lombiq.Tests.UI/Docs/Troubleshooting.md
YevgeniyShunevych: I'm sorry, I really don't have time to investigate and play with this feature. Anyway, sorry, but this feature was not planned in scope of this task.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
Piedone: Please push to the OSOCE repo so the build can run.

YevgeniyShunevych: Pushed. Let's wait and see.

Lombiq.Tests.UI/EmbeddedResourceProvider.cs
Piedone: I see, it would be complex to make it async all the way up. Let's leave it.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
YevgeniyShunevych: I fixed that issue the way you mentioned. Now `TestContensAdminAsMonkeyRecursivelyShouldWorkWithAdminUser` test fails as expected.

Lombiq.Tests.UI/Docs/Troubleshooting.md
YevgeniyShunevych: OK. It was not simple. But with a trick I found a solution. Basically, I write to log last Gremlins click action this way:

```
2022-01-22 16:51:34.5632  INFO - > Monkey test "/" within 7.347s with 2033103372 random seed.
2022-01-22 16:51:38.9286  INFO - - Last Gremlins click: click at 117 608.
2022-01-22 16:51:38.9287  INFO - < Monkey test "/" within 7.347s with 2033103372 random seed. (4.365s)
```

You can check it out. Let me know if this solution OK for you.

Lombiq.Tests.UI/MonkeyTesting/GremlinsScripts.cs
```diff
@@ -33,9 +42,22 @@ internal string Build()
...
 
                 string mogwaisPart = Mogwais != null ? string.Join(", ", Mogwais) : null;
-
                 return
+
 @$"(function() {{
+    const customLogger = {{
+        log: function () {{
+            console.log.apply(null, arguments);
+            if (['click', 'dblclick', 'mouseup'].some(x => arguments[2] === x)) {{
```
Piedone: Please use a proper variable name instead of `x`.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -149,18 +149,16 @@ private async ValueTask ShutdownAsync()
 
         private async Task<List<BrowserLogMessage>> GetBrowserLogAsync(RemoteWebDriver driver)
         {
-            if (_browserLogMessages != null) return _browserLogMessages;
-
-            var allMessages = new List<BrowserLogMessage>();
+            _browserLogMessages ??= new List<BrowserLogMessage>();
 
             foreach (var windowHandle in _context.Driver.WindowHandles)
```
Piedone: This causes an exception for certain tests: https://ci.lombiq.com/buildConfiguration/OrchardExtensions_OSOCE_Developer_RunUiTests/922716?buildTab=overview&showRootCauses=false&expandBuildTestsSection=true&expandBuildChangesSection=true#%2FFailureDumps%2FBasicOrchardFeaturesShouldWork(browser_-Chrome)%2FAttempt%201%2FDebugInformation

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
Piedone: While `TestEmptyPageAsMonkey` passes, Gremlin interactions still go on when the even is triggered:
image

So if the timing happens to be different, the same issue will arise with accessibility checking, and for other event handlers it's otherwise an issue as well.

Lombiq.Tests.UI/Docs/Troubleshooting.md
Piedone: Looks good!

Lombiq.Tests.UI/MonkeyTesting/GremlinsScripts.cs
YevgeniyShunevych: Renamed.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: Fixed that in another way by waiting for Gremlins indicators to disappear. This seems like the only 100% working solution.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
YevgeniyShunevych: Firstly, that fix should not have been done in scope of this PR. As it turned out, browser logs assertions was almost not working. Only first chunk of log was asserted. We fixed that code, and some old tests became to fail, as they were implemented regardless of browser log errors, and that fix uncovered them.

I made a fix for `BasicOrchardFeaturesTests.BasicOrchardFeaturesShouldWork(browser: Chrome)`.

Here are the problems with rest of the tests:

- ﻿`BasicTests.ShortcutsShouldWork(browser: Chrome)` - navigates to page "Lombiq.OSOCE.JsonEditor.Sample" that is 404.
- `ErrorHandlingTests.ErrorOnLoadedPageShouldHaltTest(browser: Chrome)` - checks app logs but doesn't consider 404 error in browser log.

Please check out those tests and, if possible, please make correct modifications.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
Piedone: That does cover certain interactions but not these:
screen

You can see on a screenshot if you put e.g `_context.Scope.Driver.GetScreenshot().SaveAsFile("screen.png");` before this line.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
Piedone: This exception still happens, unreliably. [Here](https://ci.lombiq.com/buildConfiguration/OrchardExtensions_OSOCE_Developer_RunUiTests/926093?showRootCauses=false&expandBuildTestsSection=true&expandBuildChangesSection=true) it happens, the same app executed again [here](https://ci.lombiq.com/buildConfiguration/OrchardExtensions_OSOCE_Developer_RunUiTests/926186?showRootCauses=false) worked.

Also, the `TestContensAdminAsMonkeyRecursivelyShouldWorkWithAdminUser` should fail but if not for the `WebDriverException ` it most of the time doesn't (but sometimes does), probably more frequently not failing in headless mode.

Piedone (changes requested):
See comments.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
YevgeniyShunevych: I checked the logs. So basically it depends on randomness. Test fails when it gets to "/Admin/Contents/ContentItems&Options.SelectedContentType=Article" page, which is really 404. But in [your successful build](https://ci.lombiq.com/buildConfiguration/OrchardExtensions_OSOCE_Developer_RunUiTests/926186?showRootCauses=false) it didn't visit that page, but visited "/Admin/Contents/ContentItems/4xa8ds820dgqkvx6eqh0c14rwz/Edit" instead. So it seems like fine for me.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
YevgeniyShunevych: Agree, my mistake. I've updated Gremlins indicator selector. Should be fine now.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
Piedone: Shouldn't it be deterministic though? Since we hard-coded the seed. And the `WebDriverException` shouldn't happen at all.

YevgeniyShunevych: By design it should be deterministic. But unfortunately I also noticed that Gremlins can behave differently with the same seed. Sometimes it reproduces a previous run, but sometimes doesn't. I don't know what else we can do here, and should we. It's an internal logic of Gremlins which uses [Chance JS library](https://www.npmjs.com/package/chance) for randomization.

As I understand, running with a seed is needed to rerun a failed test, which is difficult to understand the reason, locally to reproduce a failure. For most failures a reason of failure should be clear by screenshots and logs. So it seems like a rare scenario for tests that have strange and unclear failure. Anyway, with few local runs you should be able to reproduce it.

Piedone: I've opened an issue about repeatability here, please chime in if necessary: https://github.com/marmelab/gremlins.js/issues/185

The `WebDriverException` issue remains though, as also visible in [the latest run](https://ci.lombiq.com/buildConfiguration/OrchardExtensions_OSOCE_Developer_RunUiTests/926713?showRootCauses=false&expandBuildChangesSection=true&expandBuildTestsSection=true).

Piedone: Please give them the information necessary for troubleshooting there.

YevgeniyShunevych: OK. Here are the findings on those 2 issues.

1. Gremlins randomness. I tried to reproduce our issue in laboratory conditions but failed, as that execution was deterministic. So I started to find what can be dynamic on our page under test that can make monkey testing execution differ. And I found a possible reason, which is the "Privacy consent" popup that slowly slides on a page from the right side of a screen when the page appears. image
When I added temporarily `Thread.Sleep(5000);` after page load and before Gremlins script execution, to make the page frozen (popup stopped animating), the test became deterministic. So this popup sometimes can get clicks and sometimes not, depending on milliseconds when Gremlins script started and popup started its movement. It makes the page test non-deterministic. Usually, anyway it is recommended to disable such popups during UI test automation, as they can make tests flaky. The same for our case. So here are the possible solutions that I see:
   a) To fix our randomness issue, we can disable that popup at all on all pages.
   b) Or leave it, but we should consider that monkey testing can be a bit random when using the same seed.
   c) Or add a handler on every page load that will wait for that popup to finish its animation.

2. `WebDriverException`. This error was uncovered after we fixed browser log caching. I spent hours debugging and finding the reason of that. So basically the reason was in `UITestExecutionSession` and failed test rerun. When a test fails and another attempt is taken, a new second `UITestExecutionSession` instance is created. But when we get to `AssertLogsAsync`, it was executed in the scope of the first `UITestExecutionSession` instance with old context and already disposed `Driver` instance. It happened because of `_configuration.Events.AfterPageChange += _ => AssertLogsAsync();` event subscription was executed in first `UITestExecutionSession` instance, and was not unsubscribed when that instance was disposed. So logs assertion was executed in the scope of wrong and disposed `UITestExecutionSession` instance. It may be easier to understand when you look at a fix I made in commit [d8a2b54843addead81eb40cbe82dcb9422011ff8](https://github.com/Lombiq/UI-Testing-Toolbox/pull/109/commits/d8a2b54843addead81eb40cbe82dcb9422011ff8). Maybe there is a better way to fix that, as you wrote that code, you can implement a different solution I guess. And just to confirm that this `WebDriverException` gone away, I added temp `ShouldFailWith404` test method, because `TestContensAdminAsMonkeyRecursivelyShouldWorkWithAdminUser` didn't fail on TeamCity twice in a row because of a randomness issue. Here is the latest CI test result without `WebDriverException`: <https://ci.lombiq.com/viewLog.html?buildId=934218&buildTypeId=OrchardExtensions_OSOCE_Developer_RunUiTests#testNameId7163670106898338870>.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTester.cs
Piedone: Looks good.

Piedone (approved):

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
Piedone: Thank you for investigating!

1. I removed that the privacy consent banner slides in. Now it's just there right away so this shouldn't be an issue anymore.
2. Ah, that was a tricky one. Tweaked it a bit, otherwise looks good.

PR #121, opened 2022-02-02
OSOE-65: Capture screenshot after each page load

### LINKED ISSUE #120: Capture screenshot after each page load
This would also help monkey testing.

Piedone:
OSOE-65

Fixes #120


Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -223,18 +235,28 @@ private async Task CreateFailureDumpAsync(Exception ex, string dumpRootPath, int
...
+                    _context.TakeScreenshot(imagePath);
 
                     if (_configuration.ReportTeamCityMetadata)
                     {
                         TeamCityMetadataReporter.ReportImage(_testManifest.Name, "Screenshot", imagePath);
                     }
+
+                    var pageScreenshotsPath = Path.Combine(debugInformationPath, "PageScreenshots");
+                    Directory.CreateDirectory(pageScreenshotsPath);
+                    var fileNameFormat = "D" + _pageScreenshots.Count.ToTechnicalString().Length;
```
sarahelsaig:
```suggestion
                    var fileNameFormat = "D" + ((int)Math.Log10(_pageScreenshots.Count) + 1);
```

This should be slightly faster and shorter, but more importantly I think it communicates the intention more clearly than a string conversion.

sarahelsaig (changes requested):

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
Piedone: I considered this but I've chosen the string length approach specifically because it's clearer to me :D. I'll ask The People.

Piedone: I've added an extension for this, based on Oliver's suggestion.

sarahelsaig: Exquisite

Lombiq.Tests.UI/Services/UITestContext.cs
```diff
@@ -103,7 +104,11 @@ internal async Task TriggerAfterPageChangeEventAsync()
             {
                 try
                 {
-                    await afterPageChange.Invoke(this);
+                    // For some reason doing await afterPageChange.Invoke(this) will case exceptions to not propagate
+                    // when there are more than 2 subscribers.
+                    await afterPageChange.GetInvocationList()
+                        .Cast<PageChangeEventHandler>()
+                        .AwaitEachAsync(eventHandler => eventHandler(this));
```
Piedone, line 111: Do you know why this was necessary, BTW, @DAud-IcI?

sarahelsaig: As I understand it's not actually safe to call `Invoke` on a multicast delegate and should be avoided. What multicast does, is calling each function in sequence and returning the last one's value. If you think about delegate in a sync context it just has methods that return a certain type so if you call `Invoke` synchronously all it does is start three threads and returns you the last one's `Task` object which the `await` correctly awaits. So all but the last one are essentially fire-and-forget event calls. 

Here is a little experiment to demonstrate the problem:
image

sarahelsaig: I think if we make an InvokeAsync extension method then AsnycFixer should complain about using Invoke. Do you want to do one or both of the following?
1) Expand scope here and make such an exception or create an issue to tackle it later.
2) Open an issue at AsyncFixer that it shouldn't allow Invoke on Task-returning delegates (at least if `+=` is ever used on them).

Piedone: Ah, that makes sense. I actually looked into what multicast delegates do and couldn't figure out the reason - in my understanding, async support is available but it's definite that what actually happens is what you said.

Perhaps we need to do something similar for other events in `UITestExecutionEvents` too. Though the others are actually used in a sync way: https://github.com/Lombiq/UI-Testing-Toolbox/blob/4ebe418a21450c1cf7711cf85bcbeb1eb0cce98e/Lombiq.Tests.UI/Extensions/NavigationWebElementExtensions.cs#L31

Hmm, AsyncFixer didn't complain here, what do you mean? I'd perhaps only add an `MulticastDelegate` extension to do what we have above.

sarahelsaig: In general, if you have a `SomeType.Method()` method and create a `MethodAsync(this SomeType sometype)` extension method then AsyncFixer will give you `error VSTHRD103: condition synchronously blocks. Await MethodAsync instead` whenever a `someType.Method()` call happens.

So if we'd create an InvokeAsync extension for Task returning delegates that would effectively deprecate the regular Invoke method. I realize this is not trivial, becuase you are using a custom delegate and not just `Task<T1,Task>` so this probably deserves its own issue.

Piedone: Yep. Could you please create one here? https://github.com/Lombiq/Helpful-Libraries/issues

sarahelsaig: https://github.com/Lombiq/Helpful-Libraries/issues/115

Piedone: Thanks!

PR #122, opened 2022-02-04
OSOE-67: Using new InvokeAsync() extension to safely invoke async MulticastDelegates

Piedone:
OSOE-67


Lombiq.Tests.UI/Services/UITestContext.cs
```diff
@@ -100,15 +99,11 @@ public async Task AssertBrowserLogAsync()
 
         internal async Task TriggerAfterPageChangeEventAsync()
         {
-            if (IsNoAlert() && Configuration.Events.AfterPageChange is { } afterPageChange)
```
Piedone, line 103: No need for the null check since `InvokeAsync()` does that.

PR #134, opened 2022-02-21
OSOE-79: Remove GetAwater(), use async instead

wAsnk:
OSOE-79

### DIFF
--- a/Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
@@ -192,37 +192,37 @@ namespace Lombiq.Tests.UI.Extensions
                 () => new WebDriverWait(context.Driver, TimeSpan.FromSeconds(10)).Until(
                     d => ((IJavaScriptExecutor)d).ExecuteScript("return document.readyState").Equals("complete")));
 
-        public static void SetTaxonomyFieldByIndex(this UITestContext context, string taxonomyId, int index)
+        public static Task SetTaxonomyFieldByIndexAsync(this UITestContext context, string taxonomyId, int index)
         {
             var baseSelector = FormattableString.Invariant($".tags[data-taxonomy-content-item-id='{taxonomyId}']");
-            SetFieldDropdownByIndex(context, baseSelector, index);
+            return SetFieldDropdownByIndexAsync(context, baseSelector, index);
         }
 
-        public static void SetContentPickerByIndex(this UITestContext context, string part, string field, int index)
+        public static Task SetContentPickerByIndexAsync(this UITestContext context, string part, string field, int index)
         {
             var baseSelector = FormattableString.Invariant($"*[data-part='{part}'][data-field='{field}']");
-            SetFieldDropdownByIndex(context, baseSelector, index);
+            return SetFieldDropdownByIndexAsync(context, baseSelector, index);
         }
 
-        private static void SetFieldDropdownByIndex(UITestContext context, string baseSelector, int index)
+        private static async Task SetFieldDropdownByIndexAsync(UITestContext context, string baseSelector, int index)
         {
             var byItem = By.CssSelector(FormattableString.Invariant(
                 $"{baseSelector} .multiselect__element:nth-child({index + 1}) .multiselect__option"));
 
             while (!context.Exists(byItem.Safely()))
             {
-                context.ClickReliablyOn(By.CssSelector(baseSelector + " .multiselect__select"));
+                await context.ClickReliablyOnAsync(By.CssSelector(baseSelector + " .multiselect__select"));
             }
 
-            context.ClickReliablyOn(byItem);
+            await context.ClickReliablyOnAsync(byItem);
         }
 
         /// <summary>
         /// A convenience method that merges <see cref="ElementRetrievalUITestContextExtensions.Get"/> and <see
-        /// cref="NavigationWebElementExtensions.ClickReliably(IWebElement,UITestContext)"/> so the <paramref
+        /// cref="NavigationWebElementExtensions.ClickReliablyAsync(IWebElement,UITestContext)"/> so the <paramref
         /// name="context"/> doesn't have to be passed twice.
         /// </summary>
-        public static void ClickReliablyOn(this UITestContext context, By by) => context.Get(by).ClickReliably(context);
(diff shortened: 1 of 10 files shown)


Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -192,37 +192,37 @@ public static async Task<T> GoToPageAsync<T>(this UITestContext context, string
...
         }
 
         /// <summary>
         /// A convenience method that merges <see cref="ElementRetrievalUITestContextExtensions.Get"/> and <see
-        /// cref="NavigationWebElementExtensions.ClickReliably(IWebElement,UITestContext)"/> so the <paramref
+        /// cref="NavigationWebElementExtensions.ClickReliablyAsync(IWebElement,UITestContext)"/> so the <paramref
         /// name="context"/> doesn't have to be passed twice.
         /// </summary>
-        public static void ClickReliablyOn(this UITestContext context, By by) => context.Get(by).ClickReliably(context);
+        public static Task ClickReliablyOnAsync(this UITestContext context, By by) => context.Get(by).ClickReliablyAsync(context);
```
Piedone, line 225: Look up the references of all the changed methods to see that no await was left out. `ClickReliablyOnSubmit` remained sync.

Lombiq.Tests.UI/Extensions/OrchardCoreDashboardUITestContextExtensions.cs
```diff
@@ -8,27 +8,27 @@ namespace Lombiq.Tests.UI.Extensions
     public static class OrchardCoreDashboardUITestContextExtensions
     {
         public static void FillContentItemTitle(this UITestContext context, string title) =>
-            context.ClickAndFillInWithRetries(By.Id("TitlePart_Title"), title);
+            context.ClickAndFillInWithRetriesAsync(By.Id("TitlePart_Title"), title);
 
         public static void GoToEditorTab(this UITestContext context, string tabText) =>
-            context.ClickReliablyOn(By.XPath($"//*[text()='{tabText}' and @class='nav-item nav-link']"));
+            context.ClickReliablyOnAsync(By.XPath($"//*[text()='{tabText}' and @class='nav-item nav-link']"));
```
Piedone: These two methods remained sync.

Lombiq.Tests.UI/Extensions/OrchardCoreDashboardUITestContextExtensions.cs
```diff
@@ -8,27 +8,27 @@ namespace Lombiq.Tests.UI.Extensions
...
-                context.ClickReliablyOn(By.Name("submit.Publish"));
+                await context.ClickReliablyOnAsync(By.Name("submit.Publish"));
             }
         }
 
         /// <summary>
         /// Clicks on the "Ok" button on the Bootstrap modal window.
         /// </summary>
-        public static void ClickModalOk(this UITestContext context) => context.ClickReliablyOn(By.Id("modalOkButton"));
+        public static void ClickModalOk(this UITestContext context) => context.ClickReliablyOnAsync(By.Id("modalOkButton"));
```
Piedone: Remained sync.

Lombiq.Tests.UI/Extensions/ReliabilityUITestContextExtensions.cs
```diff
@@ -137,21 +138,21 @@ public static class ReliabilityUITestContextExtensions
         /// <param name="timeout">Timeout of the operation.</param>
         /// <param name="interval">Time between retries.</param>
         /// <param name="existsTimeout">Timeout of checking the existence of the given element.</param>
-        public static void DoWithRetriesUntilMissing(
+        public static Task DoWithRetriesUntilMissingAsync(
             this UITestContext context,
             Action process,
```
Piedone: This should also be `Func<Task>`.

Lombiq.Tests.UI/Helpers/ReliabilityHelper.cs
```diff
@@ -198,5 +250,19 @@ public static void DoWithRetriesOrFail(Func<bool> process, TimeSpan? timeout = n
...
+            TimeSpan? timeout = null,
+            TimeSpan? interval = null)
+        {
+            var wait = new SafeWait<object>(new object());
+
+            // If no values are supplied then the defaults specified in AtataFactory will be used.
+            if (timeout != null) wait.Timeout = timeout.Value;
+            if (interval != null) wait.PollingInterval = interval.Value;
+
+            return (await wait.Until(async _ => await processAsync()), wait);
```
Piedone: `Until()` doesn't actually support async, it'll treat the `Task` return value as any other value.

Lombiq.Tests.UI/Extensions/ReliabilityUITestContextExtensions.cs
```diff
@@ -27,15 +28,15 @@ public static class ReliabilityUITestContextExtensions
...
             TimeSpan? timeout = null,
             TimeSpan? interval = null) =>
-            ReliabilityHelper.DoWithRetriesOrFail(
-                process,
-                timeout ?? context.Configuration.TimeoutConfiguration.RetryTimeout,
-                interval ?? context.Configuration.TimeoutConfiguration.RetryInterval);
+                ReliabilityHelper.DoWithRetriesOrFailAsync(
+                    processAsync,
+                    timeout ?? context.Configuration.TimeoutConfiguration.RetryTimeout,
+                    interval ?? context.Configuration.TimeoutConfiguration.RetryInterval);
```
Piedone: Here and in `RetryIfStaleOrFailAsync` the indentation was correct.

Piedone (changes requested):

Lombiq.Tests.UI/Helpers/ReliabilityHelper.cs
wAsnk: Ahh, but this is somewhat different, bcs the SafeWait.Until repeatedly does the process until it times out or if it runs OK.
So here we should implement an async version of it, not sure if that is already done somewhere else as it doesn't seem to be trivial.

Piedone: Check if anybody has done this. Otherwise, it seems pretty simple, just change the method to be async, the parameter to `Func<T, Task<TResult>> condition`, and add an await here:
image

PR #150, opened 2022-04-05
SNOW-50: Add GoToContentItemEditorByIdAsync extension.

sarahelsaig:
[SNOW-50](https://lombiq.atlassian.net/browse/SNOW-50)

### DIFF
--- a/Lombiq.Tests.UI/Extensions/ElementRetrievalUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/ElementRetrievalUITestContextExtensions.cs
@@ -38,6 +38,16 @@ public static class ElementRetrievalUITestContextExtensions
         context.ExecuteLogged(nameof(GetAll), by, () => context.CreateSearchContext().FindElements(by));
 
     /// <summary>
+    /// Retrieves all the matching elements with retries within the configured timeout. At least 1 item must exist
+    /// otherwise it will throw <see cref="NotFoundException"/>.
+    /// </summary>
+    public static ReadOnlyCollection<IWebElement> GetAllWhenOneExists(this UITestContext context, By by)
+    {
+        context.Exists(by);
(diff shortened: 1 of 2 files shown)


Lombiq.Tests.UI/Extensions/ElementRetrievalUITestContextExtensions.cs
```diff
@@ -37,6 +37,16 @@ public static class ElementRetrievalUITestContextExtensions
     public static ReadOnlyCollection<IWebElement> GetAll(this UITestContext context, By by) =>
         context.ExecuteLogged(nameof(GetAll), by, () => context.CreateSearchContext().FindElements(by));
 
+    /// <summary>
+    /// Retrieves all the matching elements with retries within the configured timeout. At least 1 item must exist
+    /// otherwise it will throw <see cref="NotFoundException"/>.
+    /// </summary>
+    public static ReadOnlyCollection<IWebElement> GetAllWhenOneExists(this UITestContext context, By by)
+    {
+        context.Exists(by);
```
barthamark, line 46: Is this one that has the retry logic? Because I don't see another retry here. Wanted to ask if it wouldn't be better to just check the count of the `GetAll` result but I guess it's because that doesn't have retry and it'll return inconsistent numbers.

sarahelsaig: No explicit retry logic. I was referring to the default behavior of `.Exists`. Its docstring says "Checks the existence of the element **with retries within the configured timeout**." so I copied that over for documentation uniformity.

The problem this solves is that GetAll doesn't wait for the elements to appear (e.g. with dynamic/async content). If you know there should be some elements and you want to get all of them then you need to get the first via Exists to know when it's safe to get all.

barthamark (approved):

PR #151, opened 2022-04-06
OSOE-97: Maximizing concurrent tests running

wAsnk:
OSOE-97

### DIFF
--- a/Lombiq.Tests.UI/Docs/Configuration.md
+++ b/Lombiq.Tests.UI/Docs/Configuration.md
@@ -56,7 +56,7 @@ Note that this will execute tests in headless mode, so no browser windows will b
 
 We encourage you to experiment with a `RetryTimeoutSeconds` value suitable for your hardware. Higher, paradoxically, is usually less safe.
 
-`MaxRunningConcurrentTests` sets how many tests should run at the same time. Use a value of `0` to indicate that you would like the default behavior. Use a value of `-1` to indicate that you do not wish to limit the number of tests running at the same time. The default behaviour and `0` uses the [System.Environment.ProcessorCount](https://docs.microsoft.com/en-us/dotnet/api/system.environment.processorcount?view=net-6.0#remarks) property. Set any other positive integer to limit to the exact number.
+`MaxRunningConcurrentTests` sets how many UI tests should run at the same time. It is an important property if you want to run your UI tests parallel, checkout the inline documentation in [`OrchardCoreUITestExecutorConfiguration`](../Services/OrchardCoreUITestExecutorConfiguration.cs).
 
 ## <a name="multi-process"></a>Multi-process test execution
 
--- a/Lombiq.Tests.UI/Docs/Configuration.md
+++ b/Lombiq.Tests.UI/Docs/Configuration.md
@@ -56,7 +56,7 @@ Note that this will execute tests in headless mode, so no browser windows will b
 
 We encourage you to experiment with a `RetryTimeoutSeconds` value suitable for your hardware. Higher, paradoxically, is usually less safe.
 
-`MaxRunningConcurrentTests` sets how many UI tests should run at the same time. It is an important property if you want to run your UI tests parallel, checkout the inline documentation in [`OrchardCoreUITestExecutorConfiguration`](../Services/OrchardCoreUITestExecutorConfiguration.cs).
+`MaxRunningConcurrentTests` sets how many UI tests should run at the same time. It is an important property if you want to run your UI tests in parallel, check out the inline documentation in [`OrchardCoreUITestExecutorConfiguration`](../Services/OrchardCoreUITestExecutorConfiguration.cs).
 
 ## <a name="multi-process"></a>Multi-process test execution
 
--- a/Lombiq.Tests.UI/Docs/Configuration.md
+++ b/Lombiq.Tests.UI/Docs/Configuration.md
@@ -42,7 +42,7 @@ Here's a full *TestConfiguration.json* file example, something appropriate durin
     "OrchardCoreUITestExecutorConfiguration": {
       "MaxRetryCount": 0,
       "RetryIntervalSeconds": 0,
-      "MaxParrallelTests": 0
+      "MaxParallelTests": 0
     },
     "BrowserConfiguration": {
       "Headless": true
@@ -56,7 +56,7 @@ Note that this will execute tests in headless mode, so no browser windows will b
 
 We encourage you to experiment with a `RetryTimeoutSeconds` value suitable for your hardware. Higher, paradoxically, is usually less safe.
 
-`MaxParrallelTests` sets how many UI tests should run at the same time. It is an important property if you want to run your UI tests in parallel, check out the inline documentation in [`OrchardCoreUITestExecutorConfiguration`](../Services/OrchardCoreUITestExecutorConfiguration.cs).
+`MaxParallelTests` sets how many UI tests should run at the same time. It is an important property if you want to run your UI tests in parallel, check out the inline documentation in [`OrchardCoreUITestExecutorConfiguration`](../Services/OrchardCoreUITestExecutorConfiguration.cs).
 
 ## <a name="multi-process"></a>Multi-process test execution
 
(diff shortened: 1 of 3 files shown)


Lombiq.Tests.UI/Docs/Configuration.md
```diff
@@ -55,6 +56,7 @@ Note that this will execute tests in headless mode, so no browser windows will b
 
 We encourage you to experiment with a `RetryTimeoutSeconds` value suitable for your hardware. Higher, paradoxically, is usually less safe.
 
+`MaxRunningConcurrentTests` sets how many tests should run at the same time. Use a value of `0` to indicate that you would like the default behavior. Use a value of `-1` to indicate that you do not wish to limit the number of tests running at the same time. The default behaviour and `0` uses the [System.Environment.ProcessorCount](https://docs.microsoft.com/en-us/dotnet/api/system.environment.processorcount?view=net-6.0#remarks) property. Set any other positive integer to limit to the exact number.
```
Piedone: This doc is better suited as docs on the property. Also explain the relationship with the xUnit config there (as you mentioned under the issue: Why do we need this? Do we need this only for UI test? Why are UI tests unique in that matter?). Nevertheless, emphasize it being an important config here.

Lombiq.Tests.UI/Services/UITestExecutor.cs
```diff
@@ -4,13 +4,17 @@
...
+using System.Threading;
 using System.Threading.Tasks;
 using Xunit.Abstractions;
 
 namespace Lombiq.Tests.UI.Services;
 
 public static class UITestExecutor
 {
+    private static readonly object _initLock = new();
+    private static SemaphoreSlim _semaphoreSlim;
```
Piedone: Name these more expressively, since we're in a large class.

Lombiq.Tests.UI/Services/UITestExecutor.cs
```diff
@@ -51,12 +55,25 @@ private static async Task ExecuteOrchardCoreTestInnerAsync(UITestManifest testMa
...
+            }
+        }
+
         var retryCount = 0;
         var passed = false;
         while (!passed)
         {
             try
             {
+                if (_semaphoreSlim != null)
```
Piedone: This can never be true here though, right? On line 92 too.

Piedone (changes requested):

Lombiq.Tests.UI/Services/UITestExecutor.cs
wAsnk: _semaphoreslim will be set on line 62. So it won't be null, which means there will be a limit and `await _semaphoreSlim.WaitAsync();` will be run. 

It will be null only when you want to set it to unlimited number of concurrent tests with passing `-1` to `MaxRunningConcurrentTests`.

Piedone: Right.

Lombiq.Tests.UI/Docs/Configuration.md
```diff
@@ -56,7 +56,7 @@ Note that this will execute tests in headless mode, so no browser windows will b
 
 We encourage you to experiment with a `RetryTimeoutSeconds` value suitable for your hardware. Higher, paradoxically, is usually less safe.
 
-`MaxRunningConcurrentTests` sets how many tests should run at the same time. Use a value of `0` to indicate that you would like the default behavior. Use a value of `-1` to indicate that you do not wish to limit the number of tests running at the same time. The default behaviour and `0` uses the [System.Environment.ProcessorCount](https://docs.microsoft.com/en-us/dotnet/api/system.environment.processorcount?view=net-6.0#remarks) property. Set any other positive integer to limit to the exact number.
+`MaxRunningConcurrentTests` sets how many UI tests should run at the same time. It is an important property if you want to run your UI tests parallel, checkout the inline documentation in [`OrchardCoreUITestExecutorConfiguration`](../Services/OrchardCoreUITestExecutorConfiguration.cs).
```
Piedone:
```suggestion
`MaxRunningConcurrentTests` sets how many UI tests should run at the same time. It is an important property if you want to run your UI tests in parallel, check out the inline documentation in [`OrchardCoreUITestExecutorConfiguration`](../Services/OrchardCoreUITestExecutorConfiguration.cs).
```

Lombiq.Tests.UI/Services/OrchardCoreUITestExecutorConfiguration.cs
```diff
@@ -34,6 +34,22 @@ public class OrchardCoreUITestExecutorConfiguration
             $"{nameof(OrchardCoreUITestExecutorConfiguration)}:RetryIntervalSeconds",
             0));
 
+    /// <summary>
+    /// Gets or sets how many tests should run at the same time. Use a value of '0' to indicate that you would like the
+    /// default behavior. Use a value of '-1' to indicate that you do not wish to limit the number of tests running at
+    /// the same time. The default behaviour and '0' uses the <see cref="Environment.ProcessorCount"/> property. Set any
```
Piedone: 0 and 1 are not characters. Also, use American English.

Lombiq.Tests.UI/Services/OrchardCoreUITestExecutorConfiguration.cs
```diff
@@ -34,6 +34,22 @@ public class OrchardCoreUITestExecutorConfiguration
...
+    /// other positive integer to limit to the exact number.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// The XUnit MaxParrallelThreads property controls only the threads, not the actual processes started. See <see
+    /// href="https://github.com/xunit/xunit/issues/2003"></see>.
+    /// </para>
+    /// <para>
+    /// This is important only for UI tests as there will be a running instance of the site for each UI test, which can
+    /// cause performance issues, like out of memory.
```
Piedone:
```suggestion
    /// cause performance issues, like running out of memory.
```

Lombiq.Tests.UI/Services/OrchardCoreUITestExecutorConfiguration.cs
```diff
@@ -34,6 +34,22 @@ public class OrchardCoreUITestExecutorConfiguration
...
+    /// </para>
+    /// </remarks>
     public int MaxRunningConcurrentTests { get; set; } =
```
Piedone:
```suggestion
    public int MaxParrallelTests { get; set; } =
```

Lombiq.Tests.UI/Services/UITestExecutor.cs
```diff
@@ -12,8 +12,8 @@ namespace Lombiq.Tests.UI.Services;
 
 public static class UITestExecutor
 {
-    private static readonly object _initLock = new();
-    private static SemaphoreSlim _semaphoreSlim;
+    private static readonly object _lockObjectOfSemaphoreSlimInitialization = new();
```
Piedone: We know it's an object, no need for Hungarian notation. But this doesn't convey where do we use this lock (there can be any semaphore here).

Piedone (changes requested):

Lombiq.Tests.UI/Services/UITestExecutor.cs
wAsnk: What name do you suggest?

Piedone: I was always happy with Zoltán.

Lombiq.Tests.UI/Services/OrchardCoreUITestExecutorConfiguration.cs
```diff
@@ -25,10 +25,36 @@ public class OrchardCoreUITestExecutorConfiguration
...
+
+    /// <summary>
+    /// Gets or sets how many tests should run at the same time. Use a value of 0 to indicate that you would like the
+    /// default behavior. Use a value of -1 to indicate that you do not wish to limit the number of tests running at the
+    /// same time. The default behavior and 0 uses the <see cref="Environment.ProcessorCount"/> property. Set any other
+    /// positive integer to limit to the exact number.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// The XUnit MaxParrallelThreads property controls only the threads, not the actual processes started. See <see
```
Piedone: Also fix the same typo in `MaxParrallelTests`.

```suggestion
    /// The XUnit MaxParallelThreads property controls only the threads, not the actual processes started. See <see
```

Piedone (changes requested):

Piedone (approved):

PR #152, opened 2022-04-07
OSOE-98: Save per-page screenshots to files directly to conserve memory

Piedone:
OSOE-98


Lombiq.Tests.UI/Constants/Paths.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.IO;
+
+namespace Lombiq.Tests.UI.Constants;
+
+public static class Paths
+{
+    public const string DefaultSetupSnapshotDirectoryPath = "SetupSnapshot";
+    public const string TempDirectoryPath = "Temp";
+    public const string ScreenshotsDirectoryName = "Screenshots";
```
sarahelsaig: Two are called *DirectoryPath one is called *DirectoryName. If this is a meaningful distinction, please comment.

By the way if you renamed the class to `DirectoryPaths` then you wouldn't need to prefix the variables and you could use `nameof`.

Lombiq.Tests.UI/Constants/Paths.cs
```diff
@@ -0,0 +1,17 @@
...
+
+    public static string GetTempSubDirectoryPath(string contextId, string subDirectoryName) =>
```
sarahelsaig: I see you pass `string.Empty` to this [below](https://github.com/Lombiq/UI-Testing-Toolbox/pull/152/files#diff-7296bb52a59df5fa30fd8222aa55a27fb459709e7e0cb6539be92a6ac2606069R159). Would it be nicer to make `""` the default value of the `subDirectoryName` argument? Or better yet, make it `params string[] subDirectoryNames`?

sarahelsaig (changes requested):

Lombiq.Tests.UI/Constants/Paths.cs
Piedone: How they're used there can be a difference, but let's not have that.

Lombiq.Tests.UI/Constants/DirectoryPaths.cs
```diff
@@ -9,8 +10,8 @@ public static class DirectoryPaths
     public const string Temp = nameof(Temp);
     public const string Screenshots = nameof(Screenshots);
 
-    public static string GetTempSubDirectoryPath(string contextId, string subDirectoryName) =>
-        Path.Combine(Environment.CurrentDirectory, Temp, contextId, subDirectoryName);
+    public static string GetTempSubDirectoryPath(string contextId, params string[] subDirectoryNames) =>
+        Path.Combine(new[] { Environment.CurrentDirectory, Temp, contextId }.Union(subDirectoryNames).ToArray());
```
sarahelsaig: `.Union()` is a set operator. Only use it if you want to treat your collection as a set. In this case use `.Concat()`.
image

Piedone: Right.

PR #153, opened 2022-04-09
SANTA-6: Refactor monkey testing to stay on localhost sites

DemeSzabolcs:
SANTA-6

OSOCE: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/88

### DIFF
--- a/Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
+++ b/Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
@@ -48,7 +48,7 @@ public sealed class MonkeyTestingOptions
     {
         @"gremlins.species.clicker({
             log: true,
-            canClick: (element) => /https:\/\/localhost:\d\d\d\d.*/.test(element.href))}",
+            canClick: (element) => /https:\/\/localhost:\d\d\d\d.*/.test(element.href)})",
 
         "gremlins.species.toucher()",
         "gremlins.species.formFiller()",
--- a/Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
+++ b/Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
@@ -46,7 +46,10 @@ public sealed class MonkeyTestingOptions
     /// </summary>
     public List<string> GremlinsSpecies { get; } = new()
     {
-        "gremlins.species.clicker({ log: true })",
+        @"gremlins.species.clicker({
+            log: true,
+            canClick: (element) => element.getAttribute(""href"").includes(""localhost:"")})",
+
         "gremlins.species.toucher()",
         "gremlins.species.formFiller()",
         "gremlins.species.scroller()",
--- a/Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
+++ b/Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
@@ -48,7 +48,8 @@ public sealed class MonkeyTestingOptions
     {
         @"gremlins.species.clicker({
             log: true,
-            canClick: (element) => /https:\/\/localhost:\d\d\d\d.*/.test(element.href)})",
+            canClick: (element) => /https:\/\/localhost:\d\d\d\d.*/.test(element.href) ||
+                !element.hasAttribute('href')})",
 
         "gremlins.species.toucher()",
         "gremlins.species.formFiller()",
--- a/Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
+++ b/Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
@@ -50,7 +50,8 @@ public sealed class MonkeyTestingOptions
     {
         @"gremlins.species.clicker({
             log: true,
-            canClick: (element) => /https:\/\/localhost:\d\d\d\d.*/.test(element.href) ||
+            canClick: (element) =>
+                /https:\/\/localhost:\d\d\d\d.*/.test(element.href) ||
                 !element.hasAttribute('href')})",
 
         "gremlins.species.toucher()",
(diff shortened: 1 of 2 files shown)


Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
```diff
@@ -46,7 +46,10 @@ public sealed class MonkeyTestingOptions
     /// </summary>
     public List<string> GremlinsSpecies { get; } = new()
     {
-        "gremlins.species.clicker({ log: true })",
+        @"gremlins.species.clicker({
+            log: true,
+            canClick: (element) => /https:\/\/localhost:\d\d\d\d.*/.test(element.href)})",
```
Piedone: Update the docs above too (with a verbal description of what's happening instead of copying the whole code).

Piedone: This doesn't look like it's working correctly. Contrast what `TestAdminPagesAsMonkeyRecursivelyShouldWorkWithAdminUser` does in `dev` with what it does here: Here, it doesn't leave the starting page and runs for 40s on my machine, the `dev` one visits a lot of admin pages and runs for 2,5 mins.

Piedone (changes requested):

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
```diff
@@ -48,7 +51,15 @@ public sealed class MonkeyTestingOptions
     {
         @"gremlins.species.clicker({
             log: true,
-            canClick: (element) => /https:\/\/localhost:\d\d\d\d.*/.test(element.href)})",
+            canClick: (element) => {
+                    for (; element && element !== document; element = element.parentNode) {
+                        if (!(/https:\/\/localhost:\d\d\d\d.*/.test(element.href) || !element.hasAttribute('href'))) {
```
Piedone: Wouldn't switching the order of the clauses make this faster, since the element not having an `href` is an excluding test anyway? Or rather, doesn't this fail with elements not having an `href`?

Or if the point is to find the first element with an `href` (or reach the root) then I don't understand why the second clause is necessary.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
```diff
@@ -48,7 +51,15 @@ public sealed class MonkeyTestingOptions
...
+                            return false;
+                        }
+                    }
+
+                    return true;
+                }})",
```
Piedone: This causes a notable different execution compared to `dev` too, this one visits a lot more pages (e.g. in `TestAdminPagesAsMonkeyRecursivelyShouldWorkWithAdminUser`). This is not an issue per see but please investigate what's the reason behind it, since `dev` doesn't click on any non-localhost links either.

Piedone (changes requested):

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
DemeSzabolcs: The point is, to check the current element and its parents.
If an element has no `href` it can click it. If an element has a `href` it must be localhost. If there is no `href` it will not fail, because it will be undefined and it doesn't match the pattern. Then it will check if it's empty, and it will be clicked because it is. So yeah, as you said it's better to switch the order of the statements. 

The reason why we are checking parents, is because for example in this case: 
```
<a href="external site">
   <span>"Hello"</span>
</a>
```
`<a>` will be not clicked, but the `<span>` will be and it activates the link. 

So yeah, the only to-do here is to switch the order of the statements.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
Piedone: What did you find?

Piedone (changes requested):
See comment.

Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
DemeSzabolcs: I'm still investigating, I just put it to "Done", because everything else is done.

Piedone: So the investigation is not done, the issue shouldn't be either :).

DemeSzabolcs: Is there any evidence (I mean evidence as data) about it visiting significantly more pages? Because I ran the `dev` one and `SANTA-6` one, and they did the same thing. Both of them ran for between 1,5-2,5 mins. I think it's just pure luck because it's random.

Even the CI history has consistent time values and they are the same in `SANTA-6` and `dev`.

Piedone: I've watched the two running side by side a couple of times and this one always took longer with more pages. Running headless in the background always took longer too.

I tried again with a deliberate exception at the end, just to get screenshots for all pages, and obviously now I observe the exact opposite, `dev` visiting more pages. After multiple tries, I did get back to this visiting more (like 38 vs 104) but yeah, this seems to be just random. (Gremlin actions aren't supposed to be, but timings of the app responding vary and can cause this.)

```
    [Theory, Chrome]
    public Task TestAdminPagesAsMonkeyRecursivelyShouldWorkWithAdminUser(Browser browser) =>
        ExecuteTestAfterSetupAsync(
            async context =>
            {
                // Monkey tests needn't all start from the homepage. This one starts from the Orchard admin dashboard.
                await context.TestAdminAsMonkeyRecursivelyAsync(CreateMonkeyTestingOptions());
                throw new Exception("hehh");
            },
            browser,
            configuration =>
                // This is necessary to work around this bug: https://github.com/OrchardCMS/OrchardCore/issues/11420.
                configuration.AssertBrowserLog = messages => messages.ShouldNotContain(
                    message => IsValidAdminBrowserLogMessage(message),
                    messages.Where(IsValidAdminBrowserLogMessage).ToFormattedString()));
```

Piedone (approved):

PR #156, opened 2022-04-23
OSOE-91: Add gitattributes

sarahelsaig:
[OSOE-91](https://lombiq.atlassian.net/browse/OSOE-91)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,12 +1,10 @@
 .vs/
+.idea/
+.vscode/
 obj/
 bin/
-*.user
 artifacts/
-.vs/
-wwwroot
+wwwroot/
 node_modules/
+*.user
 .pnpm-debug.log
-package-lock.json
-Localization/
-/.editorconfig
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,12 +1,12 @@
...
+wwwroot/
 node_modules/
-.pnpm-debug.log
-package-lock.json
 Localization/
-/.editorconfig
\ No newline at end of file
+*.user
+.pnpm-debug.log
+/.editorconfig
```
0liver: I've restored `Localization/` and `/.editorconfig` here.

0liver (approved):

.gitignore
sarahelsaig: Are they needed though? I assumed they are outdated, because I did a rebuild and nothing materialized.

0liver: Well, I take your word for it. I'll remove them then.

0liver: Is the deletion in the other two submodules also intended then?

sarahelsaig: Yes.

0liver: Reverted my commits.

PR #158, opened 2022-05-02
OSOE-77: Adding UITestContext extension to take screenshot of an element

dministro:
OSOE-77

### DIFF
--- a/Lombiq.Tests.UI/Extensions/ScreenshotUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/ScreenshotUITestContextExtensions.cs
@@ -1,5 +1,7 @@
 using Lombiq.Tests.UI.Services;
 using OpenQA.Selenium;
+using System.Drawing;
+using System.IO;
 
 namespace Lombiq.Tests.UI.Extensions;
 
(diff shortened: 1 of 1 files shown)


Lombiq.Tests.UI/Extensions/ScreenshotUITestContextExtensions.cs
```diff
@@ -16,4 +18,16 @@ public static class ScreenshotUITestContextExtensions
...
         context.Scope.Driver.GetScreenshot();
+
+    /// <summary>
+    /// Takes a screenshot of an element.
+    /// </summary>
+    public static Bitmap TakeScreenshotImage(this UITestContext context, IWebElement element)
+    {
+        var screen = context.TakeScreenshot();
+        using var screenRaw = new MemoryStream(screen.AsByteArray);
+        using var screenImage = Image.FromStream(screenRaw) as Bitmap;
```
sarahelsaig, line 29: Use cast instead of `as`, becuase you actually require the `Bitmap` type for this code to work.

The `as` operator should only be used is the type not matching is acceptable, and in that case the result must be null-checked (e.g. `screenImage?.Clone`). When you expect a specific type for sure, casting is appropriate becuase not matching that type is indeed an exception.

sarahelsaig (changes requested):

PR #159, opened 2022-05-02
KARGO-11: Fix bug that breaks ExecuteTestFromExistingDBAsync

DemeSzabolcs:
KARGO-11


Lombiq.Tests.UI.Samples/Tests/BasicOrchardFeaturesTests.cs
```diff
@@ -13,18 +14,40 @@ namespace Lombiq.Tests.UI.Samples.Tests;
...
         ExecuteTestAsync(
-            context => context.TestBasicOrchardFeaturesAsync("Lombiq.OSOCE.BasicOrchardFeaturesTests"),
+            context => context.TestBasicOrchardFeaturesAsync(BasicOrchardFeaturesTestsRecipeId),
             browser);
+
+    // For testing, we can use the already existing databases. Here, we set up the application, then we take a snapshot
+    // of it, then we use the "ExecuteTestFromExistingDBAsync()" to run the test on that. Then we test the basic Orchard
+    // features as we did above.
+    [Theory, Chrome]
+    public Task BasicOrchardFeaturesShouldWorkWithExistingDBSetup(Browser browser) =>
```
DemeSzabolcs: I put this test here (`BasicOrchardFeaturesTests.cs`), because I'm using the `TestBasicOrchardFeaturesExceptSetupAsync()` method inside the test, so I thought this would fit here. 

I'm using that method because we need some tests to do inside `ExecuteTestFromExistingDBAsync()`, and it's better to check that everything works correctly from the previously created snapshot.

Piedone: It would be better in a new class, inserted into the series of tutorials, because using an existing DB snapshot is not really something most users would need.

DemeSzabolcs: Okay, I will move it then.

Lombiq.Tests.UI.Samples/Constants/RecipeIds.cs
```diff
@@ -0,0 +1,5 @@
+namespace Lombiq.Tests.UI.Samples.Constants;
+public static class RecipeIds
```
Piedone:
```suggestion
namespace Lombiq.Tests.UI.Samples.Constants;

public static class RecipeIds
```

Lombiq.Tests.UI/OrchardCoreUITestBase.cs
```diff
@@ -145,6 +145,17 @@ public abstract class OrchardCoreUITestBase
...
+        ExecuteTestFromExistingDBAsync(testAsync, browser, customAppAssemblyPath: null, changeConfigurationAsync);
+
+    /// <summary>
+    /// Executes the given UI test, starting the app from an existing SQLite database available in the App_Data or in
+    /// the given folder.
+    /// </summary>
+    protected virtual Task ExecuteTestFromExistingDBAsync(
+        Func<UITestContext, Task> testAsync,
+        Browser browser,
+        string customAppAssemblyPath = null,
```
Piedone: If we're talking about a snapshot then it won't contain the assembly.

```suggestion
        string customAppFolderPath = null,
```

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -40,6 +40,7 @@ internal sealed class UITestExecutionSession : IAsyncDisposable
     private SynchronizingWebApplicationSnapshotManager _currentSetupSnapshotManager;
     private string _snapshotDirectoryPath;
     private bool _hasSetupOperation;
+    private bool _IsDBSetupSnapshotDirectoryPath;
```
Piedone:
```suggestion
    private bool _isDBSetupSnapshotDirectoryPath;
```

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -68,6 +69,8 @@ public async Task<bool> ExecuteAsync(int retryCount, string dumpRootPath)
             var setupConfiguration = _configuration.SetupConfiguration;
             _hasSetupOperation = setupConfiguration.SetupOperation != null;
 
+            _IsDBSetupSnapshotDirectoryPath = setupConfiguration.SetupSnapshotDirectoryPath == "AppFolder";
```
Piedone: This shouldn't depend on that ad-hoc folder name from the other class, or use a new field. Instead, `ExecuteTestFromExistingDBAsync` should configure the snapshot path.

Lombiq.Tests.UI.Samples/Tests/DataBaseSnapshotTests.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.IO;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Lombiq.Tests.UI.Samples.Tests;
+
+// We can execute test on already existing databases. To test this, we will take a snapshot of the running application,
+// and use that.
+public class DataBaseSnapshotTests : UITestBase
```
Piedone: Also add to the Readme.

Lombiq.Tests.UI.Samples/Tests/DataBaseSnapshotTests.cs
```diff
@@ -0,0 +1,42 @@
...
+using Lombiq.Tests.UI.Services;
+using System.IO;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Lombiq.Tests.UI.Samples.Tests;
+
+// We can execute test on already existing databases. To test this, we will take a snapshot of the running application,
+// and use that.
```
Piedone:
```suggestion
// We can execute tests on already existing databases. To demo this, we will take a snapshot of the running application,
// and use that.
```

Lombiq.Tests.UI.Samples/Tests/DataBaseSnapshotTests.cs
```diff
@@ -0,0 +1,42 @@
...
+public class DataBaseSnapshotTests : UITestBase
+{
+    public DataBaseSnapshotTests(ITestOutputHelper testOutputHelper)
```
Piedone: Don't forget to rename in `MonkeyTests` and the other instances here too.

```suggestion
    public DatabaseSnapshotTests(ITestOutputHelper testOutputHelper)
```

Lombiq.Tests.UI.Samples/Tests/DataBaseSnapshotTests.cs
```diff
@@ -0,0 +1,42 @@
...
+        : base(testOutputHelper)
+    {
+    }
+
+    // Here, we set up the application, then we take a snapshot of it, then we use the
+    // "ExecuteTestFromExistingDBAsync()" to run the test on that. Then we test the basic Orchard features to check that
+    // the application was set up correctly.
```
Piedone:
```suggestion
    // "ExecuteTestFromExistingDBAsync()" to run the test on that. Finally, we test the basic Orchard features to check
    // that the application was set up correctly.
```

Lombiq.Tests.UI.Samples/Tests/DataBaseSnapshotTests.cs
Piedone: Also explain why we take a snapshot and use that, and what would you use this feature for normally (which isn't usually to start from a snapshot taken during a test).

Lombiq.Tests.UI.Samples/Tests/DataBaseSnapshotTests.cs
```diff
@@ -0,0 +1,42 @@
...
+    [Theory, Chrome]
+    public Task BasicOrchardFeaturesShouldWorkWithExistingDBSetup(Browser browser) =>
```
Piedone:
```suggestion
    public Task BasicOrchardFeaturesShouldWorkWithExistingDatabase(Browser browser) =>
```

Lombiq.Tests.UI.Samples/Tests/DataBaseSnapshotTests.cs
```diff
@@ -0,0 +1,42 @@
...
+                {
+                    const string AppForDataBaseTestFolder = "AppForDataBaseTest";
+
+                    await context.GoToSetupPageAndSetupOrchardCoreAsync(RecipeIds.BasicOrchardFeaturesTests);
+                    await context.Application.TakeSnapshotAsync(AppForDataBaseTestFolder);
+
+                    await ExecuteTestFromExistingDBAsync(
+                         async context => await context.TestBasicOrchardFeaturesExceptSetupAsync(),
+                         browser,
+                         Directory.GetCurrentDirectory() + "//" + AppForDataBaseTestFolder);
```
Piedone: Use the const here too, it should be (made) usable just as for taking the snapshot.

Piedone (changes requested):

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
DemeSzabolcs: `ExecuteTestFromExistingDBAsync` is already configuring that. The problem is that there is a case, when there is no setup operation, and there is a setup snapshot directory path. But it's not from `ExecuteTestFromExistingDBAsync` (it's not `AppFolder`, it's `Temp\SetupSnapshot`). And if I let it set the `OrchardCoreConfiguration.SnapshotDirectoryPath` to that parameter, it breaks things (like setup recipes won't appear on the setup page). This happens when we a run a normal `ExecuteTestAsync`.

So in some way we need to check the path. I could do what I did above, or the other solution could be checking if there is a setup snapshot directory path and checking that it's not `Temp\SetupSnaphot`.

Piedone: You can change the logic here to load the snapshot if the folder exists and there's no setup operation.

DemeSzabolcs: You mean folder as "AppFolder" right? Because if we check if the `setupConfiguration.SetupSnapshotDirectoryPath`  exists it will be true with `Temp\SetupSnapshot` too, so it will cause the problem that I mentioned above.

Piedone: I meant `SetupSnapshotDirectoryPath`. You can also check if it's not empty.

DemeSzabolcs: `Temp\SetupSnapshot` is not empty it has a subfolder, so that's not enough. So I think  `Directory.Exists(setupConfiguration.SetupSnapshotDirectoryPath + "\\App_Data")` would be ok. What do you think? (I mean it's working that way, I'm just asking if it's okay. To avoid review ping-ponging)

Piedone: Actually, loading the DB as a snapshot can be a setup operation, no? It's also applicable semantically And then no such check is necessary.

If not, what you've written sounds OK too.

Lombiq.Tests.UI.Samples/Tests/DataBaseSnapshotTests.cs
DemeSzabolcs: Sorry, I don't understand this one, which const do you mean?

Piedone: `AppForDataBaseTestFolder`

DemeSzabolcs: But I'm already using it, or should it be modified in some way?
image

Piedone: I mean, use only the const, without `Directory.GetCurrentDirectory() + "//"`: `TakeSnapshotAsync` should work in the same as `ExecuteTestFromExistingDBAsync`.

Lombiq.Tests.UI.Samples/Constants/RecipeIds.cs
Piedone: Please commit such suggestions directly so they needn't be manually checked and resolved.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -69,7 +69,7 @@ public async Task<bool> ExecuteAsync(int retryCount, string dumpRootPath)
             var setupConfiguration = _configuration.SetupConfiguration;
             _hasSetupOperation = setupConfiguration.SetupOperation != null;
 
-            _IsDBSetupSnapshotDirectoryPath = setupConfiguration.SetupSnapshotDirectoryPath == "AppFolder";
+            _doesSetupSnapshotDirectoryExists = Directory.Exists(setupConfiguration.SetupSnapshotDirectoryPath + "\\App_Data");
```
Piedone: Use `Path.Combine()` instead of manually concating paths.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
Piedone: So, it can't be a setup operation?

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -40,7 +40,7 @@ internal sealed class UITestExecutionSession : IAsyncDisposable
     private SynchronizingWebApplicationSnapshotManager _currentSetupSnapshotManager;
     private string _snapshotDirectoryPath;
     private bool _hasSetupOperation;
-    private bool _IsDBSetupSnapshotDirectoryPath;
+    private bool _doesSetupSnapshotDirectoryExists;
```
Piedone:
```suggestion
    private bool _setupSnapshotDirectoryExists;
```

Lombiq.Tests.UI.Samples/Tests/DataBaseSnapshotTests.cs
Piedone: No "Setup" suffix, it's not a "database setup" that should work with OC features, but a database.

Lombiq.Tests.UI.Samples/Tests/DataBaseSnapshotTests.cs
```diff
@@ -2,39 +2,40 @@
...
 
                     await context.GoToSetupPageAndSetupOrchardCoreAsync(RecipeIds.BasicOrchardFeaturesTests);
-                    await context.Application.TakeSnapshotAsync(AppForDataBaseTestFolder);
+                    await context.Application.TakeSnapshotAsync(AppForDatabaseTestFolder);
 
                     await ExecuteTestFromExistingDBAsync(
                          async context => await context.TestBasicOrchardFeaturesExceptSetupAsync(),
                          browser,
-                         Directory.GetCurrentDirectory() + "//" + AppForDataBaseTestFolder);
+                         AppForDatabaseTestFolder);
```
Piedone: Does this work the same as `Directory.GetCurrentDirectory() + "//" + AppForDataBaseTestFolder` before though? I'm not saying you can just use `AppForDatabaseTestFolder` here without any other changes. I'm saying, the API of `ExecuteTestFromExistingDBAsync` should allow using this relative path, instead of having to prefix is.

Piedone (changes requested):

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
DemeSzabolcs: Yeah, if it has setup operation it does things that would break the test in case of a custom snapshot (like it combines setup snapshot directory path, with other things),

Lombiq.Tests.UI.Samples/Tests/DataBaseSnapshotTests.cs
DemeSzabolcs: Yes, because it operates in the same directory, so `Directory.GetCurrentDirectory()` was not necessary. A good example for this is:
https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/OrchardCoreUITestBase.cs#L158-L160

Where `appFolder` is not a path just the name of the folder.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -94,6 +98,13 @@ public async Task<bool> ExecuteAsync(int retryCount, string dumpRootPath)
 
                 await SetupAsync();
             }
+            else
+            {
+                if (_setupSnapshotDirectoryExists)
+                {
+                    _configuration.OrchardCoreConfiguration.SnapshotDirectoryPath = setupConfiguration.SetupSnapshotDirectoryPath;
+                }
+            }
```
Piedone: Merge to an `else if`.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -94,6 +98,13 @@ public async Task<bool> ExecuteAsync(int retryCount, string dumpRootPath)
 
                 await SetupAsync();
             }
+            else
+            {
+                if (_setupSnapshotDirectoryExists)
```
Piedone: Add a comment about when it can happen that there's a snapshot but not a setup operation.

Lombiq.Tests.UI.Samples/Tests/DataBaseSnapshotTests.cs
```diff
@@ -0,0 +1,43 @@
...
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Lombiq.Tests.UI.Samples.Tests;
+
+// We can execute tests on already existing databases. To demo this, we will take a snapshot of the running application,
+// and use that, because we don't have a pre-beaked database. Normally this feature is used to run tests when there is
+// an already existing database, so you don't need to take a snapshot before using the
+// "ExecuteTestFromExistingDBAsync()" method.
+public class DatabaseSnapshotTests : UITestBase
```
Piedone: The file name remained wrong.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
Piedone: OK then!

Lombiq.Tests.UI.Samples/Tests/DataBaseSnapshotTests.cs
Piedone: OK then.

Lombiq.Tests.UI/OrchardCoreUITestBase.cs
```diff
@@ -145,6 +145,17 @@ public abstract class OrchardCoreUITestBase
...
+        ExecuteTestFromExistingDBAsync(testAsync, browser, customAppFolderPath: null, changeConfigurationAsync);
+
+    /// <summary>
+    /// Executes the given UI test, starting the app from an existing SQLite database available in the App_Data or in
+    /// the given folder.
+    /// </summary>
+    protected virtual Task ExecuteTestFromExistingDBAsync(
+        Func<UITestContext, Task> testAsync,
+        Browser browser,
+        string customAppFolderPath = null,
```
Piedone: I noticed that this clashes with `appFolder`.

```suggestion
        string customSnapshotFolderPath = null,
```

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -40,6 +40,7 @@ internal sealed class UITestExecutionSession : IAsyncDisposable
     private SynchronizingWebApplicationSnapshotManager _currentSetupSnapshotManager;
     private string _snapshotDirectoryPath;
     private bool _hasSetupOperation;
+    private bool _setupSnapshotDirectoryExists;
```
Piedone:
```suggestion
    private bool _setupSnapshotDirectoryContainsApp;
```

Lombiq.Tests.UI.Samples/Tests/DataBaseSnapshotTests.cs
DemeSzabolcs: Well, that is strange: 
image

For some reason GitExtensions didn't detected the file name change (I guess because it was just casing). Fixed now.

Piedone (approved):

PR #167, opened 2022-05-24
OSOE-49: Refactoring suggestion

0liver:
OSOE-49 review suggestion

### DIFF
--- a/Lombiq.Tests.UI/Services/WebDriverFactory.cs
+++ b/Lombiq.Tests.UI/Services/WebDriverFactory.cs
@@ -22,32 +22,32 @@ public static class WebDriverFactory
     {
         ChromeDriver CreateDriverInner(ChromeDriverService service)
         {
-            var config = new ChromeConfiguration { Options = new ChromeOptions().SetCommonOptions() };
+            var chromeConfig = new ChromeConfiguration { Options = new ChromeOptions().SetCommonOptions() };
 
-            config.Options.AddArgument("--lang=" + configuration.AcceptLanguage);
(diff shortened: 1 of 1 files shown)


Lombiq.Tests.UI/Services/WebDriverFactory.cs
```diff
@@ -20,76 +20,73 @@ public static class WebDriverFactory
...
             // However, this makes the executing machine vulnerable to browser-based attacks so it should only be used
             // with trusted code (like our own).
             config.Options.AddArgument("no-sandbox");
 
             // Linux-specific setting, may be necessary for running in containers, see
-            // https://developers.google.com/web/tools/puppeteer/troubleshooting#tips
+            // https://developers.google.com/web/tools/puppeteer/troubleshooting#tips for more information.
             config.Options.AddArgument("disable-dev-shm-usage");
 
             if (configuration.Headless) config.Options.AddArgument("headless");
```
sarahelsaig: Originally I didn't dare to name it `config` because there is already a configuration and assumed that would be confusing.
Now it's going to be your fault. 😈

0liver: Renaming them right now!

0liver: Done.

0liver: Named it `chromeConfig`.

0liver: I still feel a bit guilty of using an abbreviation 🤔 now that you've brought up the topic.

PR #169, opened 2022-05-27
SNOW-109: Fix visibility with element lookup.

sarahelsaig:
[SNOW-109](https://lombiq.atlassian.net/browse/SNOW-109)


Psichorex (approved):

Lombiq.Tests.UI/Extensions/NavigationWebElementExtensions.cs
```diff
@@ -32,7 +32,22 @@ public static class NavigationWebElementExtensions
                     await context.Configuration.Events.BeforeClick
                         .InvokeAsync<ClickEventHandler>(eventHandler => eventHandler(context, element));
 
-                    context.Driver.Perform(actions => actions.MoveToElement(element).Click());
+                    // When the button is under some overhanging UI element, the MoveToElement sometimes fails with the
+                    // "move target out of bounds" exception message. In this case it should be retried.
+                    const int maxTries = 3;
+                    var notFound = true;
+                    for (var i = 1; notFound && i <= maxTries; i++)
```
Psichorex, line 39: Is there any particular reason using var instead of int in the for loop or it's just a matter of taste?

sarahelsaig: In some scenarios the code analyzer throws `IDE0007: use 'var' instead of explicit type` if you don't use `var` so (at least for me) it's easier to just be consistent and always use `var`. That way I don't have to think about it ever.

In this particular case it's apparently allowed (I don't know what are the criteria), but if you do a solution-wide search for `for (var` and `for (int`, the former is much more common than the latter.

Psichorex: Yes it's all clear now it was rather a question out of curiosity as when I learned it years ago cases where there couldn't be any other type used just one e.g. int as an iterator in a loop we should use the exact type. But I did do a little research if theres any principal about this and there isnt.

PR #119, opened 2022-01-22
OSOE-49: Replace Windows-only code in UI Testing Toolbox, upgrade to latest Atata/Selenium

sarahelsaig:
[OSOE-49](https://lombiq.atlassian.net/browse/OSOE-49)

Fixes #123.

## Release Notes

- Please add these new lines to your _.gitattributes_, because we get IDE0055 warnings when the C# source file has Unix style line endings. Some .cs files will be affected, please run `git add --renormalize .` on your projects and `git submodule foreach git add --renormalize .` if they include submodules.
- The setup and configuration for using SQL Server from a Docker container has changed. See new instructions [here](https://github.com/Lombiq/UI-Testing-Toolbox/blob/e764b5c3cb4e40d3a9043b520e71d0c5c44173f7/Lombiq.Tests.UI/Docs/Configuration.md#using-sql-server-from-a-docker-container). You can also find a sample config and setup steps in the _.github_ directory.
- We have upgraded to Atata 2 which in turn upgraded to Selenium 4. Both introduce breaking changes, although most of it is pretty easy to fix. Replace every `RemoteWebDriver` with `IWebDriver` and if you execute scripts or take screenshots then use the extension methods ([here](https://github.com/Lombiq/UI-Testing-Toolbox/blob/e764b5c3cb4e40d3a9043b520e71d0c5c44173f7/Lombiq.Tests.UI/Extensions/ScreenshotUITestContextExtensions.cs), [here](https://github.com/Lombiq/UI-Testing-Toolbox/blob/e764b5c3cb4e40d3a9043b520e71d0c5c44173f7/Lombiq.Tests.UI/Extensions/ScriptingUITestContextExtensions.cs) and [here](https://github.com/Lombiq/UI-Testing-Toolbox/blob/e764b5c3cb4e40d3a9043b520e71d0c5c44173f7/Lombiq.Tests.UI/Extensions/ScriptingWebDriverExtensions.cs)).
- If you have several UI test projects it can be cumbersome to maintain a _TestConfiguration.json_ file for each. Instead, you can set the value of the `LOMBIQ_UI_TESTING_TOOLBOX_SHARED_TEST_CONFIGURATION` environment variable to the absolute path of a central configuration file and then each project will look it up. If you place an individual _TestConfiguration.json_ into a test directory it will still take precedence in case you need special configuration for just that one.
- New _HelpfulLibraries.Cli_ project has been added, please use and expand it for all your `Process` and `CliWrap` needs.

PR #170, opened 2022-05-27
OSOE-123: Clean up and simplify OrchardCoreUITestBase

0liver:
OSOE-123

### DIFF
--- a/Lombiq.Tests.UI/Delegates/MultiSizeTest.cs
+++ b/Lombiq.Tests.UI/Delegates/MultiSizeTest.cs
@@ -1,5 +1,4 @@
 using Lombiq.Tests.UI.Services;
-using System.Diagnostics.CodeAnalysis;
 using System.Threading.Tasks;
 
 namespace Lombiq.Tests.UI.Delegates;
@@ -20,12 +19,7 @@ public delegate Task MultiSizeTestAsync(UITestContext context, bool isStandardSi
 
 public static class MultiSizeTestExtensions
 {
-    [SuppressMessage(
-        "Minor Code Smell",
-        "S4261:Methods should be named according to their synchronicities",
-        Justification =
-            "This is a conversion method, and we want to make it explicit that the result has an async signature.")]
-    public static MultiSizeTestAsync ToAsync(this MultiSizeTest test) =>
+    public static MultiSizeTestAsync AsCompletedTask(this MultiSizeTest test) =>
         (context, isStandardSize) =>
         {
             test(context, isStandardSize);
--- a/Lombiq.Tests.UI/Delegates/MultiSizeTest.cs
+++ b/Lombiq.Tests.UI/Delegates/MultiSizeTest.cs
@@ -1,4 +1,5 @@
 using Lombiq.Tests.UI.Services;
+using System.Diagnostics.CodeAnalysis;
 using System.Threading.Tasks;
 
 namespace Lombiq.Tests.UI.Delegates;
@@ -16,3 +17,18 @@ public delegate void MultiSizeTest(UITestContext context, bool isStandardSize);
 
 /// <inheritdoc cref="MultiSizeTest"/>
 public delegate Task MultiSizeTestAsync(UITestContext context, bool isStandardSize);
+
+public static class MultiSizeTestExtensions
+{
+    [SuppressMessage(
+        "Minor Code Smell",
+        "S4261:Methods should be named according to their synchronicities",
+        Justification =
+            "This is a conversion method, and we want to make it explicit that the result has an async signature.")]
+    public static MultiSizeTestAsync ToAsync(this MultiSizeTest test) =>
+        (context, isStandardSize) =>
+        {
+            test(context, isStandardSize);
+            return Task.CompletedTask;
+        };
+}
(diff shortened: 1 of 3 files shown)


Lombiq.Tests.UI/Delegates/MultiSizeTest.cs
```diff
@@ -16,3 +17,18 @@ namespace Lombiq.Tests.UI.Delegates;
...
 public delegate Task MultiSizeTestAsync(UITestContext context, bool isStandardSize);
+
+public static class MultiSizeTestExtensions
+{
+    [SuppressMessage(
+        "Minor Code Smell",
+        "S4261:Methods should be named according to their synchronicities",
+        Justification =
+            "This is a conversion method, and we want to make it explicit that the result has an async signature.")]
+    public static MultiSizeTestAsync ToAsync(this MultiSizeTest test) =>
```
sarahelsaig: I think it's still misleading becuase you'd assume this method does something similar to `Task.Run()` instead of just returning `Task.CompletedTask`. So to me this hurts more than it helps. How about one of these mouthfuls?
- `ReturnCompletedTask`
- `WithCompletedTask`
- `ToMultiSizeTestAsyncWithCompletedTask`

Lombiq.Tests.UI/Extensions/AsyncExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Lombiq.Tests.UI.Extensions;
+
+public static class AsyncExtensions
+{
+    [SuppressMessage(
+        "Minor Code Smell",
+        "S4261:Methods should be named according to their synchronicities",
+        Justification =
+            "This is a conversion method, and we want to make it explicit that the result has an async signature.")]
+    public static Func<T, Task> ToAsync<T>(this Action<T> action) =>
```
sarahelsaig: Same.

Lombiq.Tests.UI/OrchardCoreUITestBase.cs
```diff
@@ -13,49 +13,40 @@ namespace Lombiq.Tests.UI;
 
 public abstract class OrchardCoreUITestBase
 {
+    private const string AppFolder = "AppFolder";
```
sarahelsaig: Is there a reason this isn't using `nameof`?

Lombiq.Tests.UI/OrchardCoreUITestBase.cs
```diff
@@ -13,49 +13,40 @@ namespace Lombiq.Tests.UI;
...
 
     private static bool _appFolderCreated;
 
+    protected abstract string AppAssemblyPath { get; }
+
     protected virtual Size StandardBrowserSize => CommonDisplayResolutions.Standard;
     protected virtual Size MobileBrowserSize => CommonDisplayResolutions.NhdPortrait;
 
-    protected abstract string AppAssemblyPath { get; }
+    static OrchardCoreUITestBase() => AtataFactory.SetupShellCliCommandFactory();
```
sarahelsaig, line 29: No, this will break customizability. Pasting what I wrote [here](https://github.com/Lombiq/UI-Testing-Toolbox/pull/119#discussion_r874210773):

> We can't. Static constructors are called when the type is first referenced. So if we want to let the library consumer customize it in the `OrchardCoreUITestBase` overload as described in _Linux.md_, then we must be absolutely certain that this code has run before the setup customization. In general you should only use static constructor to set data that will be immediately interacted with. Every test calls to `OrchardCoreUITestBase` very early so that's perfect.
In this case I have verified that `AtataFactory`'s static constructor actually executes later than e.g. the configuration in `Lombiq.OSOCE.Tests.UI.UITestBase.ExecuteTestAsync`.

sarahelsaig (changes requested):

Lombiq.Tests.UI/Delegates/MultiSizeTest.cs
0liver: Maybe `.AsCompletedTask()`?

Lombiq.Tests.UI/OrchardCoreUITestBase.cs
0liver: Only that it used to be a method-local constant named `appFolder`. After moving it to the top of the file it became this - and I didn't catch it.

Lombiq.Tests.UI/Delegates/MultiSizeTest.cs
sarahelsaig: That's fine too.

0liver: I'm thinking that the `Completed` infix is a lot of implementation noise that shouldn't be there, don't you think? Considering just `.AsTask()` now.

sarahelsaig: I think whether a task will actually start a thread or not is useful information, not noise.

Lombiq.Tests.UI/OrchardCoreUITestBase.cs
0liver: I (R#) just moved this line from the bottom of the file up here.

Looks to me like this is what we have after you did [this](https://github.com/Lombiq/UI-Testing-Toolbox/pull/119#discussion_r874224878):
> encapsulate this into AtataFactory so we don't have basic Atata configuration in multiple places

Shouldn't that be good now?

sarahelsaig: Never mind, I'm blind

PR #166, opened 2022-05-24
OSOE-111: Detect visual errors (visual verification testing) in the UI Testing Toolbox

dministro:
Fixes GH-85

OSOE-111

PR #172, opened 2022-05-28
OSOE-113: Annotate GitHub repo files with warnings in GitHub workflows

sarahelsaig:
OSOE-113


Lombiq.Tests.UI/OrchardCoreUITestBase.cs
```diff
@@ -218,6 +218,10 @@ public abstract class OrchardCoreUITestBase
     {
         var testManifest = new UITestManifest(_testOutputHelper) { TestAsync = testAsync };
 
+        var originalTestOutputHelper = _testOutputHelper;
+        (_testOutputHelper, var afterTest) =
+            GitHubActionsGroupingTestOutputHelper.CreateWrapper(_testOutputHelper, testManifest);
```
Piedone, line 223: Keep in mind that GH Actions is just one CI system that these tests might be run under, so it shouldn't be used all the time. See how `TeamCityMetadataReporter` is used.

Lombiq.Tests.UI/Services/GitHubActionsGroupingTestOutputHelper.cs
```diff
@@ -0,0 +1,65 @@
+﻿using Lombiq.Tests.UI.Models;
+using System;
+using Xunit.Abstractions;
+
+namespace Lombiq.Tests.UI.Services;
+
+internal sealed class GitHubActionsGroupingTestOutputHelper : ITestOutputHelper
```
Piedone, line 7: Document how this can be utilized, similar to TeamCity metadata reporting, both [in the wiki]https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2455732262/Continuous+Integration) and in the Readme.

Lombiq.Tests.UI/OrchardCoreUITestBase.cs
sarahelsaig: If you open the method, when `GitHubActionsGroupingTestOutputHelper.IsGitHubEnvironment.Value` is `false` (in other words when `$GITHUB_ENV` is null or empty) this method returns `_testOutputHelper`, so nothing is changed. This way the logic is centralized.

Lombiq.Tests.UI/Services/GitHubActionsGroupingTestOutputHelper.cs
sarahelsaig: I'm not entirely sure what you have in mind, but I've added an entry to the wiki and mentioned them in the main and UI testing readmes. As mentioned in the other point it's environmentally activated so nothing in particular needs to be done to utilize it.

Lombiq.Tests.UI/Services/GitHubAnnotationWriter.cs
```diff
@@ -0,0 +1,75 @@
...
+    private readonly ITestOutputHelper _testOutputHelper;
+
+    public GitHubAnnotationWriter(ITestOutputHelper testOutputHelper) => _testOutputHelper = testOutputHelper;
+
+    public void Annotate(LogLevel severity, string title, string message, string file, int line = 1)
+    {
+        ArgumentNullException.ThrowIfNull(message);
+        ArgumentNullException.ThrowIfNull(file);
+
+        if (file.Contains(',')) throw new ArgumentException("File name mustn't contain commas.", nameof(file));
```
Piedone, line 22: That's a suspiciously specific check. Why is it necessary? Document it inline.

Lombiq.Tests.UI/Services/GitHubAnnotationWriter.cs
```diff
@@ -0,0 +1,75 @@
...
+        title = title.Replace(',', '⹁').Replace("::", "⸬");
+
+        // Sanitize message:
+        message = message.Replace("\r", string.Empty).Replace("\n", " ");
+
+        _testOutputHelper.WriteLine(FormattableString.Invariant(
+            $"::{command} file={file},line={line},title={title}::{message}"));
+    }
+
+    public void ErrorInTest(Exception exception, ITestCase testCase)
```
Piedone, line 51: What Jocó has done under https://github.com/Lombiq/UI-Testing-Toolbox/pull/166 related to call stacks might be interesting here. Just food for thought.

Lombiq.Tests.UI/Services/SqlServerManager.cs
```diff
@@ -122,7 +120,7 @@ public SqlServerRunningContext CreateDatabase()
         var filePathLocal = GetSnapshotFilePath(snapshotDirectoryPathLocal ?? snapshotDirectoryPathRemote);
         var directoryPathLocal =
             Path.GetDirectoryName(filePathLocal) ??
-            throw new InvalidOperandException($"Failed to get the directory path for local path \"{filePathLocal}\".");
```
Piedone, line 125: What the heck was this? :D

Lombiq.Tests.UI/Services/WebDriverFactory.cs
```diff
@@ -48,6 +48,7 @@ ChromeDriver CreateDriverInner(ChromeDriverService service)
             configuration.BrowserOptionsConfigurator?.Invoke(chromeConfig.Options);
 
             chromeConfig.Service = service ?? ChromeDriverService.CreateDefaultService();
+            chromeConfig.Service.SuppressInitialDiagnosticInformation = true;
```
Piedone, line 51: Isn't this available for the other drivers?

Piedone (changes requested):

Lombiq.Tests.UI/Services/GitHubAnnotationWriter.cs
sarahelsaig: I intentionally want to pick out the stack frame in the test class, because that's likely to be in the PR's changed files, and because it's the last specific part to that test case.

Lombiq.Tests.UI/Services/SqlServerManager.cs
sarahelsaig: Some SQL related type mismatch exception. It keeps getting suggested when I type "invalidop", if the `Microsoft.SqlServer.Management.Dmf` namespace is already imported.

Lombiq.Tests.UI/Services/WebDriverFactory.cs
sarahelsaig: I've found it for Edge, but not for Fx or IE.

Piedone (approved):

PR #173, opened 2022-05-30
SNOW-72: Restoring the Atata 1.x style default element lookup behavior

Psichorex:
[SNOW-72](https://lombiq.atlassian.net/browse/SNOW-72)

only find visible elements


Piedone:
Use a more descriptive PR title.

PR #182, opened 2022-07-01
OSOE-111: Adding extension to set the inline style of an element

dministro:
OSOE-111

### DIFF
--- a/Lombiq.Tests.UI/Extensions/ElementStyleUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/ElementStyleUITestContextExtensions.cs
@@ -6,7 +6,7 @@ namespace Lombiq.Tests.UI.Extensions;
 public static class ElementStyleUITestContextExtensions
 {
     /// <summary>
-    /// Sets the element inline style.
+    /// Sets the element's inline style.
     /// </summary>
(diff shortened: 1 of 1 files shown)


Lombiq.Tests.UI/Extensions/ElementStyleUITestContextExtensions.cs
```diff
@@ -0,0 +1,21 @@
+using Lombiq.Tests.UI.Services;
+using OpenQA.Selenium;
+
+namespace Lombiq.Tests.UI.Extensions;
+
+public static class ElementStyleUITestContextExtensions
+{
+    /// <summary>
+    /// Sets the element inline style.
```
Piedone:
```suggestion
    /// Sets the element's inline style.
```

Lombiq.Tests.UI/Extensions/ElementStyleUITestContextExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// </summary>
+    /// <param name="context">The <see cref="UITestContext"/> in which the extension is executed on.</param>
```
Piedone: Documenting the `this` parameter is not necessary.

Piedone (changes requested):
Use a descriptive PR title.

Piedone:
Fixed these.

PR #194, opened 2022-08-21
WALMA-19: Extend UI Testing Toolbox to support custom admin url prefixes

Psichorex:
WALMA-19

### DIFF
--- a/Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
@@ -53,14 +53,19 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// specified <paramref name="setupRecipeId"/>.
     /// </para>
     /// <para>The test method assumes that the site is not set up.</para>
+    /// <para>When running headless version of Orchard Core, ContentOperations shall be excluded.
+    /// Utilize <paramref name="dontCheckFrontend"></paramref>> for this purpose.</para>
     /// </summary>
     /// <param name="setupRecipeId">The ID of the recipe to be used to set up the site.</param>
+    /// <param name="dontCheckFrontend">Boolean to decide whether ContentOperations shall be executed.</param>>
     /// <returns>The same <see cref="UITestContext"/> instance.</returns>
-    public static Task TestBasicOrchardFeaturesExceptRegistrationAsync(this UITestContext context, string setupRecipeId) =>
-        context.TestBasicOrchardFeaturesExceptRegistrationAsync(new OrchardCoreSetupParameters(context)
-        {
-            RecipeId = setupRecipeId,
-        });
+    public static Task TestBasicOrchardFeaturesExceptRegistrationAsync(
+        this UITestContext context,
+        string setupRecipeId,
+        bool dontCheckFrontend = false) =>
+        context.TestBasicOrchardFeaturesExceptRegistrationAsync(
+            dontCheckFrontend,
+            new OrchardCoreSetupParameters(context) { RecipeId = setupRecipeId, });
 
     /// <summary>
     /// <para>
@@ -69,64 +74,19 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// <c>"SaaS"</c><see cref="OrchardCoreSetupParameters.RecipeId"/> value.
     /// </para>
     /// <para>The test method assumes that the site is not set up.</para>
+    /// <para>When running headless version of Orchard Core, ContentOperations shall be excluded.
+    /// Utilize <paramref name="dontCheckFrontend"></paramref>> for this purpose.</para>
     /// </summary>
     /// <param name="setupParameters">The setup parameters.</param>
+    /// <param name="dontCheckFrontend">Boolean to decide whether ContentOperations shall be executed.</param>>
     /// <returns>The same <see cref="UITestContext"/> instance.</returns>
     public static async Task TestBasicOrchardFeaturesExceptRegistrationAsync(
         this UITestContext context,
+        bool dontCheckFrontend = false,
         OrchardCoreSetupParameters setupParameters = null)
     {
         await context.TestSetupWithInvalidAndValidDataAsync(setupParameters);
-        await context.TestBasicOrchardFeaturesExceptSetupAndRegistrationAsync();
-    }
-
-    /// <summary>
-    /// <para>
-    /// Tests all the basic Orchard features except for registration and content operations. At first sets up Orchard
-    /// with the recipe with the specified <paramref name="setupRecipeId"/>.
-    /// </para>
-    /// <para>The test method assumes that the site is not set up.</para>
-    /// </summary>
-    /// <param name="setupRecipeId">The ID of the recipe to be used to set up the site.</param>
-    /// <returns>The same <see cref="UITestContext"/> instance.</returns>
-    public static Task TestBasicOrchardFeaturesExceptRegistrationInHeadlessAppAsync(
-        this UITestContext context,
-        string setupRecipeId) =>
-        context.TestBasicOrchardFeaturesExceptRegistrationInHeadlessAppAsync(new OrchardCoreSetupParameters(context)
-        {
-            RecipeId = setupRecipeId,
-        });
-
-    /// <summary>
-    /// <para>
-    /// Tests all the basic Orchard features except for registration and content operations. At first sets up Orchard
-    /// with the recipe with the specified <paramref name="setupParameters"/>. By default uses new
-    /// <see cref="OrchardCoreSetupParameters"/> instance with <c>"SaaS"</c><see cref="OrchardCoreSetupParameters.RecipeId"/> value.
-    /// </para>
-    /// <para>The test method assumes that the site is not set up.</para>
-    /// </summary>
-    /// <param name="setupParameters">The ID of the recipe to be used to set up the site.</param>
-    /// <returns>The same <see cref="UITestContext"/> instance.</returns>
-    public static async Task TestBasicOrchardFeaturesExceptRegistrationInHeadlessAppAsync(
-        this UITestContext context,
-        OrchardCoreSetupParameters setupParameters = null)
-    {
-        await context.TestSetupWithInvalidAndValidDataAsync(setupParameters);
-        await context.TestBasicOrchardFeaturesExceptSetupAndRegistrationInHeadlessAppAsync();
-    }
-
-    /// <summary>
-    /// <para>Tests all the basic Orchard features except for setup, registration and content operations.</para>
-    /// <para>The test method assumes that the site is set up.</para>
-    /// </summary>
-    /// <returns>The same <see cref="UITestContext"/> instance.</returns>
-    public static async Task TestBasicOrchardFeaturesExceptSetupAndRegistrationInHeadlessAppAsync(
-        this UITestContext context)
-    {
-        await context.TestLoginWithInvalidDataAsync();
-        await context.TestLoginAsync();
-        await context.TestTurningFeatureOnAndOffAsync();
-        await context.TestLogoutAsync();
+        await context.TestBasicOrchardFeaturesExceptSetupAndRegistrationAsync(dontCheckFrontend);
     }
 
     /// <summary>
@@ -149,13 +109,18 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// <summary>
     /// <para>Tests all the basic Orchard features except for setup and registration.</para>
     /// <para>The test method assumes that the site is set up.</para>
+    /// <para>When running headless version of Orchard Core, ContentOperations shall be excluded.
+    /// Utilize <paramref name="dontCheckFrontend"></paramref>> for this purpose.</para>
     /// </summary>
+    /// <param name="dontCheckFrontend">Boolean to decide whether ContentOperations shall be executed.</param>>
     /// <returns>The same <see cref="UITestContext"/> instance.</returns>
-    public static async Task TestBasicOrchardFeaturesExceptSetupAndRegistrationAsync(this UITestContext context)
+    public static async Task TestBasicOrchardFeaturesExceptSetupAndRegistrationAsync(this UITestContext context, bool dontCheckFrontend)
     {
         await context.TestLoginWithInvalidDataAsync();
         await context.TestLoginAsync();
-        await context.TestContentOperationsAsync();
+
+        if (!dontCheckFrontend) await context.TestContentOperationsAsync();
+
         await context.TestTurningFeatureOnAndOffAsync();
         await context.TestLogoutAsync();
     }
--- a/Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
@@ -54,7 +54,7 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// </para>
     /// <para>The test method assumes that the site is not set up.</para>
     /// <para>
-    /// When running the headless version of Orchard Core, frontend operations shall be excluded. Utilize the <paramref
+    /// When running the headless version of Orchard Core, frontend operations shall be excluded. Utilize <paramref
     /// name="dontCheckFrontend"></paramref>> for this purpose.
     /// </para>
     /// </summary>
@@ -76,8 +76,10 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// <c>"SaaS"</c><see cref="OrchardCoreSetupParameters.RecipeId"/> value.
     /// </para>
     /// <para>The test method assumes that the site is not set up.</para>
-    /// <para>When running headless version of Orchard Core, frontend operations shall be excluded.
-    /// Utilize <paramref name="dontCheckFrontend"></paramref>> for this purpose.</para>
+    /// <para>
+    /// When running the headless version of Orchard Core, frontend operations shall be excluded. Utilize <paramref
+    /// name="dontCheckFrontend"></paramref>> for this purpose.
+    /// </para>
     /// </summary>
     /// <param name="setupParameters">The setup parameters.</param>
     /// <param name="dontCheckFrontend">Boolean to decide whether to check content on frontend.</param>>
@@ -113,7 +115,7 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// <para>The test method assumes that the site is set up.</para>
     /// <para>When running headless version of Orchard Core, ContentOperations shall be excluded.</para>
     /// <para>
-    /// When running the headless version of Orchard Core, frontend operations shall be excluded. Utilize the <paramref
+    /// When running the headless version of Orchard Core, frontend operations shall be excluded. Utilize <paramref
     /// name="dontCheckFrontend"></paramref>> for this purpose.
     /// </para>
     /// </summary>
@@ -423,7 +425,7 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// </list>
     /// <para>The test method assumes that there is currently a logged in admin user session.</para>
     /// <para>
-    /// When running the headless version of Orchard Core, frontend operations shall be excluded. Utilize the <paramref
+    /// When running the headless version of Orchard Core, frontend operations shall be excluded. Utilize <paramref
     /// name="dontCheckFrontend"></paramref>> for this purpose.
     /// </para>
     /// </summary>
--- a/Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
@@ -114,7 +114,9 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// </summary>
     /// <param name="dontCheckFrontend">Boolean to decide whether ContentOperations shall be executed.</param>>
     /// <returns>The same <see cref="UITestContext"/> instance.</returns>
-    public static async Task TestBasicOrchardFeaturesExceptSetupAndRegistrationAsync(this UITestContext context, bool dontCheckFrontend)
+    public static async Task TestBasicOrchardFeaturesExceptSetupAndRegistrationAsync(
+        this UITestContext context,
+        bool dontCheckFrontend)
     {
         await context.TestLoginWithInvalidDataAsync();
         await context.TestLoginAsync();
@@ -136,7 +138,9 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// </summary>
     /// <param name="setupParameters">The setup parameters.</param>
     /// <returns>The same <see cref="UITestContext"/> instance.</returns>
-    public static async Task TestSetupWithInvalidAndValidDataAsync(this UITestContext context, OrchardCoreSetupParameters setupParameters = null)
+    public static async Task TestSetupWithInvalidAndValidDataAsync(
+        this UITestContext context,
+        OrchardCoreSetupParameters setupParameters = null)
     {
         await context.TestSetupWithInvalidDataAsync();
         await context.TestSetupAsync(setupParameters);
--- a/Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
@@ -53,8 +53,10 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// specified <paramref name="setupRecipeId"/>.
     /// </para>
     /// <para>The test method assumes that the site is not set up.</para>
-    /// <para>When running headless version of Orchard Core, ContentOperations shall be excluded.
-    /// Utilize <paramref name="dontCheckFrontend"></paramref>> for this purpose.</para>
+    /// <para>
+    /// When running the headless version of Orchard Core, ContentOperations shall be excluded. Utilize <paramref
+    /// name="dontCheckFrontend"></paramref>> for this purpose.
+    /// </para>
     /// </summary>
     /// <param name="setupRecipeId">The ID of the recipe to be used to set up the site.</param>
     /// <param name="dontCheckFrontend">Boolean to decide whether to check content on frontend.</param>>
--- a/Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
@@ -412,6 +412,7 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// <item><description>Create the page with the given <paramref name="pageTitle"/>.</description></item>
     /// <item><description>Publish the page.</description></item>
     /// <item><description>Verify that the page is created.</description></item>
+    ///  <item><description>Check whether frontend operations shall be executed.</description></item>
     /// <item><description>Navigate to view the published page.</description></item>
     /// <item><description>Verify the page title and header.</description></item>
     /// </list>
--- a/Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
@@ -57,7 +57,7 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// Utilize <paramref name="dontCheckFrontend"></paramref>> for this purpose.</para>
     /// </summary>
     /// <param name="setupRecipeId">The ID of the recipe to be used to set up the site.</param>
-    /// <param name="dontCheckFrontend">Boolean to decide whether ContentOperations shall be executed.</param>>
+    /// <param name="dontCheckFrontend">Boolean to decide whether to check content on frontend.</param>>
     /// <returns>The same <see cref="UITestContext"/> instance.</returns>
     public static Task TestBasicOrchardFeaturesExceptRegistrationAsync(
         this UITestContext context,
@@ -78,7 +78,7 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// Utilize <paramref name="dontCheckFrontend"></paramref>> for this purpose.</para>
     /// </summary>
     /// <param name="setupParameters">The setup parameters.</param>
-    /// <param name="dontCheckFrontend">Boolean to decide whether ContentOperations shall be executed.</param>>
+    /// <param name="dontCheckFrontend">Boolean to decide whether to check content on frontend.</param>>
     /// <returns>The same <see cref="UITestContext"/> instance.</returns>
     public static async Task TestBasicOrchardFeaturesExceptRegistrationAsync(
         this UITestContext context,
@@ -112,7 +112,7 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// <para>When running headless version of Orchard Core, ContentOperations shall be excluded.
     /// Utilize <paramref name="dontCheckFrontend"></paramref>> for this purpose.</para>
     /// </summary>
-    /// <param name="dontCheckFrontend">Boolean to decide whether ContentOperations shall be executed.</param>>
+    /// <param name="dontCheckFrontend">Boolean to decide whether to check content on frontend.</param>>
     /// <returns>The same <see cref="UITestContext"/> instance.</returns>
     public static async Task TestBasicOrchardFeaturesExceptSetupAndRegistrationAsync(
         this UITestContext context,
@@ -120,9 +120,7 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     {
         await context.TestLoginWithInvalidDataAsync();
         await context.TestLoginAsync();
-
-        if (!dontCheckFrontend) await context.TestContentOperationsAsync();
-
+        await context.TestContentOperationsAsync(dontCheckFrontend);
         await context.TestTurningFeatureOnAndOffAsync();
         await context.TestLogoutAsync();
     }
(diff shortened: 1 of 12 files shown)


Luko6 (approved):

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -142,13 +142,19 @@ public static async Task<T> GoToPageAsync<T>(this UITestContext context, string
         context.GoToPageAsync<OrchardCoreRegistrationPage>();
 
     public static Task<OrchardCoreDashboardPage> GoToDashboardAsync(this UITestContext context) =>
-        context.GoToPageAsync<OrchardCoreDashboardPage>();
+        string.IsNullOrEmpty(context.CustomAdminUrl)
+            ? context.GoToPageAsync<OrchardCoreDashboardPage>()
+            : context.GoToPageAsync<OrchardCoreDashboardPage>(context.CustomAdminUrl);
```
Piedone: Instead of doing these if-elses, add a `GoToAdminPageAsync<T>`, then handle the prefixing there. The `[Url("Admin")]` attributes won't be needed anymore if you just supply the prefix at all times.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -271,5 +277,7 @@ public static void DismissAlert(this UITestContext context)
         context.GoToRelativeUrlAsync("/Contents/ContentItems/" + contentItemId);
 
     public static Task GoToContentItemEditorByIdAsync(this UITestContext context, string contentItemId) =>
-        context.GoToRelativeUrlAsync($"/Admin/Contents/ContentItems/{contentItemId}/Edit");
+        string.IsNullOrEmpty(context.CustomAdminUrl)
+            ? context.GoToRelativeUrlAsync($"/Admin/Contents/ContentItems/{contentItemId}/Edit")
+            : context.GoToRelativeUrlAsync($"/{context.CustomAdminUrl}/Contents/ContentItems/{contentItemId}/Edit");
```
Piedone: Instead of having this if-else here and similar ones in `OrchardCoreDashboardUITestContextExtensions`, add a `GoToAdminRelativeUrlAsync` and handle this prefixing there.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -271,5 +277,7 @@ public static void DismissAlert(this UITestContext context)
         context.GoToRelativeUrlAsync("/Contents/ContentItems/" + contentItemId);
 
     public static Task GoToContentItemEditorByIdAsync(this UITestContext context, string contentItemId) =>
```
Piedone: Move this to `OrchardCoreDashboardUITestContextExtensions` too.

Lombiq.Tests.UI/Services/UITestContext.cs
```diff
@@ -89,6 +89,13 @@ public class UITestContext
     /// </summary>
     public string TenantName { get; set; } = "Default";
 
+    /// <summary>
+    /// Gets or sets the current admin prefix if a custom one is needed. When creating UI tests on a site that
+    /// has custom admin prefix this value should be set in the UI test. After that navigation methods will be able
+    /// to catch up to the custom url.
+    /// </summary>
+    public string CustomAdminUrl { get; set; }
```
Piedone: This should be used everywhere where currently it's "/Admin". See:

- `AdminMonkeyTestingUrlFilter`
- `IsOrchardPage`
- `TestAdminAsMonkeyRecursivelyAsync`

Lombiq.Tests.UI/Services/UITestContext.cs
```diff
@@ -89,6 +89,13 @@ public class UITestContext
     /// </summary>
     public string TenantName { get; set; } = "Default";
 
+    /// <summary>
+    /// Gets or sets the current admin prefix if a custom one is needed. When creating UI tests on a site that
+    /// has custom admin prefix this value should be set in the UI test. After that navigation methods will be able
+    /// to catch up to the custom url.
```
Piedone: Break to 120 characters.

```suggestion
    /// Gets or sets the current Orchard Core admin prefix. When running UI tests on a site that
    /// uses a custom admin prefix, this value should be set in the test. After that, navigation methods will be able
    /// to use the custom URL.
```

Lombiq.Tests.UI/Services/UITestContext.cs
Piedone: By having the default specified here, you don't need a lot of if-elses otherwise.

```suggestion
    public string AdminUrlPrefix { get; set; } = "/Admin";
```

Lombiq.Tests.UI/Services/UITestContext.cs
```diff
@@ -206,4 +213,11 @@ private bool IsNoAlert()
             return true;
         }
     }
+
+    /// <summary>
+    /// Sets the custom admin url to the parameter value.
+    /// </summary>
+    /// <param name="customUrl">The desired custom admin url value.</param>
+    public void SetCustomAdminUrl(string customUrl) =>
```
Piedone: I don't think this is necessary, given that the property is writable.

Piedone (changes requested):
Note: If https://github.com/Lombiq/UI-Testing-Toolbox/pull/195 is merged before this, then `GoToContentTypesListAsync` and `GoToContentTypeEditorAsync` needs this supported too.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Psichorex: So basically instead of else-if-ing just created another method everywhere with RelativeUrl in the method name and pass the custom url?

Piedone: No, pass $"/Contents/ContentItems/{contentItemId}/Edit" to `GoToAdminRelativeUrlAsync` and only access `CustomAdminUrl` in the latter.

Psichorex: That is basically a new method instead of case dividing or not? 
I am asking this because if I create new extension methods like GoToContentItemPageRelativeAsync() because you suggested to leave the IF-ELSE syntax note that there are several other extension methods that are using these basic ones. So if I create a GoToAdminRelativeUrlAsync instead of GoToDashboardAsync with if-else that means that I have to address every single occurence of every method and case-divide there.

For example TestContentOperationsAsync() is using GoToContentItemsPageAsync() now it is perfect because if there is a CUstomUrl it will use the custom if not then /Admin no problem. But I remove if-else and there will be a GoToContentItemsRelative() then I have to case divide in these methods. Probably you understand it I just tried to make it absolutely clear what I am thinking and ask you what do you think about this.

Psichorex: Or I am the one that didn't catch your line of thoughts is also a possibility :D

Piedone: It's about adding a new method, yes. Instead of having to do this if-else everywhere, you do it once (and not even once, see my remark under `CustomAdminUrl`). But I'm talking about adding a new method while keeping `GoToDashboardAsync` and the others still, just not copy-paste this `string.IsNullOrEmpty(context.CustomAdminUrl) ? ... : ...` everywhere.

Psichorex: What if we don't create a single new method just give basic value to the AdminPrefix as you suggested and after that we just use the GoToPage<T>(string relativeUrl) overload everywhere?

Piedone: The issue with that is that you have to remember to reference and use `CustomAdminUrl` everywhere, instead of a dedicated method making it obvious. And if we want to change how such admin URLs are created and navigated to (what we're just doing) then we'd need to do it in every place instead of one. I.e., DRY.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Psichorex: So I can remove the [Url("Admin")] from OrchardCoreDashboardPage basically? Or shall I leave it and do
`GoToAdminPageAsync<T>(this UITestContext) => GoToPageAsync<OrchardCoreDashboardPage>(context.AdminPrefix)` ?
Removing the [Url] attribute will cause all UI tests to fail that uses this method I guess so it won't be a good choice.

Piedone: Remove, yes. All affected calls should go through `GoToDashboardAsync` anyway, so nothing should break.

Psichorex: And what do you suggest for extending `GoToFeaturesPageAsync()` and similar? They are using extended [Url("Admin/Features")] attributes. Easiest way would be to create `GoToRelativeFeaturesPage() => GoToAdminRelativeUrl("/Features")` but this way if a bigger method is using `GoToFeaturesPage()` we have to decide there if it has regular prefix or custom.

Piedone: In exactly the same fashion, you can get rid of that `Url` attribute, and all the others for dashboard pages, if they're used via their own methods like `GoToFeaturesPageAsync`. In them, you can supply the URL to `GoToAdminPageAsync<T>`.

An alternative that I just thought about is that you can keep the attributes, but change `BaseUrl` of the current `AtataContext` before navigation, then change back to the original one, but that's probably hackish.

The consumer won't have to decide, that's the point: What I'm explaining here, is that all of this prefix management should be in a single place, and all such page navigating methods should utilize that. The consuming code shouldn't change at all, hence why `BasicOrchardFeaturesTest` should also work without changes. E.g. `GoToFeaturesPageAsync` will continue to be `public static Task<OrchardCoreFeaturesPage> GoToFeaturesPageAsync(this UITestContext context)`, nothing changes in its signature, and no consumer of it needs to change anything.

Psichorex: So basically I can let the attributes go and change `GoToPageAsync()` calls to `GoToAdminRelativeUrl("/Features")` for example. This is clear then.

Piedone: Yes.

Lombiq.Tests.UI/Services/UITestContext.cs
Psichorex: At `AdminMonkeyTestingUrlFilter` I can't find a viable way to inject the AdminPrefix as it is not using UITestContext directly.
So next to the /admin there should be the context.AdminPrefix. Is UITestContext injectable btw?

Piedone:
Since https://github.com/Lombiq/UI-Testing-Toolbox/pull/195 is merged, GoToContentTypesListAsync and GoToContentTypeEditorAsync needs this supported too.

Lombiq.Tests.UI/Services/UITestContext.cs
Piedone: Don't you see `UITestContext` in `AdminMonkeyTestingUrlFilter`? ;) Please put more effort into figuring out these things on your own. You can also try injecting `UITestContext` and see if it works.

Psichorex: I am putting all my effort into my current issues. I am asking because this is an Open Source Library and I don't want to spaghetti it. There is `UITestContext ` in `AdminMonkeyTestingUrlFilter` but not in the class itself just the class's method takes it as a parameter.  Which is for me probably useless because I need to supply the new instance of `StartsWithMonkeyTestingUrlFilter` with the `context.AdminPrefix` as it is a readonly property I can't change it after creating it. If I create a CTOR `AdminMonkeyTestingUrlFilter` requiring UITestContext could be a possibility but that will result in chain so I will also have to address `NotAdminMonkeyTestingUrlFilter` with the same CTOR which seems overkill and spaghetti for me.

Psichorex: Well I guess I found a solution

Piedone: You see! `UITestContext` not being there in the ctor of `AdminMonkeyTestingUrlFilter` shouldn't be a problem, of course.

Psichorex: I just thought it is not as clean as it should be.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -17,6 +17,18 @@ public static class NavigationUITestContextExtensions
...
 
+    public static Task GoToAdminRelativeUrlAsync(
+        this UITestContext context,
+        string urlWithoutAdminPrefix = null,
+        bool onlyIfNotAlreadyThere = true)
+    {
+        if (string.IsNullOrEmpty(urlWithoutAdminPrefix)) return context.GoToDashboardAsync();
+
+        return context.GoToAbsoluteUrlAsync(
+            context.GetAbsoluteUri($"{context.AdminUrlPrefix + urlWithoutAdminPrefix}"),
```
Piedone: Use `UriBuilder`.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -115,6 +127,21 @@ public static async Task<T> GoToPageAsync<T>(this UITestContext context, string
         return page;
     }
 
+    public static async Task<T> GoToAdminPageAsync<T>(this UITestContext context, string relativeUrl = null)
+        where T : PageObject<T>
+    {
+        var page = context.ExecuteLogged(
+            $"{typeof(T).FullName} - {context.AdminUrlPrefix + relativeUrl}",
+            typeof(T).FullName,
+            () => context.Scope.AtataContext.Go.To<T>(url: context.GetAbsoluteUri(context.AdminUrlPrefix + relativeUrl).ToString()));
```
Piedone: The `GetAbsoluteUri(...)` logic should be DRY with `GoToAdminRelativeUrlAsync`.

Lombiq.Tests.UI/Extensions/OrchardCoreDashboardUITestContextExtensions.cs
```diff
@@ -80,22 +78,22 @@ public static async Task GoToContentItemListAndCreateNewAsync(this UITestContext
         this UITestContext context,
         string contentType,
         bool onlyIfNotAlreadyThere = true) =>
-        string.IsNullOrEmpty(context.CustomAdminUrl)
-            ? context.GoToRelativeUrlAsync($"/Admin/Contents/ContentTypes/{contentType}/Create", onlyIfNotAlreadyThere)
-            : context.GoToRelativeUrlAsync($"/{context.CustomAdminUrl}/Contents/ContentTypes/{contentType}/Create", onlyIfNotAlreadyThere);
+            context.GoToAdminRelativeUrlAsync(
+                $"/Contents/ContentTypes/{contentType}/Create",
+                onlyIfNotAlreadyThere);
```
Piedone: Can be one line.

Lombiq.Tests.UI/Helpers/UrlCheckHelper.cs
```diff
@@ -21,11 +21,12 @@ public static bool IsOrchardPage(UITestContext context)
         }
 
         return path.StartsWithOrdinalIgnoreCase("/admin") ||
-            path.StartsWithOrdinalIgnoreCase("/Register") ||
-            path.StartsWithOrdinalIgnoreCase("/Login") ||
-            path.StartsWithOrdinalIgnoreCase("/ChangePassword") ||
-            path.StartsWithOrdinalIgnoreCase("/ExternalLogins") ||
-            context.IsSetupPage();
+               path.StartsWithOrdinalIgnoreCase($"{context.AdminUrlPrefix}") ||
```
Piedone: String interpolation is not needed.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -277,7 +298,5 @@ public static void DismissAlert(this UITestContext context)
         context.GoToRelativeUrlAsync("/Contents/ContentItems/" + contentItemId);
 
     public static Task GoToContentItemEditorByIdAsync(this UITestContext context, string contentItemId) =>
```
Piedone: This is now duplicated.

Lombiq.Tests.UI/Services/UITestContext.cs
```diff
@@ -90,11 +90,11 @@ public class UITestContext
     public string TenantName { get; set; } = "Default";
 
     /// <summary>
-    /// Gets or sets the current admin prefix if a custom one is needed. When creating UI tests on a site that
-    /// has custom admin prefix this value should be set in the UI test. After that navigation methods will be able
-    /// to catch up to the custom url.
+    /// Gets or sets the current Orchard Core admin prefix. When running UI tests on a site that
+    /// uses a custom admin prefix, this value should be set in the test. After that, navigation methods will be able
+    /// to use the custom URL.
```
Piedone: Break to 120 characters.

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Psichorex: I don't get this. Are you suggesting to replace `GetAbsoluteUri()` with `GoToAdminRelativeUrlAsnyc()` ? 
Because that would save the context.AdminUrlPrefix parameter but the first is a `Uri` returning method and the second is a `Task` so I can't really replace one with the other.

Piedone: No. You have `context.GetAbsoluteUri(context.AdminUrlPrefix + relativeUrl)` copied in two places. This logic will get more involved with the `UriBuilder` usage I mentioned elsewhere. So, have it only in a single place and reuse it.

Psichorex: I did remove this line and replaced it with UriBuilder could you check?

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -24,9 +24,12 @@ public static class NavigationUITestContextExtensions
...
 
-        return context.GoToAbsoluteUrlAsync(
-            context.GetAbsoluteUri($"{context.AdminUrlPrefix + urlWithoutAdminPrefix}"),
-            onlyIfNotAlreadyThere);
+        var uriBuilder = new UriBuilder(context.Scope.BaseUri)
+        {
+            Path = context.AdminUrlPrefix + urlWithoutAdminPrefix,
+        };
+
+        return context.GoToAbsoluteUrlAsync(uriBuilder.Uri, onlyIfNotAlreadyThere);
```
Piedone: This isn't DRY, see my comment here: https://github.com/Lombiq/UI-Testing-Toolbox/pull/194#discussion_r953782995

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Piedone: Can you explain in your own words how you understand this suggestion? Why did I tell it? What may be that'd be better on this line when using `UriBuilder`?

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Psichorex: Addressed it.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Psichorex: My though was that  `$"{context.AdminUrlPrefix + urlWithoutAdminPrefix}"` isn't really a nice way of connecting 2 pieces of uri. I haven't really ever used UriBuilder before but I checked it out and I have a wild guess that apart from my usage of UriBuilder supplies the needed Uri it is still not the best as I still connecting the strings the same way. But I couldn't find a proper other way as both AdminPrefix + RelativeUrl should be in the Path property.

Piedone: Yes, exactly. The point is not to simply string concat `context.AdminUrlPrefix` and `urlWithoutAdminPrefix`. However, now I see that `UriBuilder` indeed doesn't have something like `Path.Combine()`, which I remembered it has, sorry about that. However, you can achieve the same concatenation with something like `new Uri(context.AdminUrlPrefix, urlWithoutAdminPrefix)`.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -307,4 +299,7 @@ public static void DismissAlert(this UITestContext context)
 
     public static Task GoToContentItemEditorByIdAsync(this UITestContext context, string contentItemId) =>
         context.GoToAdminRelativeUrlAsync($"/Contents/ContentItems/{contentItemId}/Edit");
+
+    private static UriBuilder ReturnRelativeUri(this UITestContext context, string relativeUrl) =>
+        new(context.Scope.BaseUri) { Path = context.AdminUrlPrefix + relativeUrl, };
```
Piedone: Keep using `GetAbsoluteUri`.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -307,4 +299,7 @@ public static void DismissAlert(this UITestContext context)
 
     public static Task GoToContentItemEditorByIdAsync(this UITestContext context, string contentItemId) =>
         context.GoToAdminRelativeUrlAsync($"/Contents/ContentItems/{contentItemId}/Edit");
+
+    private static UriBuilder ReturnRelativeUri(this UITestContext context, string relativeUrl) =>
```
Piedone: More like `GetAdminRelativeUrl`.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Psichorex: There isn't a ctor overload for `Uri()` that takes 2 string aswell. I added Flurl library that has Url.Combine and solved it with that. It's up to you now.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Psichorex: Now I think the solution looks clean.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Piedone: There's one that takes a `Uri` and a string or two `Uri`s.

Psichorex: Isn't this Url.Combine good? There is one with 2 Uri s that I know.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Piedone: What's the advantage of your custom code compared to `GetAbsoluteUri` that's exactly for creating the absolute URL from a relative one?

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Piedone: It's good but it would be better not to add and have to maintain a dependency for a one-liner that the framework provides functionality for.

image

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Psichorex: You are right I missed that Uri has the LocalPath which is equal to UriBuilder.Path.
I thought that I had to hack out the path if I return Uri.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Psichorex: I removed it as in the end it can be done with GetAbosuleUri

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Psichorex: With GetAbosuleUri I don't even need to connect Uri's because this actually does that and I can reach uri.LocalPath and the Uri itself that I need.

Psichorex: @Piedone Okay non of those constructors work for us. Literally they are only able to attach a base Uri to a relative One
so new `Uri("https:\\localhost:5002", "/Admin")` will result in `https:\\localhost:5002/Admin`
BUT `new Uri("localhost:5002/Admin", "/Features")` will result in `localhost:5002/Features` because `localhost:5002/Admin` is not a baseUri

Psichorex: So we either go with the flurl library Url.Combine or we just + operator them together

Psichorex: I went with a string.Concat and I created a `GetAsboluteAdminUrl()` method which takes a relativeUrl as a parameter will attach it inside to "/Admin".

Piedone: I see, OK then.

Lombiq.Tests.UI/Extensions/OrchardCoreDashboardUITestContextExtensions.cs
```diff
@@ -113,7 +111,4 @@ public static async Task GoToUsersAsync(this UITestContext context)
         await context.ClickReliablyOnAsync(By.CssSelector("#security .title"));
         await context.ClickReliablyOnAsync(By.CssSelector(".item-label.users .title"));
     }
-
-    public static Task GoToContentItemEditorByIdAsync(this UITestContext context, string contentItemId) =>
```
Piedone: This is good here, I previously asked about moving it to this class. Delete the other duplicate.

Lombiq.Tests.UI/Services/UITestContext.cs
```diff
@@ -90,9 +90,9 @@ public class UITestContext
     public string TenantName { get; set; } = "Default";
 
     /// <summary>
-    /// Gets or sets the current Orchard Core admin prefix. When running UI tests on a site that
-    /// uses a custom admin prefix, this value should be set in the test. After that, navigation methods will be able
-    /// to use the custom URL.
+    /// Gets or sets the current Orchard Core admin prefix. When running UI tests on a site that uses a custom admin
+    /// prefix, this value should be set in the test.
+    /// After that, navigation methods will be able to use the custom URL.
```
Piedone: You needn't a line break before that (it can be one paragraph, but if you'd want multiple paragraphs, `<para>` is for that).

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -80,6 +80,55 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
         await context.TestBasicOrchardFeaturesExceptSetupAndRegistrationAsync();
     }
 
+    /// <summary>
+    /// <para>
+    /// Tests all the basic Orchard features except for registration and content operations. At first sets up Orchard
```
Piedone: The comments don't tell about headless.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -80,6 +80,55 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
...
+        context.TestBasicOrchardFeaturesExceptRegistrationInHeadlessAppAsync(new OrchardCoreSetupParameters(context)
+        {
+            RecipeId = setupRecipeId,
+        });
+
+    /// <summary>
+    /// <para>
+    /// Tests all the basic Orchard features except for registration and content operations. At first sets up Orchard
+    /// with the recipe with the specified <paramref name="setupParameters"/>. By default uses new
+    /// <see cref="OrchardCoreSetupParameters"/> instance with <c>"SaaS"</c><see cref="OrchardCoreSetupParameters.RecipeId"/> value.
```
Piedone: Break at 120 characters.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -80,6 +80,55 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
         await context.TestBasicOrchardFeaturesExceptSetupAndRegistrationAsync();
     }
 
+    /// <summary>
```
Piedone: These new methods don't actually have anything to do with running Orchard in headless mode, specifically. Rather, they're just without content operations.

What I suggest is, add a parameter to `TestContentOperationsAsync` so it optionally doesn't check the frontend, and surface it in the methods using it. Then, utilize it in headless app.

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
Psichorex: Do you mean giving the main method TestBasicOrchardFestures... a headless boolean default false and pass it towards ContentOperations and if it is true then we neglect the useless operations. This way we dont create an additional method but extending the existing one with headless app functionality.

Piedone: I'd rather have the parameter something like `dontCheckFrontend` to make it more apparent what happens, but yes.

Psichorex: I made the needed changes and now the original method can be used to test the features without frontend check.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -53,14 +53,19 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// specified <paramref name="setupRecipeId"/>.
     /// </para>
     /// <para>The test method assumes that the site is not set up.</para>
+    /// <para>When running headless version of Orchard Core, ContentOperations shall be excluded.
+    /// Utilize <paramref name="dontCheckFrontend"></paramref>> for this purpose.</para>
```
Piedone: Add the "the" and format the other occasions similarly.

```suggestion
    /// <para>
    /// When running the headless version of Orchard Core, ContentOperations shall be excluded. Utilize <paramref
    /// name="dontCheckFrontend"></paramref>> for this purpose.
    /// </para>
```

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -100,13 +109,20 @@ public static async Task TestBasicOrchardFeaturesExceptSetupAsync(this UITestCon
...
-    public static async Task TestBasicOrchardFeaturesExceptSetupAndRegistrationAsync(this UITestContext context)
+    public static async Task TestBasicOrchardFeaturesExceptSetupAndRegistrationAsync(
+        this UITestContext context,
+        bool dontCheckFrontend)
     {
         await context.TestLoginWithInvalidDataAsync();
         await context.TestLoginAsync();
-        await context.TestContentOperationsAsync();
+
+        if (!dontCheckFrontend) await context.TestContentOperationsAsync();
```
Piedone: This is not what I sad (or what the parameter name implies) under https://github.com/Lombiq/UI-Testing-Toolbox/pull/194#discussion_r956957217 Rather, it shouldn't check the **frontend**. It should still check the content operations, just not open the page on the frontend.

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
Psichorex: Ah got you.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
Piedone: This isn't addressed.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -432,8 +435,11 @@ public static Task TestRegistrationAsync(this UITestContext context, UserRegistr
                     .CreateNewPage()
                         .Title.Set(pageTitle)
                         .Publish.ClickAndGo()
-                    .AlertMessages.Should.Contain(message => message.IsSuccess)
-                    .Items[item => item.Title == pageTitle].View.Click();
+                    .AlertMessages.Should.Contain(message => message.IsSuccess);
+
+                if (dontCheckFrontend) return;
```
Piedone, line 453: Call `TriggerAfterPageChangeEventAsync` before this too.

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
Psichorex: Do you mean change this to `Utilize **the** <paramref name="dontCheckFrontend">` ?

Psichorex: Got it.

PR #200, opened 2022-09-12
NEST-282: Adding extension method

porgabi:
[NEST-282](https://lombiq.atlassian.net/browse/NEST-282)


Lombiq.Tests.UI/Extensions/TenantsUITestContextExtensions.cs
```diff
@@ -0,0 +1,43 @@
+﻿using Lombiq.Tests.UI.Services;
+using OpenQA.Selenium;
+using System.Threading.Tasks;
+
+namespace Lombiq.Tests.UI.Extensions;
+
+public static class TenantsUITestContextExtensions
+{
+    public static async Task CreateNewTenantAsync(
```
wAsnk: There is a test controller [here](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI.Shortcuts/Controllers/TenantsController.cs) that already implements tenant creation.

And you can see how it is used [here.](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs#L237-L238)

Use that in your tests.

wAsnk (changes requested):

wAsnk (approved):

Lombiq.Tests.UI/Extensions/TenantsUITestContextExtensions.cs
porgabi: I'm running into some trouble when using this method in the OSOCE.NuGet UI tests. When it runs, it produces an exception:
image

This is especially puzzling as it runs just fine in the normal OSOCE solution's UI tests.

I found that when manually creating and setting up the tenant during a UI test, it works fine in the NuGet solution too. I'm still looking into this, but I'm also considering simply re-adding the method that this thread is attached to, perhaps with the name `CreateNewTenantManuallyAsync()`.

Is this exception familiar to you by any chance? 😃

wAsnk: I didn't meet with this error before. So it might be better to ask it in the general channel on Teams. Dávid might know it as he was the one who wrote this UI test function.

PR #207, opened 2022-10-18
OSOE-364: Support multi-tenancy

### LINKED ISSUE #80: Support multi-tenancy (OSOE-364)

Piedone:
OSOE-364
Fixes #80.


Lombiq.Tests.UI.Samples/Tests/ErrorHandlingTests.cs
```diff
@@ -129,6 +129,9 @@ public ErrorHandlingTests(ITestOutputHelper testOutputHelper)
                         throw new InvalidOperationException(
                             "This point shouldn't be reachable if the logs are properly kept.");
                     };
+
+                    // No need to add to create a failure dump folder for this test, since it'll always fail.
```
sarahelsaig: Did you mean this?
```suggestion
                    // No need to create a failure dump folder for this test, since it'll always fail.
```

Lombiq.Tests.UI.Samples/Tests/TenantTests.cs
```diff
@@ -32,14 +29,20 @@ public TenantTests(ITestOutputHelper testOutputHelper)
...
                 const string tenantAdminName = "tenantAdmin";
+
                 await context.SignInDirectlyAsync();
 
-                // Create the tenant with custom admin name.
-                await context.CreateAndEnterTenantAsync(
-                    TestTenantDisplayName,
+                // Create the tenant with a custom admin user.
+                await context.CreateAndSwitchToTenantAsync(
+                    TestTenantUrlPrefix,
```
sarahelsaig: Why did you replace the display name with URL prefix? If the `ShellSettings.Name` has to be a technical name, then it would be better to create a new value for it (perhaps the slugified version of the `TestTenantDisplayName`), because `TestTenantUrlPrefix` twice in a row looks confusing here.

Lombiq.Tests.UI.Tests.UI/Readme.md
```diff
@@ -0,0 +1,5 @@
+# Lombiq UI Testing Toolbox - UI Tests
+
+UI test project with tests that run tests on the UI Testing Toolbox itself. The project is intentionally not published on NuGet, since it's only needed for the UI Testing Toolbox.
```
sarahelsaig: Are the tests actually running tests? I don't see it.
```suggestion
UI test project that tests the UI Testing Toolbox itself. The project is intentionally not published on NuGet, since it's only needed for the UI Testing Toolbox.
```

Lombiq.Tests.UI.Tests.UI/Recipes/Lombiq.Tests.UI.Tests.UI.WorkflowShortcutsTests.recipe.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "Lombiq.Tests.UI.Tests.UI.WorkflowShortcutsTests",
+  "displayName": "Lombiq.Tests.UI.Tests.UI.WorkflowShortcutsTests",
```
sarahelsaig: None of the other recipes use a technical display name. How about "Lombiq UI Testing Toolbox - UI Tests - WorkflowShortcutsTests"?

Lombiq.Tests.UI.Tests.UI/TestCases/SecurityShortcutsTestCases.cs
```diff
@@ -0,0 +1,82 @@
...
+        await context.SignInDirectlyAsync(UserUserName);
+        (await context.GetCurrentUserNameAsync()).ShouldBe(UserUserName);
+    }
+
+    private static void CleanUpLogs(UITestContext context)
+    {
+        // Remove logs to have a clean slate.
+        foreach (var log in context.Application.GetLogs()) log.Remove();
+        context.ClearHistoricBrowserLog();
+    }
```
sarahelsaig: The same thing exists in `ErrorHandlingTests`. Would you mind turning it into a method of `UITestContext`?

Lombiq.Tests.UI.Tests.UI/TestCases/WorkflowShortcutsTestCases.cs
```diff
@@ -0,0 +1,25 @@
...
+        ExecuteTestAfterSetupAsync executeTestAfterSetupAsync, Browser browser) =>
+        executeTestAfterSetupAsync(
+            async context =>
+            {
+                await context.EnableFeatureDirectlyAsync(ShortcutsFeatureIds.Workflows);
+                await context.ExecuteRecipeDirectlyAsync("Lombiq.Tests.UI.Tests.UI.WorkflowShortcutsTests");
+                (await context.GenerateHttpEventUrlAsync(
+                    "testworkflow000000",
+                    "testhttpevent00000"))
+                    .ShouldStartWith("/workflows/Invoke?token=");
```
sarahelsaig: imo looks cleaner and easier to comprehend. I'm not a fan of the arguments and the subsequent method call being on the same indentation level.
```suggestion
                (await context.GenerateHttpEventUrlAsync("testworkflow000000", "testhttpevent00000"))
                    .ShouldStartWith("/workflows/Invoke?token=");
```

Lombiq.Tests.UI/Extensions/IWebApplicationInstanceExtensions.cs
```diff
@@ -22,27 +35,71 @@ public static class IWebApplicationInstanceExtensions
...
         finally
         {
-            httpContextAccessor.HttpContext = null;
+            httpContextAccessor.HttpContext = originalHttpContext;
+        }
+    }
+}
+
+// This is from the 1.4 version of Orchard Core. During an Orchard upgrade to 1.5, remove this, and use the now public
+// implementation in OrchardCore.Abstractions.
```
sarahelsaig, line 68: Should this be mentioned in https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/issues/221 ?

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -50,11 +50,11 @@ await context.Configuration.Events.AfterNavigation
     public static string GetCurrentAbsolutePath(this UITestContext context) => context.GetCurrentUri().AbsolutePath;
 
     public static Uri GetAbsoluteUri(this UITestContext context, string relativeUrl) =>
-        new(context.Scope.BaseUri, relativeUrl);
+        new(context.Scope.BaseUri, relativeUrl.StartsWithOrdinal("/") ? relativeUrl[1..] : relativeUrl);
```
sarahelsaig: So like `relativeUrl.TrimStart('/')`?

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
```diff
@@ -142,34 +146,34 @@ public static async Task<string> GetCurrentUserNameAsync(this UITestContext cont
...
+                            EmailConfirmed = true,
+                            IsEnabled = true,
+                        },
+                        password,
+                        (key, error) => errors.Add(key, error));
+
+                    if (user == null)
+                    {
+                        var exceptionLines = new StringBuilder();
+                        exceptionLines.AppendLine("Create user error:");
```
sarahelsaig: Shouldn't this be either `"User creation error:"` or `$"{nameof(CreateUserAsync)} error:"`?

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
```diff
@@ -142,34 +146,34 @@ public static async Task<string> GetCurrentUserNameAsync(this UITestContext cont
...
+                        errors.ForEach(entry =>
+                            exceptionLines.AppendLine(CultureInfo.InvariantCulture, $"{entry.Key}: {entry.Value}"));
```
sarahelsaig: Since this is always a multiline exception it would be good it indent or bullet the subsequent lines, e.g.
```suggestion
                            exceptionLines.AppendLine(CultureInfo.InvariantCulture, $"- {entry.Key}: {entry.Value}"));
```

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
```diff
@@ -359,21 +365,21 @@ private sealed class ExecuteRecipeShortcut { }
 
                     if (recipe == null)
                     {
-                        throw new RecipeNotFoundException($"{recipeName} not found!");
+                        throw new RecipeNotFoundException($"Recipe with the name {recipeName} not found.");
```
sarahelsaig: The other similar exceptions (e.g. [here](https://github.com/Lombiq/UI-Testing-Toolbox/blob/f084fd99e82ee0d7dc91975c3859d43649a6c14d/Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs#L201)) had the name quoted, so it should be here as well.

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
```diff
@@ -431,74 +437,113 @@ public interface IShortcutsApi
...
         string name,
         string urlPrefix,
-        string recipe,
-        CreateTenant model = null)
+        OrchardCoreSetupParameters setupParameters)
     {
-        model ??= new CreateTenant();
+        setupParameters ??= new OrchardCoreSetupParameters(context);
+        var databaseProvider = setupParameters.DatabaseProvider == OrchardCoreSetupPage.DatabaseType.SqlServer
+            ? "SqlConnection"
```
sarahelsaig, line 489: Might want to mark this for OC 1.5 changes too, because there is a constant for it in the main branch: `OrchardCore.Data.DatabaseProviderValue.SqlConnection`

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
```diff
@@ -431,74 +437,113 @@ public interface IShortcutsApi
...
+                    {
+                        await shellFeatureManager.EnableFeaturesAsync(new[] { themeFeature }, force: true);
+                    }
+                },
+                tenant,
+                activateShell);
 
     /// <summary>
-    /// Creates, sets up and navigates to a new URL prefixed tenant. Also changes <see cref="UITestContext.TenantName"/>.
+    /// Creates, sets up, switches to, and navigates to a new URL-prefixed tenant.
```
sarahelsaig: It still changes `TenantName` (and the others) via `context.SwitchCurrentTenant()`. I think this should be mentioned at least in a `<remarks>` to avoid confusion as the context's internal state has been changed.

Lombiq.Tests.UI/OrchardCoreUITestBase.cs
```diff
@@ -17,6 +17,19 @@ internal static class OrchardCoreUITestBaseCounter
     public static bool AppFolderCreated { get; set; }
 }
 
+/// <summary>
+/// Delegate with the signature of <see
+/// cref="OrchardCoreUITestBase{TEntryPoint}.ExecuteTestAfterSetupAsync(Func{UITestContext, Task}, Browser,
+/// Func{OrchardCoreUITestExecutorConfiguration, Task})"/> so test case classes like <c>CustomAdminPrefixTestCase</c> in
+/// <c>Lombiq.Tests.UI.Tests.UI</c> can easily depend on the method without having to define custom delegeate
```
sarahelsaig:
```suggestion
/// <c>Lombiq.Tests.UI.Tests.UI</c> can easily depend on the method without having to define custom delegate
```

Lombiq.Tests.UI/Services/AtataScope.cs
```diff
@@ -10,14 +10,26 @@ namespace Lombiq.Tests.UI.Services;
...
+
     public AtataContext AtataContext { get; }
     public IWebDriver Driver => AtataContext.Driver;
-    public Uri BaseUri { get; }
+
+    public Uri BaseUri
+    {
+        get => _baseUri;
+
+        set
```
sarahelsaig: Is the empty line necessary?

Lombiq.Tests.UI/Services/AtataScope.cs
```diff
@@ -10,14 +10,26 @@ namespace Lombiq.Tests.UI.Services;
...
+        {
+            _baseUri = value;
+            AtataContext.BaseUrl = value.ToString();
```
sarahelsaig, line 25: This will throw NRE if `value is null`. If this is intentional, then add `ArgumentNullException.ThrowIfNull(value);` at the beginning of the setter.

sarahelsaig (changes requested):

Lombiq.Tests.UI.Samples/Tests/ErrorHandlingTests.cs
Piedone: Yep!

Lombiq.Tests.UI.Tests.UI/Readme.md
Piedone: There are no test classes per se indeed.

Lombiq.Tests.UI.Tests.UI/TestCases/WorkflowShortcutsTestCases.cs
Piedone: This was like that already, but yeah, `ShouldStartWith()` should really start one tab to the left. But the preceding line doesn't need to be broken up indeed.

Lombiq.Tests.UI.Samples/Tests/TenantTests.cs
Piedone: The display name isn't replaced, it's set on line 41, and the method requires a technical name and a prefix (as this is what's needed for creating the tenant itself, the rest is setup). The duplication here is only an artifact of simplification in the test (since "test" as the technical name and URL prefix is both suitable), but I'll introduce a second const for it to be nicer.

Lombiq.Tests.UI.Tests.UI/Recipes/Lombiq.Tests.UI.Tests.UI.WorkflowShortcutsTests.recipe.json
Piedone: I was being lazy here, since nobody will actually see this recipe name (as opposed to the test recipes in `Lombiq.OSOCE.Web`. But yeah, better.

Lombiq.Tests.UI/Extensions/IWebApplicationInstanceExtensions.cs
Piedone: Not really, since our OC upgrade guide already tells you to search for, among others "Orchard upgrade". The points under that issue are rather things that aren't this obvious.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Piedone: Yeah :D.

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
Piedone: While I haven't written this, I suppose the intent was to follow the name of the method in a natural language way. But "User creation error:" sounds better.

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
Piedone: Why I didn't do this is because this is a dot-separated technical name, so (as opposed to display names that can contain spaces) it's apparent what's the name (which is actually more like an ID). But for consistency, we can have it with quotes.

Lombiq.Tests.UI/Extensions/IWebApplicationInstanceExtensions.cs
sarahelsaig: I'm not too happy about that, because that's just a todo comment by another name. Seems like a somewhat fragile format too, because if the "Orchard upgrade" text is caught by a line break (maybe automatically) then search won't be able to fine it later.
It's not a problem here, so I'm not going start a revolution over it... for now!

Piedone: I see the issue. The alternative would require everyone creating or finding the corresponding upgrade issue, also across projects, which would bring its own issues, however.

sarahelsaig: That wouldn't be a problem if we always maintain a single vNext issue (so create a new one once the latest is resolved) in the OSOCE repo, and link it in the https://lombiq.atlassian.net/wiki/spaces/OSOE/overview?homepageId=49283109 so anyone can easily find it. Or just pin it in GitHub.

Piedone: Yeah, but then you have to also keep it updated :). I mean, it's not less work or less fragile, and you'd still need to link the issue from the code for clarity. The line break issue should be possible to work around with a regex search.

sarahelsaig: I don't get how it's no less fragile tbh, a shared list of all todos sounds pretty robust.

Piedone: You have to know that the issue exists and you have to update it. It moves information about the code from the code to an issue. Since there can be a lot of time between us noticing an issue that'll be fixed by an OC upgrade (or that needs special care during it) and when the upgrade happens, it's safer to keep it in the code, as it relies less on people knowing about things that are less noticable.

0liver: Judging by previous comments of yours, Zoltán, on other TODO-ish comments, I feel like this one should also be moved into an issue, much like Dávid suggested in his first comment. I don't really understand the difference between this and other situations. If we know of this todo today, let's put it on an explicit list in the update issue and be done with it, I would say.

Piedone: Really, as I explained above. These are not just random TO DO comments but part of an organized process of documenting areas of the code that need special care during an Orchard upgrade (this can BTW be continuous too, like the NLog version in the UI Testing Toolbox, not just a one-off task). These are not "do this when you have time" kind of tasks but rather "if you change this code in a particular way in the future, then do that".

PR #208, opened 2022-10-19
OSOE-158: Typo fixes

Luko6:
OSOE-158


Lombiq.Tests.UI/Services/AccessibilityCheckingConfiguration.cs
```diff
@@ -45,10 +45,10 @@ public class AccessibilityCheckingConfiguration
     /// Gets or sets a predicate that determines whether accessibility checking and asserting the results should run for
     /// the current page. This is only used if <see cref="RunAccessibilityCheckingAssertionOnAllPageChanges"/> is set to
     /// <see langword="true"/>. Defaults to <see
-    /// cref="EnableOnValidatablePagesAccessbilityCheckingAndAssertionOnPageChangeRule"/>.
+    /// cref="EnableOnValidatablePagesAccessibilityCheckingAndAssertionOnPageChangeRule"/>.
     /// </summary>
-    public Predicate<UITestContext> AccessbilityCheckingAndAssertionOnPageChangeRule { get; set; } =
-        EnableOnValidatablePagesAccessbilityCheckingAndAssertionOnPageChangeRule;
+    public Predicate<UITestContext> AccessibilityCheckingAndAssertionOnPageChangeRule { get; set; } =
```
Piedone, line 58: This is a breaking change, so re-add the property with the old name with `[Obsolete]`, and let it just proxy to this one.

Piedone (changes requested):

Lombiq.Tests.UI/Services/AccessibilityCheckingConfiguration.cs
Luko6: I couldn't find a way without making the correctly named method static, which is unsafe so, I added this back the way it was with obsolete.

Piedone: But that won't work, since nothing uses it.

Luko6: What do you recommend then?

0liver: Intrigued by your comments, I checked and this works for me:
```c#
    public Predicate<UITestContext> AccessbilityCheckingAndAssertionOnPageChangeRule
    {
        get => AccessibilityCheckingAndAssertionOnPageChangeRule;
        set => AccessibilityCheckingAndAssertionOnPageChangeRule = value;
    }

    public Predicate<UITestContext> AccessibilityCheckingAndAssertionOnPageChangeRule { get; set; } =
        EnableOnValidatablePagesAccessbilityCheckingAndAssertionOnPageChangeRule;
```

Piedone: This, i.e. a proxy property, is what I've recommended.

Piedone (approved):

Lombiq.Tests.UI/Services/AccessibilityCheckingConfiguration.cs
```diff
@@ -41,6 +41,7 @@ public class AccessibilityCheckingConfiguration
     /// </summary>
     public bool RunAccessibilityCheckingAssertionOnAllPageChanges { get; set; }
 
+    // Once this method is removed, remove 'Accessbility' from the spelling whitelist.
```
Piedone:
```suggestion
    // Once this method is removed, remove 'Accessbility' from the spelling whitelist in GitHub Actions.
```

Piedone (changes requested):

Piedone (approved):

PR #211, opened 2022-10-19
SNOW-142: Adjusting MonkeyTestingOptions

Psichorex:
SNOW-142


Lombiq.Tests.UI/MonkeyTesting/MonkeyTestingOptions.cs
```diff
@@ -53,6 +53,9 @@ public sealed class MonkeyTestingOptions
             log: true,
             canClick: (element) => {
                     for (; element && element !== document; element = element.parentNode) {
+                        if (element.getAttribute('action') === '/Users/LogOff') {
+                            return false;
+                        }
```
Piedone: This shouldn't be there in general. By default, we want monkey testing to also be able to log out the user.

Psichorex: By the behavior of the test if it logs out, how would it log back again? The problem was it logged out and tried to navigate back to a link which was not available for anonymous users

Piedone: Logging back in isn't necessarily desired. The point of monkey testing is to do random interactions. If that leads to logout, then that's fine, and the application should keep working as designed even after that.

That behavior should result in the login screen being shown, and that should be accepted by the test.

Psichorex: That is understandable. My question is then how is that possible that after logging out it tried to navigate to /RequestDemo/Success or something like that needs permission. Is it intended functionality?

Piedone: I suppose even after logout, there was a UI element that when interacted with, brought the browser to /RequestDemo/Success.

Psichorex: It could be for example tge go back to last page but after logout that could cause problems on several sites not only in snow

Psichorex: Even the UITest's name is AuthenticatedFrontendShouldNotBreak so the test really should not log out. Possibly we could have 2 kinds of monkey tests if logging out has a purpose by the way.

Piedone: Monkey tests don't navigate back unless there's a UI element in the app for it. And if there is, then the app should still keep on working, as that's something that any user can do. (Even if you e.g. log out in one tab and click on something in another tab that only an authenticated user would see, the app should still not break.) So again, I don't see why preventing monkey tests from logging out would make sense.

Psichorex: The reason of fail is RequestDemo/Success will be a 404 unless you are submitting a legit request for a demo.
If you go back to /Success there will be problems in the controller and it will return NotFound() which is legit and the app should work and it is working just because you got a 404 for a route. But a 404 will fail a UITest and that is a problem.

Piedone: You can configure the test with a custom browser log assertion that excludes that page from the 404 fail. You can BTW also change `GremlinsSpecies` from the test if you want to prevent logout in your app, but it's not suitable as a blanket configuration for everyone.

Psichorex: @Piedone Should n't this code do the same as in the Javascript? Cause I believe it should thus we can save the JS magic.
germ

Piedone: Please check it out in the code and you should get a definitive answer.

barthamark (approved):

PR #216, opened 2022-10-24
OSOE-353: Add duplicate SQL query detector in Lombiq.UITestingToolbox

### LINKED ISSUE #93: Add duplicate SQL query detector (OSOE-353)
When one implements a `SELECT` N+1 query how it manifests is that the same query (apart from the parameters) will be executed multiple times during page load. While not every duplicated query is a problem, we could add a feature to detect such queries and fail the given test.

### Implementation notes

- An approach would be to use the Orchard Core Mini Profiler feature for this. A better one would be to implement our similar profiling logic with an `ISession` decorator to tap into the OC app since that wouldn't potentially break the UI (as the Mini Profile toolbar can). Perhaps it's possible to get the info out of Mini Profiler without showing the UI.
- The initial implementation under https://github.com/Lombiq/UI-Testing-Toolbox/pull/216 may be used for inspiration, but not the code directly.
- Common APIs for https://github.com/OrchardCMS/OrchardCore/issues/12251 can help here too.

### Checklist

- [x] Add per-test configurable count thresholds (that are kept for a given DI scope, so roughly an HTTP request) for the following:
  - A SQL command is executed N times, regardless of its parameters. This can surface cases where instead of the same query being executed multiple times with different parameters you should rather have a single query that matches all desired rows.
  - A SQL command is executed N times, but duplicates are only counted if the parameters are the same too. This can pinpoint cases where the result should be just cached.
  - A SQL command's result set has more than N items. This can surface queries that lack appropriate filtering (and perhaps only filter in .NET instead of SQL).
- [x] While the counting needs to happen in the Orchard Core app, the configurations and assertions (also when and how they're executed) should happen in the test project, following the patterns of accessibility checking and HTML validation.
- [ ] Add a heavily documented sample test that demonstrates all the basic capabilities of this feature to `Lombiq.Tests.UI.Samples`.
- [x] Add a documentation page about the feature and mention it as a highlight in the root Readme.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-353)

dministro:
[OSOE-353](https://lombiq.atlassian.net/browse/OSOE-353)
Fixes #93 

[OSOE-353]: https://lombiq.atlassian.net/browse/OSOE-353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - Introduced a "Duplicated SQL query detector" to improve test efficiency and resource usage.
  - Added configuration settings for UI Testing Toolbox to enhance SQL query monitoring.
  - Implemented a new set of tests for detecting duplicated SQL queries.

- **Enhancements**
  - Improved test environment with additional configuration options for database command execution thresholds.
  - Enhanced navigation and page load testing with new counter data collection and assertion capabilities.
  - Expanded testing toolbox with classes and interfaces to support performance metric collection.

- **Documentation**
  - Updated the Readme to reflect new testing features and configurations.

- **Bug Fixes**
  - Fixed issues related to database command execution tracking during tests.

- **Refactor**
  - Refactored database connection and command wrappers to support performance data collection.

- **Tests**
  - Added tests for new functionality and performance monitoring features.

- **Chores**
  - Updated internal test infrastructure to incorporate new counter data collector services.


Lombiq.Tests.UI/Exceptions/CounterThresholdException.cs
```diff
@@ -0,0 +1,50 @@
+using Lombiq.Tests.UI.Services.Counters;
+using System;
+using System.Text;
+
+namespace Lombiq.Tests.UI.Exceptions;
+
+// We need constructors with required informations.
+#pragma warning disable CA1032 // Implement standard exception constructors
```
0liver: The warning does not disallow custom constructors, but it enforces implementation of all default constructors incl. one to support serialization.

I don't think we should opt out of that.

sarahelsaig: I concur, you must have [these constructors](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1032#rule-description) for safe error handling, but you can have additional constructors if you want to include custom logic.

Lombiq.Tests.UI/Services/PhaseCounterConfiguration.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Lombiq.Tests.UI.Services;
+
+public class PhaseCounterConfiguration
+{
+    public Action<ICounterProbe> AssertCounterData { get; set; }
+    public int DbCommandExecutionRepetitionPerNavigationThreshold { get; set; } = 11;
+    public int DbCommandExecutionRepetitionThreshold { get; set; } = 22;
+    public int DbReaderReadPerNavigationThreshold { get; set; } = 11;
+    public int DbReaderReadThreshold { get; set; } = 11;
```
Piedone: Add docs to the properties.

Lombiq.Tests.UI/Services/PhaseCounterConfiguration.cs
```diff
@@ -0,0 +1,13 @@
+using Lombiq.Tests.UI.Services.Counters;
+using System;
+
+namespace Lombiq.Tests.UI.Services;
+
+public class PhaseCounterConfiguration
```
Piedone: Mention this feature in the root Readme.

Lombiq.Tests.UI/Services/UITestContext.cs
```diff
@@ -100,13 +100,19 @@ public class UITestContext
     /// </summary>
     public string AdminUrlPrefix { get; set; } = "/Admin";
 
+    /// <summary>
+    /// Gets the currently running <see cref="CounterDataCollector"/> instance.
+    /// </summary>
+    public CounterDataCollector CounterDataCollector { get; init; }
```
Piedone: This needs to be possible to be completely disabled too.

Lombiq.Tests.UI/Services/CounterDataCollector.cs
```diff
@@ -0,0 +1,34 @@
...
+        _probes.Clear();
+        Clear();
+    }
+
+    public override void Increment(ICounterKey counter)
+    {
+        _probes.SelectWhere(probe => probe, probe => probe.IsRunning)
+            .ForEach(probe => probe.Increment(counter));
+
+        base.Increment(counter);
```
Piedone:
```suggestion
        _probes.Where(probe => probe.IsRunning).ForEach(probe => probe.Increment(counter));
        base. Increment(counter);
```

Lombiq.Tests.UI/Services/Counters/CounterProbeBase.cs
```diff
@@ -0,0 +1,39 @@
...
+    public abstract string DumpHeadline();
+
+    public abstract string Dump();
+
+    private static TCounter TryConvertAndUpdate<TCounter>(ICounterValue counter, Action<TCounter> update)
+    {
+        if (counter is not TCounter value)
+        {
+            throw new ArgumentException(
+                $"The type of ${nameof(counter)} is not compatible with ${typeof(TCounter).Name}");
```
Piedone:
```suggestion
                $"The type of ${nameof(counter)} is not compatible with ${typeof(TCounter).Name}.");
```

Lombiq.Tests.UI/Services/Counters/ICounterProbe.cs
```diff
@@ -0,0 +1,38 @@
+using System.Collections.Generic;
+
+namespace Lombiq.Tests.UI.Services.Counters;
+
+/// <summary>
+/// Represents a probe for the counter infrastructure.
```
Piedone: Explain what a probe is: "Represents a probe (...)..."

Lombiq.Tests.UI/Services/Counters/ICounterValue.cs
```diff
@@ -0,0 +1,13 @@
...
+
+/// <summary>
+/// Represents a value in <see cref="ICounterProbe.Counters"/>.
+/// </summary>
+public interface ICounterValue
+{
+    /// <summary>
+    /// Dumps the value content to a human readable format.
+    /// </summary>
+    /// <returns>A human readable string representation of instance.</returns>
```
Piedone:
```suggestion
    /// <returns>A human-readable string representation of the instance.</returns>
```

Lombiq.Tests.UI/Services/Counters/ICounterProbe.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    /// <summary>
+    /// Increments the <see cref="Value.IntegerCounterValue"/> selected by <paramref name="counter"/>. If the
+    /// <see cref="Value.IntegerCounterValue"/> does not exists, creates a new instance.
+    /// </summary>
+    /// <param name="counter">The counter key.</param>
+    void Increment(ICounterKey counter);
+
+    /// <summary>
+    /// Dumps the probe headline to a human readable format.
```
Piedone: "human-readable" everywhere.

Lombiq.Tests.UI/Services/Counters/NavigationProbe.cs
```diff
@@ -0,0 +1,14 @@
...
+
+public class NavigationProbe : CounterProbe
+{
+    public Uri AbsoluteUri { get; init; }
+
+    public NavigationProbe(ICounterDataCollector counterDataCollector, Uri absoluteUri)
+        : base(counterDataCollector) =>
+        AbsoluteUri = absoluteUri;
+
+    public override string DumpHeadline() => $"{nameof(NavigationProbe)}, AbsoluteUri={AbsoluteUri}";
```
Piedone:
```suggestion
    public override string DumpHeadline() => $"{nameof(NavigationProbe)}, AbsoluteUri = {AbsoluteUri}";
```

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -39,7 +40,10 @@ public static class NavigationUITestContextExtensions
                 await context.Configuration.Events.BeforeNavigation
                     .InvokeAsync<NavigationEventHandler>(eventHandler => eventHandler(context, absoluteUri));
 
-                context.Driver.Navigate().GoToUrl(absoluteUri);
+                using (new NavigationProbe(context.CounterDataCollector, absoluteUri))
+                {
+                    context.Driver.Navigate().GoToUrl(absoluteUri);
+                }
```
Piedone: Note that this is not the only way we navigate between pages. Check `UITestExecutionEvents` for safer ways to make sure you run something before and after navigation.

Lombiq.Tests.UI.Samples/Tests/BasicOrchardFeaturesTests.cs
```diff
@@ -25,7 +25,17 @@ public BasicOrchardFeaturesTests(ITestOutputHelper testOutputHelper)
...
         ExecuteTestAsync(
             context => context.TestBasicOrchardFeaturesAsync(RecipeIds.BasicOrchardFeaturesTests),
-            browser);
+            browser,
+            configuration =>
+            {
+                // The UI Testing Toolbox includes DbCommand execution counter, after the end of test it checks the
+                // number of executed commands with the same Sql command and parameter set against the threshold value
+                // in its configuration. If the executed command count is greater then the threshold, it raises a
+                // CounterThresholdException. So here we set the minimum required value to avoid it.
```
Piedone:
```suggestion
                // The UI Testing Toolbox includes a DbCommand execution counter. After the end of the test, it checks the
                // number of executed commands with the same SQL command and parameter set against the threshold value
                // in its configuration. If the executed command count is greater than the threshold, it raises a
                // CounterThresholdException. So here we set the minimum required value to avoid it.
```

Lombiq.Tests.UI.Samples/Tests/BasicOrchardFeaturesTests.cs
```diff
@@ -25,7 +25,17 @@ public BasicOrchardFeaturesTests(ITestOutputHelper testOutputHelper)
...
+                configuration.CounterConfiguration.Running.DbCommandExecutionRepetitionThreshold = 26;
```
Piedone: Also demonstrate how this can be set differently for just a given page. We'll tune these values for each feature/page we test.

Lombiq.Tests.UI.Samples/Tests/BasicOrchardFeaturesTests.cs
```diff
@@ -25,7 +25,17 @@ public BasicOrchardFeaturesTests(ITestOutputHelper testOutputHelper)
     public Task BasicOrchardFeaturesShouldWork(Browser browser) =>
         ExecuteTestAsync(
             context => context.TestBasicOrchardFeaturesAsync(RecipeIds.BasicOrchardFeaturesTests),
-            browser);
+            browser,
+            configuration =>
+            {
+                // The UI Testing Toolbox includes DbCommand execution counter, after the end of test it checks the
+                // number of executed commands with the same Sql command and parameter set against the threshold value
```
Piedone: It's only really the SQL command, regardless of the parameters, that we need to count. A fully repeated query is an issue, but it's a subset of duplicated queries, especially if we want to guard against `SELECT` N+1 queries (where the same query is executed repeatedly, but with different parameters).

Lombiq.Tests.UI.Samples/Tests/BasicOrchardFeaturesTests.cs
```diff
@@ -25,7 +25,17 @@ public BasicOrchardFeaturesTests(ITestOutputHelper testOutputHelper)
     public Task BasicOrchardFeaturesShouldWork(Browser browser) =>
         ExecuteTestAsync(
             context => context.TestBasicOrchardFeaturesAsync(RecipeIds.BasicOrchardFeaturesTests),
-            browser);
+            browser,
+            configuration =>
+            {
+                // The UI Testing Toolbox includes DbCommand execution counter, after the end of test it checks the
```
Piedone: Demonstrating this feature looks more suitable for its own sample test.

Lombiq.Tests.UI/Exceptions/CounterThresholdException.cs
```diff
@@ -0,0 +1,64 @@
...
+        ICounterProbe probe,
+        ICounterKey counter,
+        ICounterValue value,
+        string message,
+        Exception innerException)
+        : base(FormatMessage(probe, counter, value, message), innerException)
+    {
+    }
+
+    private static string FormatMessage(
```
Piedone, line 51: Generate exception messages that are easier to understand than:

> DbExecuteCounterKey
> 	SELECT [Document].* FROM [Document] WHERE [Document].[Type] = @Type LIMIT 1
> 		[0]Type = OrchardCore.Settings.SiteSettings, OrchardCore.Settings
> 
> IntegerCounterValue value: 26
> Counter value is greater then DbCommandExecutionRepetitionThreshold, threshold: 22

I as an ordinary developer who didn't set up these counter thresholds but just wanted to change something unrelated but accidentally implemented a `SELECT` N+1 issue, should be able to understand what I did wrong.

Lombiq.Tests.UI/Services/CounterConfiguration.cs
```diff
@@ -0,0 +1,62 @@
...
+            .ForEach(key =>
+            {
+                if (probe.Counters[key] is IntegerCounterValue counterValue
+                    && counterValue.Value > threshold)
+                {
+                    throw new CounterThresholdException(
+                        probe,
+                        key,
+                        counterValue,
+                        $"Counter value is greater then {thresholdName}, threshold: {threshold.ToTechnicalString()}");
```
Piedone:
```suggestion
                        $"Counter value is greater then {thresholdName}, threshold: {threshold.ToTechnicalString()}.");
```

Lombiq.Tests.UI/Services/OrchardCoreHosting/OrchardApplicationFactory.cs
```diff
@@ -25,16 +27,19 @@ public sealed class OrchardApplicationFactory<TStartup> : WebApplicationFactory<
     private readonly Action<IConfigurationBuilder> _configureHost;
     private readonly Action<IWebHostBuilder> _configuration;
     private readonly Action<ConfigurationManager, OrchardCoreBuilder> _configureOrchard;
-    private readonly List<IStore> _createdStores = new();
+    private readonly ConcurrentBag<IStore> _createdStores = new();
+    private readonly ICounterDataCollector _counterDataCollector;
```
Piedone: Keep the order of private fields and ctor assignments the same as the ctor parameters.

Lombiq.Tests.UI/Services/Counters/Data/ProbedDbConnection.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.ComponentModel;
+using System.Data;
+using System.Data.Common;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Transactions;
+
+namespace Lombiq.Tests.UI.Services.Counters.Data;
+
+[DesignerCategory("")]
```
Piedone, line 12: Comment why this is necessary.

Lombiq.Tests.UI/Services/Counters/CounterProbe.cs
```diff
@@ -0,0 +1,63 @@
+using System;
+using System.Text;
+
+namespace Lombiq.Tests.UI.Services.Counters;
+
+public abstract class CounterProbe : CounterProbeBase, IDisposable
```
Piedone, line 7: After a page change, write all counters' values to the TestOutput for debugging.

Piedone (changes requested):
Impressive!

Lombiq.Tests.UI/Services/CounterDataCollector.cs
dministro: The `ConcurrentBag` doesn't have `Where` method.

Piedone: There's an extension for it in `System.Linq` as usual. Do check though if such iterating needs any special care due to it being a concurrent collection.

(BTW `SelectWhere` is form Helpful Libraries.)

Lombiq.Tests.UI/Services/UITestExecutor.cs
```diff
@@ -51,15 +51,16 @@ public static class UITestExecutor
...
+        await _numberOfTestsLimitLock.WaitAsync();
+
         if (_numberOfTestsLimit == null && configuration.MaxParallelTests > 0)
         {
-            lock (_numberOfTestsLimitLock)
-            {
-                _numberOfTestsLimit ??= new SemaphoreSlim(configuration.MaxParallelTests);
-            }
+            _numberOfTestsLimit = new SemaphoreSlim(configuration.MaxParallelTests);
         }
```
dministro: [InferSharp ](https://github.com/microsoft/infersharp) reported a thread safety violation here, I can agree with it.
image
Description: Read/Write race. Non-private method `UITestExecutor.ExecuteOrchardCoreTestAsync(...)` reads without synchronization from `Services.UITestExecutor.Lombiq.Tests.UI.Services.UITestExecutor._numberOfTestsLimit`. Potentially races with write in method `UITestExecutor.ExecuteOrchardCoreTestAsync(...)`. Reporting because this access may occur on a background thread.

Piedone: Could you please fix this?

dministro: I fixed it already, just noted why I changed something out of scope.

Piedone: Ah, great! BTW: https://github.com/Lombiq/GitHub-Actions/issues/219

dministro: Agree

Lombiq.Tests.UI/Services/Counters/Configuration/PhaseCounterConfiguration.cs
```diff
@@ -0,0 +1,14 @@
+namespace Lombiq.Tests.UI.Services.Counters.Configuration;
+
+public class PhaseCounterConfiguration : CounterConfiguration
+{
+    /// <summary>
+    /// Gets or sets threshold configuration used under app phase(setup, running) lifetime.
```
Piedone:
```suggestion
    /// Gets or sets threshold configuration used under the app phase (setup, running) lifetime.
```

Lombiq.Tests.UI/Services/PhaseCounterConfiguration.cs
Piedone: This still needs to be done.

Lombiq.Tests.UI/Services/UITestContext.cs
Piedone: This still needs to be done. The page change dumps should also only be written when duplicate SQL query detection is enabled.

Lombiq.Tests.UI.Samples/Tests/BasicOrchardFeaturesTests.cs
```diff
@@ -25,7 +25,17 @@ public BasicOrchardFeaturesTests(ITestOutputHelper testOutputHelper)
...
         ExecuteTestAsync(
             context => context.TestBasicOrchardFeaturesAsync(RecipeIds.BasicOrchardFeaturesTests),
-            browser);
+            browser,
+            configuration =>
+            {
+                // The UI Testing Toolbox includes a DbCommand execution counter. After the end of the test, it checks the
+                // number of executed commands with the same SQL command and parameter set against the threshold value
+                // in its configuration. If the executed command count is greater than the threshold, it raises a
+                // CounterThresholdException. So here we set the minimum required value to avoid it.
```
Piedone:
```suggestion
                // The UI Testing Toolbox includes a DbCommand execution counter to check for duplicated SQL queries..
                // After the end of the test, it checks the number of executed commands with the same SQL command text
                // and parameter set against the threshold value in its configuration. If the executed command count is
                // greater than the threshold, it raises a CounterThresholdException. So here we set the minimum
                // required value to avoid it.
```

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+using Shouldly;
+using System;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Lombiq.Tests.UI.Samples.Tests;
+
+// Some times you may want to detect duplicated SQL queries. This can be useful if you want to make sure that your code
+// does not execute the same query multiple times.
```
Piedone:
```suggestion
// does not execute the same query multiple times, wasting time and computing resources.
```

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+public class DuplicatedSqlQueryDetectorTests : UITestBase
+{
+    public DuplicatedSqlQueryDetectorTests(ITestOutputHelper testOutputHelper)
+        : base(testOutputHelper)
+    {
+    }
+
+    // This test will fail because the app will read the the same command result more times then the configured threshold
+    // during the Admin page rendering.
```
Piedone:
```suggestion
    // This test will fail because the app will read the same command result more times than the configured threshold
    // during the Admin page rendering.
```

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+    // This test will pass because not the Admin page was loaded.
+    [Theory, Chrome]
+    public Task DbReaderReadDuringRunningPhaseShouldNotThrow(Browser browser) =>
+        Should.NotThrowAsync(() =>
+            ExecuteTestAfterSetupAsync(
+                async context => await context.GoToHomePageAsync(onlyIfNotAlreadyThere: false),
+                browser,
+                ConfigureAsync));
+
+    // We configure the test to throw an exception if a certain counter threshold is exceeded, but only in case of Admin page.
```
Piedone:
```suggestion
    // We configure the test to throw an exception if a certain counter threshold is exceeded, but only in case of Admin
    // pages.
```

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+    private static Task ConfigureAsync(OrchardCoreUITestExecutorConfiguration configuration)
+    {
+        // The test is guaranteed to fail so we don't want to retry it needlessly.
+        configuration.MaxRetryCount = 0;
+
+        var adminCounterConfiguration = new CounterConfiguration();
+        // Let's enable and configure the counter threshold for ORM sessions.
+        adminCounterConfiguration.SessionThreshold.Disable = false;
+        adminCounterConfiguration.SessionThreshold.DbReaderReadThreshold = 0;
+        // Apply the configuration to the Admin page only.
```
Piedone:
```suggestion
        // Apply the configuration to the Admin pages only.
```

Lombiq.Tests.UI/Services/Counters/Configuration/CounterThresholdConfiguration.cs
```diff
@@ -0,0 +1,29 @@
+namespace Lombiq.Tests.UI.Services.Counters.Configuration;
+
+public class CounterThresholdConfiguration
+{
+    /// <summary>
+    /// Gets or sets a value indicating whether the current threshold configuration and checking is disabled.
+    /// </summary>
+    public bool Disable { get; set; } = true;
```
Piedone: Use `IsEnabled` instead. This is more commonly used in APIs. Also, why is this property necessary in the first place? When would you add a configuration but have it disabled?

Lombiq.Tests.UI/Services/Counters/Configuration/CounterThresholdConfiguration.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    /// <summary>
+    /// Gets or sets the threshold of executed <see cref="System.Data.Common.DbCommand"/>s. Uses
+    /// <see cref="System.Data.Common.DbCommand.CommandText"/> and <see cref="System.Data.Common.DbCommand.Parameters"/>
+    /// for counting.
+    /// </summary>
+    public int DbCommandExecutionThreshold { get; set; } = 11;
```
Piedone: What are the magic 11s?

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+        Should.ThrowAsync<AggregateException>(() =>
+            ExecuteTestAfterSetupAsync(
+                context => context.SignInDirectlyAndGoToDashboardAsync(),
+                browser,
+                ConfigureAsync));
+
+    // This test will pass because not the Admin page was loaded.
+    [Theory, Chrome]
+    public Task DbReaderReadDuringRunningPhaseShouldNotThrow(Browser browser) =>
+        Should.NotThrowAsync(() =>
```
Piedone: `Should.NotThrowAsync()` is not needed.

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+    public Task DbReaderReadDuringRunningPhaseShouldThrow(Browser browser) =>
+        Should.ThrowAsync<AggregateException>(() =>
+            ExecuteTestAfterSetupAsync(
+                context => context.SignInDirectlyAndGoToDashboardAsync(),
+                browser,
+                ConfigureAsync));
+
+    // This test will pass because not the Admin page was loaded.
+    [Theory, Chrome]
+    public Task DbReaderReadDuringRunningPhaseShouldNotThrow(Browser browser) =>
```
Piedone: Demonstrate adding a config for this page that has a threshold that matches the page.

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    public DuplicatedSqlQueryDetectorTests(ITestOutputHelper testOutputHelper)
+        : base(testOutputHelper)
+    {
+    }
+
+    // This test will fail because the app will read the the same command result more times then the configured threshold
+    // during the Admin page rendering.
+    [Theory, Chrome]
+    public Task DbReaderReadDuringRunningPhaseShouldThrow(Browser browser) =>
```
Piedone:
```suggestion
    public Task PageWithTooManyDuplicatedSqlQueriesShouldThrow(Browser browser) =>
```

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
Piedone:
```suggestion
    public Task PageWithoutDuplicatedSqlQueriesShouldPass(Browser browser) =>
```

Lombiq.Tests.UI.Samples/Tests/BasicOrchardFeaturesTests.cs
```diff
@@ -32,7 +32,7 @@ public BasicOrchardFeaturesTests(ITestOutputHelper testOutputHelper)
                 // number of executed commands with the same SQL command and parameter set against the threshold value
                 // in its configuration. If the executed command count is greater than the threshold, it raises a
                 // CounterThresholdException. So here we set the minimum required value to avoid it.
-                configuration.CounterConfiguration.Running.DbCommandExecutionRepetitionThreshold = 26;
+                configuration.CounterConfiguration.Running.PhaseThreshold.DbCommandExecutionThreshold = 26;
```
Piedone: There are 26 duplicated queries during this test?

Lombiq.Tests.UI/Services/Counters/IRelativeUrlConfigurationKey.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+
+namespace Lombiq.Tests.UI.Services.Counters;
+
+/// <summary>
+/// Represents a relative URL configuration key in <see cref="Configuration.RunningPhaseCounterConfiguration"/>.
+/// </summary>
+public interface IRelativeUrlConfigurationKey : ICounterConfigurationKey
```
Piedone, line 8: This would be more straightforward with regex matching.

Lombiq.Tests.UI/Services/Counters/SessionProbe.cs
```diff
@@ -0,0 +1,71 @@
...
+    Task<DbTransaction> ISession.BeginTransactionAsync() => _session.BeginTransactionAsync();
+    Task<DbTransaction> ISession.BeginTransactionAsync(IsolationLevel isolationLevel) =>
+        _session.BeginTransactionAsync(isolationLevel);
+    ISession ISession.RegisterIndexes(IIndexProvider[] indexProviders, string collection) =>
+        _session.RegisterIndexes(indexProviders, collection);
+    async ValueTask IAsyncDisposable.DisposeAsync()
+    {
+        await _session.DisposeAsync();
+        // Should be at the end because, the Session implementation calls CommitOrRollbackTransactionAsync in DisposeAsync
+        // and we should count the executed db commands in it.
```
Piedone:
```suggestion
        // Should be at the end because the Session implementation calls CommitOrRollbackTransactionAsync in
        // DisposeAsync and we should count the executed DB commands in it.
```

Lombiq.Tests.UI.Samples/Tests/BasicOrchardFeaturesTests.cs
Piedone: I'm not sure if this is addressed. Aren't still only duplications with the same query and parameters counted?

Lombiq.Tests.UI.Samples/Tests/BasicOrchardFeaturesTests.cs
```diff
@@ -25,7 +25,17 @@ public BasicOrchardFeaturesTests(ITestOutputHelper testOutputHelper)
...
             context => context.TestBasicOrchardFeaturesAsync(RecipeIds.BasicOrchardFeaturesTests),
-            browser);
+            browser,
+            configuration =>
+            {
+                // The UI Testing Toolbox includes a DbCommand execution counter. After the end of the test, it checks the
+                // number of executed commands with the same SQL command and parameter set against the threshold value
+                // in its configuration. If the executed command count is greater than the threshold, it raises a
+                // CounterThresholdException. So here we set the minimum required value to avoid it.
+                configuration.CounterConfiguration.Running.PhaseThreshold.DbCommandExecutionThreshold = 26;
```
Piedone: Why is this configuration needed in this test?

Lombiq.Tests.UI/Exceptions/CounterThresholdException.cs
Piedone: While the current message is better, it's still cryptic when you first see it.

> SessionProbe, [GET]https://localhost:9341/Admin
Database reader read counter
	Query: SELECT [Document].* FROM [Document] WHERE [Document].[Type] = @Type LIMIT 1
		Parameters: [0]Type = OrchardCore.AdminMenu.Models.AdminMenuList, OrchardCore.AdminMenu
Count: 2
Counter value is greater then SessionThreshold.DbReaderReadThreshold, threshold: 0.

Start the exception with a message that explains, in plain English, what happened.

Lombiq.Tests.UI/Services/Counters/Configuration/CounterConfiguration.cs
```diff
@@ -0,0 +1,86 @@
...
+                new("p0", "ContentPublishedEvent"),
+                new("p1", value: true),
+            }),
+        new DbCommandExecuteCounterKey(
+            WorkflowTypeStartActivitiesQuery,
+            new List<KeyValuePair<string, object>>
+            {
+                new("p0", "ContentUpdatedEvent"),
+                new("p1", value: true),
+            }),
```
Piedone: Why are these excluded?

Lombiq.Tests.UI/Services/Counters/ICounterDataCollector.cs
```diff
@@ -26,4 +28,9 @@ public interface ICounterDataCollector : ICounterProbe
     /// Asserts the data collected by the instance.
     /// </summary>
     void AssertCounter();
+
+    /// <summary>
+    /// Postpones exception thrown by a counter in case when the exception was thrown out of the test context.
```
Piedone:
```suggestion
    /// Postpones exception thrown by a counter when the exception was thrown from the test context.
```

Lombiq.Tests.UI/Services/Counters/ICounterKey.cs
```diff
@@ -8,8 +9,13 @@ namespace Lombiq.Tests.UI.Services.Counters;
...
+    /// Gets the display name of the key.
     /// </summary>
-    /// <returns>A human readable string representation of instance.</returns>
-    string Dump();
+    string DisplayName { get; }
+
+    /// <summary>
+    /// Dumps the key content to a human-readable format.
+    /// </summary>
+    /// <returns>A human-readable representation of instance.</returns>
```
Piedone:
```suggestion
    /// <returns>A human-readable representation of the instance.</returns>
```

Lombiq.Tests.UI/Services/Counters/IRelativeUrlConfigurationKey.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// Gets the URL.
+    /// </summary>
+    Uri Url { get; }
+
+    /// <summary>
+    /// Gets a value indicating whether the URL should be matched exactly or substring match enough.
```
Piedone:
```suggestion
    /// Gets a value indicating whether the URL should be matched exactly or substring match is enough.
```

Lombiq.Tests.UI/Services/Counters/Middlewares/PageLoadProbeMiddleware.cs
```diff
@@ -0,0 +1,24 @@
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Extensions;
+using System;
+using System.Threading.Tasks;
+
+namespace Lombiq.Tests.UI.Services.Counters.Middlewares;
+
+public class PageLoadProbeMiddleware
```
Piedone: Why is this and `NavigationProbe` both necessary?

Lombiq.Tests.UI/Services/ProbedConnectionFactory.cs
```diff
@@ -19,6 +20,14 @@ public ProbedConnectionFactory(IConnectionFactory connectionFactory, ICounterDat
...
     }
 
-    public DbConnection CreateConnection() =>
-        new ProbedDbConnection(_connectionFactory.CreateConnection(), _counterDataCollector);
+    public DbConnection CreateConnection()
+    {
+        var connection = _connectionFactory.CreateConnection();
+
+        // This consition and the ProbedSqliteConnection can be removed once
+        // https://github.com/OrchardCMS/OrchardCore/issues/14217 get fixed.
```
Piedone:
```suggestion
        // This condition and the ProbedSqliteConnection can be removed once
        // https://github.com/OrchardCMS/OrchardCore/issues/14217 gets fixed.
```

Piedone (changes requested):

Lombiq.Tests.UI.Samples/Tests/BasicOrchardFeaturesTests.cs
dministro: You can configure both. `CounterThresholdConfiguration.DbCommandExecutionThreshold` is used against the count of the same command text with the same parameter value(s), while `CounterThresholdConfiguration.DbCommandTextExecutionThreshold` is used against the count of the same command text.

Lombiq.Tests.UI/Services/Counters/Middlewares/PageLoadProbeMiddleware.cs
dministro: The `NavigationProbe` is for the navigation requested by the test code e.g. `UITestContext.GoToHomePageAsync(..)` while the `PageLoadProbe` is for the requests from the web browser e.g. image requests, API calls, etc. The `NavigationProbe` captures image requests while the page loading, but not async calls.

Piedone: Wouldn't `PageLoadProbe` alone enough? Since all navigations requested by the test will in the end result in a request.

BTW I'd rather call this `RequestLoadProbleMiddleware` then.

Lombiq.Tests.UI/Services/Counters/Configuration/CounterThresholdConfiguration.cs
Piedone:
```suggestion
    public int DbCommandIncludingParametersExecutionCountThreshold { get; set; } = 11;
```

Lombiq.Tests.UI/Services/Counters/Configuration/CounterThresholdConfiguration.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    /// <summary>
+    /// Gets or sets the threshold of executed <see cref="System.Data.Common.DbCommand"/>s. Uses
+    /// <see cref="System.Data.Common.DbCommand.CommandText"/> for counting.
+    /// </summary>
+    public int DbCommandTextExecutionThreshold { get; set; } = 11;
```
Piedone:
```suggestion
    public int DbCommandExcludingParametersExecutionThreshold { get; set; } = 11;
```

Lombiq.Tests.UI/Services/Counters/Configuration/CounterThresholdConfiguration.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// <summary>
+    /// Gets or sets a value indicating whether the current threshold configuration and checking is disabled.
+    /// </summary>
+    public bool Disable { get; set; } = true;
+
+    /// <summary>
+    /// Gets or sets the threshold of executed <see cref="System.Data.Common.DbCommand"/>s. Uses
+    /// <see cref="System.Data.Common.DbCommand.CommandText"/> and <see cref="System.Data.Common.DbCommand.Parameters"/>
+    /// for counting.
+    /// </summary>
```
Piedone:
```suggestion
    /// <summary>
    /// Gets or sets the threshold for the count of <see cref="System.Data.Common.DbCommand"/> executions, with the
    /// query only counted as a duplicate if both its text (<see cref="System.Data.Common.DbCommand.CommandText"/>) and
    /// parameters (<see cref="System.Data.Common.DbCommand.Parameters"/>) match. See
    /// <see cref="DbCommandTextExecutionThreshold"/> for counting using only the command text.
    /// </summary>
```

Lombiq.Tests.UI/Services/Counters/Configuration/CounterThresholdConfiguration.cs
```diff
@@ -0,0 +1,29 @@
...
+    public int DbCommandExecutionThreshold { get; set; } = 11;
+
+    /// <summary>
+    /// Gets or sets the threshold of executed <see cref="System.Data.Common.DbCommand"/>s. Uses
+    /// <see cref="System.Data.Common.DbCommand.CommandText"/> for counting.
+    /// </summary>
```
Piedone:
```suggestion
    /// <summary>
    /// Gets or sets the threshold for the count of <see cref="System.Data.Common.DbCommand"/> executions, with the
    /// query counted as a duplicate if its text (<see cref="System.Data.Common.DbCommand.CommandText"/>) matches.
    /// Parameters (<see cref="System.Data.Common.DbCommand.Parameters"/>) are not taken into account. See
    /// <see cref="DbCommandExecutionThreshold"/> for counting using also the parameters.
    /// </summary>
```

Lombiq.Tests.UI.Samples/Tests/BasicOrchardFeaturesTests.cs
Piedone: I see, OK then.

Piedone:
@coderabbitai review

I'm testing the AI code review tool [CodeRabbit](https://coderabbit.ai/) here. Please bear with me. And if it comments some useful feedback, please address it.

Piedone:
@coderabbitai review

Sorry about that.

Piedone:
These comments seem useful, please check them out. You can also chat with the bot, see the tips under https://github.com/Lombiq/UI-Testing-Toolbox/pull/216#issuecomment-1872376517.

Lombiq.Tests.UI/Services/Counters/Configuration/CounterConfiguration.cs
```diff
@@ -0,0 +1,86 @@
...
+
+    /// <summary>
+    /// Gets or sets threshold configuration used per page load. See: <see cref="PageLoadProbe"/>.
+    /// </summary>
+    public CounterThresholdConfiguration PageLoadThreshold { get; set; } = new CounterThresholdConfiguration
+    {
+        DbCommandIncludingParametersExecutionCountThreshold = 22,
+        DbCommandExcludingParametersExecutionThreshold = 44,
+        DbReaderReadThreshold = 11,
+    };
```
Piedone: Please add a piece of docs contrasting these three configs: when would you want to use each?

Lombiq.Tests.UI/Services/Counters/Middlewares/PageLoadProbeMiddleware.cs
dministro: Yes, it results at least one request but the kex difference is not just the operations counted during the page generation but all operations (eg image - included into the page - loaded from the DB). So it counts SQL operations for multiple request connected for the given page load.

Lombiq.Tests.UI/Services/Counters/Configuration/CounterConfiguration.cs
dministro: Happening many times during the setup, so I found them useless. Used only in samples, so I can move there, and double check it.

Piedone: That would be better.

Lombiq.Tests.UI/Services/Counters/Middlewares/PageLoadProbeMiddleware.cs
Piedone: kex difference?

image

But OK then :).

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
Piedone, line 1: These tests log failures much like actually failing tests, causing error outputs and tests dumps, see e.g. https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/16277477256. Instead of having failing `ExecuteTestAfterSetupAsync`s, do what `ErrorHandlingTests` does and catch the assertion exceptions.

Lombiq.Tests.UI/Docs/CounterInfrastructure.md
```diff
@@ -0,0 +1,41 @@
...
+- `DbCommandExcludingParametersExecutionThreshold` – Compared against the internal counter, which increments each time a specific SQL command is executed, regardless of the parameter set.
+- `DbReaderReadThreshold` – Compared against the internal counter, which increments each time a read operation is performed on a SQL result set.
+
+### Lifecycles Where You Can Configure Thresholds
+
+Threshold values are configured in: [`CounterConfiguration`](../Services/Counters/Configuration/CounterConfiguration.cs).
+
+- **Navigation:** Captures operations performed during page load, including SQL queries executed for page generation and content retrieval (e.g., images from the database). The configuration is stored in `NavigationThreshold`.
+- **Page Load:** Captures operations performed strictly during page load. The configuration is stored in `PageLoadThreshold`.
+- **Session:** Captures operations executed while a given `ISession` instance is alive (e.g., a session used by a background task). The configuration is stored in `SessionThreshold`.
```
Piedone: Instead of explaining the properties here, merge this with the docs on the properties, so ultimately, the properties have the complete docs. Then, here just offer a quick shorter rundown of 3-3 bullet points on what the counters count, and when they are asserted, referring people to read the docs on the classes for in-depth info.

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
```diff
@@ -0,0 +1,141 @@
...
+            new("p1", value: true)),
+    ];
+
+    public DuplicatedSqlQueryDetectorTests(ITestOutputHelper testOutputHelper)
+        : base(testOutputHelper)
+    {
+    }
+
+    // This test will fail because the app will read the same command result more times than the configured threshold
+    // during the Admin page rendering.
```
Piedone, line 52: This is a technical description of what happens in the counter logic. This is good, but add an explanation of why this happens in higher-level terms. I.e., there's a query on the OC admin that's duplicated or something, mention that.

Same for the other two ShouldThrow samples.

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
```diff
@@ -0,0 +1,141 @@
...
+            new("p1", value: true)),
+        new DbCommandExecuteCounterKey(
+            WorkflowTypeStartActivitiesQuery,
+            new("p0", "ContentPublishedEvent"),
+            new("p1", value: true)),
+        new DbCommandExecuteCounterKey(
+            WorkflowTypeStartActivitiesQuery,
+            new("p0", "ContentUpdatedEvent"),
+            new("p1", value: true)),
+    ];
```
Piedone, line 44: Add a comment about what this is.

Lombiq.Tests.UI/Exceptions/CounterThresholdException.cs
Piedone: This isn't addressed.

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
```diff
@@ -0,0 +1,141 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Lombiq.Tests.UI.Samples.Tests;
+
+// Some times you may want to detect duplicated SQL queries. This can be useful if you want to make sure that your code
+// does not execute the same query multiple times, wasting time and computing resources.
```
Piedone: I added a note here, but explain how you'd start using this in your app. E.g., do you just start using it all of your tests with the defaults? Or only on the frontend?

Lombiq.Tests.UI/Services/Counters/IRelativeUrlConfigurationKey.cs
Piedone: This isn't addressed.

Lombiq.Tests.UI/Services/Counters/Configuration/CounterThresholdConfiguration.cs
Piedone: You didn't answer here. Why do we have both this and `CounterConfigurations.IsEnabled`?

Lombiq.Tests.UI/Services/Counters/Configuration/CounterConfiguration.cs
Piedone: This isn't addressed.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -51,9 +52,12 @@ public static Task GoToAbsoluteUrlAsync(this UITestContext context, Uri absolute
                 await context.Configuration.Events.BeforeNavigation
                     .InvokeAsync<NavigationEventHandler>(eventHandler => eventHandler(context, absoluteUri));
 
-                // Navigation can sometimes not happen on the first try.
-                await context.DoWithRetriesUntilNavigationHasOccurredOrFailAsync(
-                    () => context.Driver.Navigate().GoToUrlAsync(absoluteUri));
+                using (new NavigationProbe(context.CounterDataCollector, absoluteUri))
```
Piedone, line 55: Do we always need a `NavigationProbe` instantiation here, even if the test suite doesn't use the counters at all? Can't this be done with an optional `NavigationEventHandler`?

Lombiq.Tests.UI/Services/Counters/Configuration/PhaseCounterConfiguration.cs
```diff
@@ -0,0 +1,14 @@
...
+public class PhaseCounterConfiguration : CounterConfiguration
+{
+    /// <summary>
+    /// Gets or sets threshold configuration used under the app phase (setup, running) lifetime.
+    /// </summary>
+    public CounterThresholdConfiguration PhaseThreshold { get; set; } = new CounterThresholdConfiguration
+    {
+        DbCommandIncludingParametersExecutionCountThreshold = 22,
+        DbCommandExcludingParametersExecutionThreshold = 44,
+        DbReaderReadThreshold = 11,
```
Piedone, line 12: Where do these numbers come from?

Lombiq.Tests.UI/Services/OrchardCoreHosting/OrchardApplicationFactory.cs
```diff
@@ -103,8 +115,13 @@ .ImplementationInstance as ConfigurationManager
             {
                 AddFakeStore(builderServices);
                 AddFakeViewCompilerProvider(builderServices);
+                AddSessionProbe(builderServices);
```
Piedone, line 118: Anything related to counters should only be done here if they're actually enabled. So, the session probe and middleware shouldn't be added if it's disabled, nor `ProbedConnectionFactory`.

Lombiq.Tests.UI/Services/OrchardCoreUITestExecutorConfiguration.cs
```diff
@@ -211,4 +223,78 @@ public CancellationToken TestCancellationToken
...
+                if (probe.Counters[key] is IntegerCounterValue counterValue
+                    && counterValue.Value > threshold)
+                {
+                    throw new CounterThresholdException(
+                        probe,
+                        key,
+                        counterValue,
+                        $"Counter value is greater then {thresholdName}, threshold: {threshold.ToTechnicalString()}.");
+                }
+            });
```
Piedone, line 299: I don't think these need to be in this class.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -119,6 +120,10 @@ public async Task<bool> ExecuteAsync(int retryCount, string dumpRootPath)
             _context.TestDumpContainer.Clear();
             testDumpContainer = _context.TestDumpContainer;
 
+            BeginDataCollection(
```
Piedone, line 123: Only run this and `EndAssertDataCollection` when the counters are enabled.

Piedone (changes requested):
Be sure to address the older comments too.

Lombiq.Tests.UI/Docs/CounterInfrastructure.md
dministro: I moved the CounterThresholdConfiguration property documentation to the code, but the lifecycle description is better here.

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
dministro: Sorry, but the current comment is what you asked. What higher level do you think?

Lombiq.Tests.UI/Services/Counters/IRelativeUrlConfigurationKey.cs
dministro: I feel it is a bit overengineering. We can do it later under a different PR if it really necessary.

dministro:
I can't answer under your comments:
https://github.com/Lombiq/UI-Testing-Toolbox/pull/216#discussion_r1370964780
https://github.com/Lombiq/UI-Testing-Toolbox/pull/216#discussion_r2208939079

It was addressed, see:
```
2025-07-16T01:22:23.9882241Z  A counter value has crossed the configured threshold level. Details:
2025-07-16T01:22:23.9882342Z  SessionProbe, [GET]https://localhost:9031/Admin
2025-07-16T01:22:23.9882454Z  Database reader read counter
2025-07-16T01:22:23.9882629Z  	Query: SELECT [Document].* FROM [Document] WHERE [Document].[Type] = @Type LIMIT 1
2025-07-16T01:22:23.9882814Z  		Parameters: [0]Type = OrchardCore.AdminMenu.Models.AdminMenuList, OrchardCore.AdminMenu
2025-07-16T01:22:23.9882866Z  Count: 2
2025-07-16T01:22:23.9883060Z  Counter value is greater then SessionThreshold.DbReaderReadThreshold, threshold: 0.
```

Feel free to add more explanation.

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
Piedone: What causes this, what's the command, what's the admin actually doing to read more?

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
dministro: Comment added to where the configuration is happening.

Piedone:
You can reply there, exactly under the link, so I'll also follow up there.

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
dministro: Sorry, but I don't understand what you are requesting. This feature is connected to the persistent layer, so how the frontend is coming to the picture?

Lombiq.Tests.UI/Exceptions/CounterThresholdException.cs
dministro: It was addressed, see:
```
2025-07-16T01:22:23.9882241Z  A counter value has crossed the configured threshold level. Details:
2025-07-16T01:22:23.9882342Z  SessionProbe, [GET]https://localhost:9031/Admin
2025-07-16T01:22:23.9882454Z  Database reader read counter
2025-07-16T01:22:23.9882629Z  	Query: SELECT [Document].* FROM [Document] WHERE [Document].[Type] = @Type LIMIT 1
2025-07-16T01:22:23.9882814Z  		Parameters: [0]Type = OrchardCore.AdminMenu.Models.AdminMenuList, OrchardCore.AdminMenu
2025-07-16T01:22:23.9882866Z  Count: 2
2025-07-16T01:22:23.9883060Z  Counter value is greater then SessionThreshold.DbReaderReadThreshold, threshold: 0.
```

Feel free to add more explanation.

Piedone: Re: https://github.com/Lombiq/UI-Testing-Toolbox/pull/216#issuecomment-3085570127

This is still very technical. "A counter value has crossed the configured threshold level." tells what happened internally. It's like a debug message for the counter infrastructure. But it doesn't help the developer, who'll see their test fail because of this, and who might have not configured this, what actually the problem is.

E.g. something like "The same database query was executed more times than the threshold configuration allows, potentially causing performance issues with excessive database usage. This can indicate a SELECT N+1 problem, or a case where the same query is executed from multiple places in the code instead of using a common cache. See the details below."

And similarly for all kinds of failures.

Lombiq.Tests.UI.Samples/Tests/DuplicatedSqlQueryDetectorTests.cs
Piedone: By frontend I mean the public-facing part of the site, i.e. not the admin.

Lombiq.Tests.UI/Services/Counters/Configuration/CounterThresholdConfiguration.cs
dministro: We can disable the assertion for the given "lifecycle". (Navigation, Session, PageLoad)

Lombiq.Tests.UI/Services/Counters/IRelativeUrlConfigurationKey.cs
Piedone: Why overengineering? We're matching a URL against a substring, with an additional Boolean switch (`ExactMatch`). Matching it against a regex is a very small next step, but provides ultimate flexibility, while removing the switch.

I've written this almost two years ago, we won't defer this anymore.

Lombiq.Tests.UI/Services/Counters/Configuration/CounterThresholdConfiguration.cs
Piedone: I see, OK then.

Piedone:
Somebody else will do this.

PR #224, opened 2022-10-27
OSOE-424: Make setup compatible with any and environment-configured DB provider

### LINKED ISSUE #223: Make setup compatible with any and environment-configured DB provider (OSOE-424)
`OrchardCoreSetupPage.SetupOrchardCoreAsync()` needs the DB provider supplied as a parameter. If it comes from the environment, setup will fail. Fix this to allow environmental DB configuration.

Furthermore, add what else may be necessary for https://github.com/OrchardCMS/OrchardCore/pull/11194 to support all the DB providers, without snapshotting.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-424)

Piedone:
OSOE-424
Fixes #223.

### DIFF
--- a/Lombiq.Tests.UI/Pages/OrchardCoreSetupPage.cs
+++ b/Lombiq.Tests.UI/Pages/OrchardCoreSetupPage.cs
@@ -108,7 +108,7 @@ public sealed class OrchardCoreSetupPage : Page<_>
             {
                 throw new InvalidOperationException(
                     $"{nameof(OrchardCoreSetupParameters)}.{nameof(parameters.DatabaseProvider)}: " +
-                    "If the selected database provider is other than SQLite a connection string must be provided.");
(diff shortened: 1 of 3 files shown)


Lombiq.Tests.UI/Pages/OrchardCoreSetupPage.cs
```diff
@@ -29,8 +29,10 @@ public enum DatabaseType
         [Term("Sql Server")]
         SqlServer,
         Sqlite,
+        [Term("MySql")]
```
dministro, line 32: Nice catch.

dministro (approved):

Lombiq.Tests.UI/Pages/OrchardCoreSetupPage.cs
Piedone: Thanks!

PR #225, opened 2022-11-05
OSOE-392: Upgrade to Orchard Core 1.5 in UI-Testing-Toolbox

MZole:
[OSOE-392](https://lombiq.atlassian.net/browse/OSOE-392)


Lombiq.Tests.UI/Lombiq.Tests.UI.csproj
```diff
@@ -64,13 +64,11 @@
     <PackageReference Include="Microsoft.SqlServer.SqlManagementObjects" Version="170.8.0" />
     <PackageReference Include="Mono.Posix.NETStandard" Version="5.20.1-preview" />
     <!-- During Orchard upgrades, this needs to be kept up to date with the version Orchard uses. -->
-    <PackageReference Include="NLog.Web.AspNetCore" Version="4.14.0" />
-    <PackageReference Include="OrchardCore.Abstractions" Version="1.4.0" />
+    <PackageReference Include="NLog.Web.AspNetCore" Version="5.1.5" />
```
Piedone: Let's try to use `OrchardCore.Logging.NLog` here instead, it's more straightforward to keep in sync with the rest of the OC packages.

Lombiq.Tests.UI/Services/OrchardCoreInstance.cs
```diff
@@ -136,7 +136,7 @@ public IEnumerable<IApplicationLog> GetLogs(CancellationToken cancellationToken
                 {
                     Name = Path.GetFileName(filePath),
                     FullName = Path.GetFullPath(filePath),
-                    ContentLoader = () => File.ReadAllTextAsync(filePath, cancellationToken),
+                    ContentLoader = async () => await File.ReadAllTextAsync(filePath, cancellationToken),
```
Piedone: This is in effect the same, why is it necessary?

Piedone (changes requested):

Lombiq.Tests.UI/Services/OrchardCoreInstance.cs
```diff
@@ -136,11 +134,18 @@ public IEnumerable<IApplicationLog> GetLogs(CancellationToken cancellationToken
...
+                    Content = GetFileContent(filePath),
                 })
             : Enumerable.Empty<IApplicationLog>();
     }
 
+    public string GetFileContent(string filePath)
+    {
+        using var fileStream = new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
+        using var streamReader = new StreamReader(fileStream);
+        return streamReader.ReadToEnd();
```
Piedone: Use `ReadToEndAsync()` and revert everything back to `async`.

Piedone (changes requested):

Piedone (approved):

Lombiq.Tests.UI/Services/OrchardCoreInstance.cs
```diff
@@ -136,11 +134,20 @@ public IEnumerable<IApplicationLog> GetLogs(CancellationToken cancellationToken
...
                     Name = Path.GetFileName(filePath),
                     FullName = Path.GetFullPath(filePath),
-                    ContentLoader = () => File.ReadAllTextAsync(filePath, cancellationToken),
+                    ContentLoader = () => GetFileContentAsync(filePath),
                 })
             : Enumerable.Empty<IApplicationLog>();
     }
 
+    // This needs to be public and instance-level. If it's private static then it causes the app under test not to
+    // launch under Ubuntu for some reason, while working under Windows.
```
Piedone: There's something strange.

- In [this commit](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/264/commits/52449b4234e02cade5f74749480001000a21ffab) the Ubuntu UI testing failed, all tests failing due to the app starting with a blank screen (no error, no anything).
- Reverting my changes with [this commit](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/264/commits/f78af4dcc5bcc44d5da56f644e4a08a2b97a6b86) caused the build to pass, repeatedly.
- In [this commit](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/264/commits/0a2ce1bc2f5d3ead993db9b8c202731e719a6061) I re-added [my UI Testing Toolbox changes](https://github.com/Lombiq/UI-Testing-Toolbox/compare/c21ff811ba3f2b4013c05acf55e0125f07e325f8...0ca22dd4edaf509d8ba259c50cbbb2f1faf7624d), which caused UI testing to fail again in the same way.

So, it's some quirk with `GetFileContentAsync()` having to be `public` and instance-level? No:

- In [this commit](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/264/commits/53bf265989a0857add91948d9c0a1fc5b9b29f7f) I've only added a comment to the original code of yours. However, it still failed UI testing in the same way, but this time under Windows. Re-running fixed it.
- The changes [in this commit](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/264/commits/b19ba303778e4feb3d12a2297d3d8722123e3c41) failed the build first, with re-run fixing it, while [this](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/264/commits/99cc9d8c0947fbbc30018e41e6b455d4a0a4b2f3), [this](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/264/commits/257144d981cab59b81c5c790458f6f7c44040a71), and [this](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/264/commits/01162c6996c24573c07e3725b3893b25427165ca).
- Removing all Obsolete members (also `IsNotFoundMessage()`, I messed up the commit) [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/264/commits/c94efc4108a7de0896b1fe105d27b9aa7adc2024) first passed, then a rebuild of it failed for Ubuntu [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3445688641/jobs/5749693392), then failed for Windows [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3445690510/jobs/5749696812).
- Finally, leaving in `IsNotFoundMessage()` made the build pass three times, [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3445707723), [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3445708910/jobs/5749738662), and [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3445710905/jobs/5749742281).

So, this rather seems random, and not actually connected to a code change (these code changes shouldn't cause UI tests to fail, especially not by the app not starting, anyway), that's just a red herring. This method being non-`static` and `public` is most possibly one too, it can be `private static`.

Since you also confirmed that such builds sometimes needed a rebuild otherwise too, please investigate (such failures don't happen on `dev`). Manually running a rebuild that fixes random errors should never be necessary.

To have a faster feedback loop, when running builds for this, disable (comment out) the NuGetTest build in `build-and-test`, as well as set `build-enable-code-analysis` to `"false"`, and [filter](https://learn.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests?pivots=mstest) for just `BasicTests` and `AzureBlobStorageTests` (it's important that at >=3 tests are run with >=2 different test classes) in _Invoke-SolutionTests.ps1_ of `test-dotnet`, so not the whole test suite is run (and retried 2 more times).

Piedone (changes requested):

Lombiq.Tests.UI/Services/OrchardCoreInstance.cs
MZole: There are actually browser log errors, I downloaded the failure dumps and every test and attempt has the following log:
`[Severe] https://localhost:9066/ - Failed to load resource: the server responded with a status of 500 ()`
Based on this, do you have any suggestions on what should I look into a bit more? Tbh. I'm a bit lost with this one.

Piedone: That's a server-side error, just shows up in the browser log due to a non-2xx response code (accessibility checking failing is a similar red herring). These should also have corresponding Orchard log entries, which they don't. So, my hunch is that something is wrong with the log configuration as well.

I don't have a ready solution either. However, try to temporarily go back to the _NLog.config_ and `OrchardCoreInstance` version you used before (which we know is not suitable in the long run, with the locking log read and all), and let's see if there will be any logs then. If still not, then you can try increasing `LogLevel` to `Information` in _appsettings.json_, and see if the resulting logs contain any hints (if there are still no logs in the failure dump, then something is really messed up with logging; though it shouldn't be completely wrong, since `ErrorHandlingTests` pass). Naturally, try these out locally first before you commit.

Piedone: See https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows+and+actions#Temporary-configuration-during-troubleshooting-for-faster-builds for further tips.

Piedone: You can also use the workflow from the `issue/OSOE-433` branch with the `test-filter` parameter for easier test filtering.

MZole: I skipped the majority of the tests already, but I will check out the branch later, thanks.
Also, I won't be able to see any logs until [OSOE-466](https://lombiq.atlassian.net/browse/OSOE-466) is done, correct?

Piedone: You will, just it's a bit less convenient. They are still all there, but all at once for all tests, in a _test.out_ file in the dump, or in the workflow output.

MZole: Some information, just keep track of them:

- There were 0 occurrences where both builds failed all the tests, so every time either Windows or Linux passes.
- Couldn't repro this with fewer tests, both builds passed 6/6 times.
- Couldn't repro this locally.

Finally, I managed to log at "info" level, and the log files are getting generated, although there are no warnings/errors in them.
The logs indicate that the features are being downloaded and enabled, and the assets on the homepage are being initialized and getting 200 status codes.
The last two messages in the logs maybe can indicate something useful:
`21:42:30.5875||||OrchardCore.Modules.ModularBackgroundService|INFO|'ModularBackgroundService' is stopping. `
`21:42:30.5875||||OrchardCore.Environment.Shell.Distributed.DistributedShellHostedService|INFO|'DistributedShellHostedService' is stopping. `
Is this something that can guide us somewhere or is this normal?

Piedone: These two log messages mean that the shell is being shut down. Try to correlate the timing of these log entries with when the failing tests were started: If the log entries happen before, then they aren't happening during the current test (i.e. when we see the blank page) but previously, when the setup was run. The setup runs once for all tests (during the first test that reaches that point), then a snapshot is created from it, and that snapshot is used to start all the other tests' apps. The "Starting waiting for the setup operation." and "Finished waiting for the setup operation." messages in the log are about this, when tests wait for the setup to finish (all tests are blocked waiting for the first test to finish running the setup). The single test actually running the setup will also have "Starting setup operation." and "Finished setup operation." messages, in addition to log entries about interacting with the setup screen in between.

MZole: Thank you for explaining it in great detail.
Unfortunately, based on the timestamps these logs aren't happening during the tests, these are the setup logs.
I will probably write a post about this on the dev channel, maybe someone will chime in, because the fact that I can't really progress with this bug and thus I'm blocking other issues is starting to really bother me.

Piedone: And what some after these log entries in the test log? I.e. what happens after this in the test itself? Try to add some debug messages with `ITestOutputHelper` in various places where the UI Testing Toolbox to try to pinpoint what's still reached and what isn't.

When you download the failure dump, it'll also contain the given app instances' folders, with the SQLite DB and all. If you try to run the app locally with that, does that work?

MZole: Yes, the app is running with the generated DB file, everything seems fine with it.
After some debugging, I could find that the second log message is not reached [here](https://github.com/Lombiq/UI-Testing-Toolbox/blob/17574a1c5cc52b3a29bf2e0d5de3bea03b838c88/Lombiq.Tests.UI/Services/UITestExecutionSession.cs#L482). I will try to look into this a bit more.

Piedone: This just means that there's an exception during `GoToRelativeUrlAsync()`, which is the first thing after waiting for the setup operation. This is the exception we see in the build output, the accessibility one, that indicates an empty page.

Lombiq.Tests.UI/Services/OrchardCoreHosting/OrchardApplicationFactory.cs
```diff
@@ -57,7 +65,11 @@ protected override void ConfigureWebHost(IWebHostBuilder builder)
 
                 factory.Configuration.Variables["configDir"] = environment.ContentRootPath;
 
-                loggingBuilder.AddNLogWeb(factory, new NLogAspNetCoreOptions { ReplaceLoggerFactory = true });
+                loggingBuilder.AddNLogWeb(factory, new NLogAspNetCoreOptions
+                {
+                    ReplaceLoggerFactory = true,
+                    RemoveLoggerFactoryFilter = false,
```
snakefoot: When already using `ReplaceLoggerFactory = true` (Replacing Microsoft LoggerFactory with NLog LoggerFactory), then `RemoveLoggerFactoryFilter = false` has no effect, because it is only relevant when using Microsoft LoggerFactory.

MZole: Thank you for the advice!

Lombiq.Tests.UI/Lombiq.Tests.UI.csproj
```diff
@@ -65,9 +65,9 @@
     <PackageReference Include="Mono.Posix.NETStandard" Version="5.20.1-preview" />
     <PackageReference Include="OrchardCore.Logging.NLog" Version="1.5.0" />
     <PackageReference Include="OrchardCore.Abstractions" Version="1.5.0" />
-    <PackageReference Include="Selenium.Axe" Version="4.0.2" />
+    <PackageReference Include="Selenium.Axe" Version="4.0.5" />
     <PackageReference Include="Shouldly" Version="4.1.0" />
-    <PackageReference Include="WebDriverManager" Version="2.16.0" />
+    <PackageReference Include="WebDriverManager" Version="2.16.1" />
```
Piedone: While we're at it please update to 2.16.2.

Piedone (changes requested):

Piedone (approved):

PR #237, opened 2022-11-25
OSOE-471: In GitHub Actions workflows, surface log message if a UI test is retried

### LINKED ISSUE #232: In GitHub Actions workflows, surface log message if a UI test is retried (OSOE-471)
When a UI test fails, it's retried, by default two times. Thus, the test may succeed even if it's flaky. Let's add a warning (or if it's annoying, a notice) log message for such retries when the tests are executed in GitHub Actions, so one can be better aware.

- See how `GitHubAnnotationWriter.Annotate()`, we can use it to write the message. We'd need to be able to turn these on or off similarly to `ErrorInTest()`, see how it's used.
- We already have logging for retries [here](https://github.com/Lombiq/UI-Testing-Toolbox/blob/5d070b36e0cbe99a8bf92aac38f0558b5a876185/Lombiq.Tests.UI/Services/UITestExecutionSession.cs#L137), this needs to be extended.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-471)

Piedone:
[OSOE-471](https://lombiq.atlassian.net/browse/OSOE-471)
Fixes #232

### DIFF
--- a/Lombiq.Tests.UI/Services/UITestExecutionSession.cs
+++ b/Lombiq.Tests.UI/Services/UITestExecutionSession.cs
@@ -6,6 +6,7 @@ using Lombiq.Tests.UI.Exceptions;
 using Lombiq.Tests.UI.Extensions;
 using Lombiq.Tests.UI.Helpers;
 using Lombiq.Tests.UI.Models;
+using Lombiq.Tests.UI.Services.GitHub;
 using Microsoft.VisualBasic.FileIO;
 using Mono.Unix;
 using Newtonsoft.Json.Linq;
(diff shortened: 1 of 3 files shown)


Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -428,6 +425,27 @@ private void CaptureMarkupValidationResults(Exception ex, string debugInformatio
...
+            "The test was attempted {0} time(s). {1} more attempt(s) will be made after waiting {2}.",
+            retryCount + 1,
+            _configuration.MaxRetryCount - retryCount,
+            _configuration.RetryInterval);
+
+        if (_configuration.ExtendGitHubActionsOutput &&
+            _configuration.GitHubActionsOutputConfiguration.EnableTestRetryWarningAnnotations &&
+            GitHubHelper.IsGitHubEnvironment)
+        {
+            new GitHubAnnotationWriter(_testOutputHelper).Annotate(
```
Piedone, line 440: This is how this looks like: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3550298479

PR #247, opened 2022-12-20
OSOE-415: Visual verification update and code refactoring

sarahelsaig:
[OSOE-415](https://lombiq.atlassian.net/browse/OSOE-415)


Piedone (commented):
Is the only difference of images in dimensions? Why was that necessary?

sarahelsaig:
> Is the only difference of images in dimensions? Why was that necessary?

[This is the difference.](https://github.com/Lombiq/Orchard-Base-Theme/pull/52/files#diff-af286eac2bfdeb8096ee5ec58fa4529acffe4808e58fff2e561edbe38f3520d7L31-R33) Apparently you were supposed to place the `.navbar-brand` inside the `.navbar` because the CSS variables are tied to `.navbar`. This is a flex class and has its own vertical margin/padding too which also slightly pushes the rest of the page down a few pixels. It's not wrong for the sample, but the verification images have to be updated.

Piedone (approved):
OK then!

PR #248, opened 2023-01-03
OSOE-497: Better exception message during visual verification testing on baseline-actual dimension mismatch

### LINKED ISSUE #244: Add images to failure dump that fail assertions, e.g. of dimensions (OSOE-497)
When running [BasicVisualVerificationTests.VerifyNavbar()](https://github.com/Lombiq/UI-Testing-Toolbox/blob/d4e82bc8cea01d6f58d32370e73aa10c07ee4f07/Lombiq.Tests.UI.Samples/Tests/BasicVisualVerificationTests.cs#L58) all the while the `.navbar-brand` element has some missing or broken styles, the test will fail like this:

```
Error:
  Shouldly.ShouldAssertException : elementImageOriginal.Height
      should be greater than or equal to
  28
      but was
  18
```
See https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3694863540, Annotations, 6th error for an example.

image

The above error is not very helpful in case you're not already deep inside the concerned area of code. It would be more helpful to have the offending element **screenshots**, the expected and the actual ones, **inside the failure dump** in order to more quickly and easily identify the problematic element on the page.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-497)

Piedone:
[OSOE-497](https://lombiq.atlassian.net/browse/OSOE-497)
Fixes #244.

[OSOE-497]: https://lombiq.atlassian.net/browse/OSOE-497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
@@ -350,11 +350,12 @@ to customize the name of the dump item.";
         this UITestContext context,
         IWebElement element,
         string baselineImagePath,
-        string baselineFileName)
+        string baselineFileName,
+        bool saveToTestFolder = true)
     {
         using var suggestedImage = context.TakeElementScreenshot(element);
 
-        suggestedImage.Save(baselineImagePath, new PngEncoder());
+        if (saveToTestFolder) suggestedImage.Save(baselineImagePath, new PngEncoder());
 
         // Appending suggested baseline image to failure dump too.
         context.AppendFailureDump(
@@ -400,21 +401,29 @@ to customize the name of the dump item.";
         // We take a screenshot of the element area. This will be compared to a baseline image.
         using var elementImageOriginal = context.TakeElementScreenshot(element).ShouldNotBeNull();
 
-        // Checking the size of captured image.
-        try
+        // Checking the dimensions of captured image. This needs to happen before any other diffing, because that can
+        // only be done on images with the same dimensions.
+        var cropWidth = cropRegion.Left + cropRegion.Width;
+        var cropHeight = cropRegion.Top + cropRegion.Height;
+        if (elementImageOriginal.Width < cropWidth || elementImageOriginal.Height < cropHeight)
         {
-            var message = "This happens if the baseline image has larger dimensions than the captured area.";
-            elementImageOriginal.Width.ShouldBeGreaterThanOrEqualTo(cropRegion.Left + cropRegion.Width, message);
-            elementImageOriginal.Height.ShouldBeGreaterThanOrEqualTo(cropRegion.Top + cropRegion.Height, message);
-        }
-        catch
-        {
+            var cropRegionName = regionOfInterest == null ? "baseline image" : "selected region of interest";
+            var message = $"The dimensions of the captured element ({elementImageOriginal.Width.ToTechnicalString()}" +
+                $"px x {elementImageOriginal.Height.ToTechnicalString()}px) are smaller than the dimensions of the " +
+                $"{cropRegionName} ({cropWidth.ToTechnicalString()}px x {cropHeight.ToTechnicalString()}px). This " +
+                "can happen if due to a change in the app the captured element got smaller than before, or if the " +
+                $"{cropRegionName} is mistakenly too large.";
+
             if (approvedContext != null)
             {
-                context.SaveSuggestedImage(element, approvedContext.BaselineImagePath, approvedContext.BaselineFileName);
+                message += " The suggested baseline image with a screenshot of the captured element was saved to the " +
+                    "failure dump. Compare this with the original image used by the test and if suitable, use it as " +
+                    "the baseline going forward.";
+                context.SaveSuggestedImage(
+                    element, approvedContext.BaselineImagePath, approvedContext.BaselineFileName, saveToTestFolder: false);
             }
 
-            throw;
+            throw new VisualVerificationAssertionException(message);
         }
 
         using var baselineImageOriginal = baseline.Clone();
--- a/Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
@@ -33,7 +33,7 @@ to customize the name of the dump item.";
     /// </summary>
     /// <param name="context">The <see cref="UITestContext"/> in which the extension is executed on.</param>
     /// <param name="pixelErrorPercentageThreshold">Maximum acceptable pixel error in percentage.</param>
-    /// <param name="regionOfInterest">Region of interest. Can be  null.</param>
+    /// <param name="regionOfInterest">Region of interest. Can be null.</param>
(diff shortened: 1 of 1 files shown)


Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
```diff
@@ -350,11 +350,12 @@ public static class VisualVerificationUITestContextExtensions
         this UITestContext context,
         IWebElement element,
         string baselineImagePath,
-        string baselineFileName)
+        string baselineFileName,
+        bool saveToTestFolder = true)
```
dministro:
```suggestion
        bool saveToProjectFolder = true)
```

Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
```diff
@@ -400,22 +401,29 @@ public static class VisualVerificationUITestContextExtensions
         // We take a screenshot of the element area. This will be compared to a baseline image.
         using var elementImageOriginal = context.TakeElementScreenshot(element).ShouldNotBeNull();
```
dministro, line 400: We take element screenshot here, and again when calling:
https://github.com/Lombiq/UI-Testing-Toolbox/blob/b43d44ed235e65a487e2f66053f91e17e00cad9c/Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs#L422

Can't we reuse the existing one?

Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
```diff
@@ -400,22 +401,29 @@ public static class VisualVerificationUITestContextExtensions
...
+                "can happen if due to a change in the app the captured element got smaller than before, or if the " +
+                $"{cropRegionName} is mistakenly too large.";
+
             if (approvedContext != null)
             {
-                context.SaveSuggestedImage(element, approvedContext.BaselineImagePath, approvedContext.BaselineFileName);
+                message += " The suggested baseline image with a screenshot of the captured element was saved to the " +
+                    "failure dump. Compare this with the original image used by the test and if suitable, use it as " +
+                    "the baseline going forward.";
+                context.SaveSuggestedImage(
```
dministro: If `approvedContext` is `null`, the image not get added to `FailureDump`.
See how the `elementImageOriginal` named in the `FailureDump`:
https://github.com/Lombiq/UI-Testing-Toolbox/blob/b43d44ed235e65a487e2f66053f91e17e00cad9c/Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs#L476

Also you can add these images anywhere to `FailureDump` since these items are only created on the filesystem if the test fails.

Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
dministro: I would suggest something like this:
```csharp
    private static void SaveSuggestedImage(
        this UITestContext context,
        IWebElement element,
        string baselineImagePath,
        string baselineFileName)
    {
        using var suggestedImage = context.TakeElementScreenshot(element);

        suggestedImage.Save(baselineImagePath, new PngEncoder());

        context.AddImageToFailureDump(baselineFileName, suggestedImage);
    }

    private static void AddImageToFailureDump(
        this UITestContext context,
        string fileName,
        Image image) =>
        // Appending suggested baseline image to failure dump too.
        context.AppendFailureDump(
            Path.Combine(
                VisualVerificationMatchNames.DumpFolderName,
                $"{fileName}.png"),
            image.Clone(),
            messageIfExists: HintFailureDumpItemAlreadyExists);
```

Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
dministro: You can use `AddImageToFailureDump` here.

dministro (changes requested):

Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
Piedone: I didn't really follow what you first meant, but now I think we have some good refactoring.

Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
Piedone: Now it is reused.

Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
Piedone: I've done some slight adjustments to these.

Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
```diff
@@ -400,22 +399,34 @@ public static class VisualVerificationUITestContextExtensions
...
+            var cropRegionName = regionOfInterest == null ? "baseline image" : "selected region of interest";
+            var message = $"The dimensions of the captured element ({elementImageOriginal.Width.ToTechnicalString()}" +
+                $"px x {elementImageOriginal.Height.ToTechnicalString()}px) are smaller than the dimensions of the " +
+                $"{cropRegionName} ({cropWidth.ToTechnicalString()}px x {cropHeight.ToTechnicalString()}px). This " +
+                "can happen if due to a change in the app the captured element got smaller than before, or if the " +
+                $"{cropRegionName} is mistakenly too large.";
+
+            message += " The suggested baseline image with a screenshot of the captured element was saved to the " +
+                "failure dump. Compare this with the original image used by the test and if suitable, use it as the " +
+                "baseline going forward.";
```
dministro: Why `message +=` is required here?
```suggestion
            var message = $"The dimensions of the captured element ({elementImageOriginal.Width.ToTechnicalString()}" +
                $"px x {elementImageOriginal.Height.ToTechnicalString()}px) are smaller than the dimensions of the " +
                $"{cropRegionName} ({cropWidth.ToTechnicalString()}px x {cropHeight.ToTechnicalString()}px). This " +
                "can happen if due to a change in the app the captured element got smaller than before, or if the " +
                $"{cropRegionName} is mistakenly too large." +
                " The suggested baseline image with a screenshot of the captured element was saved to the " +
                "failure dump. Compare this with the original image used by the test and if suitable, use it as the " +
                "baseline going forward.";
```

dministro (changes requested):

Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
Piedone: It isn't, just a leftover from the previous version.

dministro (approved):

PR #251, opened 2023-01-10
WALMA-99: Flaky UI tests

DemeSzabolcs:
[WALMA-99](https://lombiq.atlassian.net/browse/WALMA-99)

[WALMA-99]: https://lombiq.atlassian.net/browse/WALMA-99?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Services/SqlServerManager.cs
```diff
@@ -41,9 +43,11 @@ public sealed class SqlServerManager : IAsyncDisposable
     private const string DbSnapshotName = "Database.bak";
 
     private static readonly PortLeaseManager _portLeaseManager;
+    private static readonly SemaphoreSlim _semaphore;
```
dministro:
```suggestion
    private static readonly SemaphoreSlim _semaphore = new(1, 1);
```

Lombiq.Tests.UI/Services/SqlServerManager.cs
```diff
@@ -59,6 +63,7 @@ static SqlServerManager()
     {
         var agentIndexTimesHundred = TestConfigurationManager.GetAgentIndexOrDefault() * 100;
         _portLeaseManager = new PortLeaseManager(13000 + agentIndexTimesHundred, 13099 + agentIndexTimesHundred);
+        _semaphore = new(1, 1);
```
dministro:
```suggestion
```

Lombiq.Tests.UI/Services/SqlServerManager.cs
```diff
@@ -114,131 +118,205 @@ public async Task<SqlServerRunningContext> CreateDatabaseAsync()
...
+                do
+                {
+                    // Copy back snapshot.
+                    await _docker.ExecuteAsync(
+                        CancellationToken.None,
+                        "cp",
+                        Path.Combine(local),
+                        $"{containerName}:{remote}");
+
+                    result = await DockerExecuteAndGetOutputAsync(
```
dministro: In the case of `InvalidOperationException` thrown by `DockerExecuteAndGetOutputAsync` the retry logic does not apply.

Can we create a new method for this check? With a new method we can avoid the `S134` warning.
E.g.: `EnsureDockerFileExists`

Lombiq.Tests.UI/Services/SqlServerManager.cs
```diff
@@ -114,131 +118,205 @@ public async Task<SqlServerRunningContext> CreateDatabaseAsync()
...
+                        containerName,
+                        "ls",
+                        remote[..remote.LastIndexOf('/')]);
```
dministro:
```suggestion
                        Path.GetFileName(remote);
```

Lombiq.Tests.UI/Services/SqlServerManager.cs
```diff
@@ -114,131 +118,205 @@ public async Task<SqlServerRunningContext> CreateDatabaseAsync()
...
+                }
+                while (result.IsNullOrEmpty() && retryCount < maxRetries + 1);
 
-        var restore = new Restore();
-        restore.Devices.AddDevice(GetSnapshotFilePath(snapshotDirectoryPathRemote), DeviceType.File);
-        restore.Database = _databaseName;
-        restore.ReplaceDatabase = true;
+                if (result.IsNullOrEmpty())
+                {
+                    throw new FileNotFoundException(
```
dministro: Please create a custom exception for this. `FileNotFoundException` doesn't fit the real meaning of the context.
E.g.: `DockerFileCopyExcption` or similar.

dministro (changes requested):

Lombiq.Tests.UI/Services/SqlServerManager.cs
DemeSzabolcs: I think this is wrong. It's on the docker so we can't use `Path.GetFileName(remote)`. UI test also fails with this.

dministro: Yes, you are right. The directory is needed here and not the file name. Sorry.
Please use `Path.GetDirectoryName(remote)` instead.

dministro (approved):

PR #257, opened 2023-01-22
GOV-3: Add custom page header check to UI Testing Toolbox

DemeSzabolcs:
[GOV-3](https://lombiq.atlassian.net/browse/GOV-3)

[GOV-3]: https://lombiq.atlassian.net/browse/GOV-3?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
@@ -24,19 +24,19 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// <para>The test method assumes that the site is not set up.</para>
     /// </summary>
     /// <param name="setupRecipeId">The ID of the recipe to be used to set up the site.</param>
-    /// <param name="customPageHeaderXpath">The XPath of a custom page header as a string, that ultimately gets passed
+    /// <param name="customPageHeaderBy">The Selenium By to locate the custom page header, that ultimately gets passed
     /// to TestContentOperationsAsync().</param>
     /// <returns>The same <see cref="UITestContext"/> instance.</returns>
     public static Task TestBasicOrchardFeaturesAsync(
         this UITestContext context,
         string setupRecipeId,
-        string customPageHeaderXpath = null) =>
+        By customPageHeaderBy = null) =>
         context.TestBasicOrchardFeaturesAsync(
             new OrchardCoreSetupParameters(context)
             {
                 RecipeId = setupRecipeId,
             },
-            customPageHeaderXpath);
+            customPageHeaderBy);
 
     /// <summary>
     /// <para>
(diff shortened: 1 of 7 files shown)


Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -24,12 +24,19 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
     /// <para>The test method assumes that the site is not set up.</para>
     /// </summary>
     /// <param name="setupRecipeId">The ID of the recipe to be used to set up the site.</param>
+    /// <param name="customPageHeaderBy">The Selenium By to locate the custom page header, that ultimately gets passed
+    /// to TestContentOperationsAsync().</param>
```
Piedone: Similarly change the others too.

```suggestion
    /// <param name="customPageHeaderBy">
    /// The Selenium By to locate the custom page header, that contains the title of the current page/content item. This
    /// ultimately gets passed to TestContentOperationsAsync().
    /// </param>
```

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -459,11 +485,21 @@ public static Task TestRegistrationAsync(this UITestContext context, UserRegistr
...
+                context.Scope.AtataContext.Go.ToNextWindow(page);
+                page.PageTitle.Should.Contain(pageTitle);
+
+                if (customPageHeaderBy == null)
+                {
+                    page.Find<H1<OrdinaryPage>>().Should.Equal(pageTitle);
+                }
+                else
+                {
+                    context.Exists(customPageHeaderBy.Visible());
```
Piedone: Shouldn't we also check its content? To make it easier, we can just check if any of the inner text of the element contains the title, no stricter equality check necessary. Or: Perhaps better altogether to be able to provide a custom header checking logic altogether instead of just a `By`?

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
DemeSzabolcs: There are some themes where the site name is not displayed and it's only in the `page.PageTitle` which we already check. So the first option is not a good solution.

The second one would be good I think. I will check it.

Lombiq.Tests.UI/Extensions/BasicOrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -24,12 +24,21 @@ public static class BasicOrchardFeaturesTestingUITestContextExtensions
...
     /// <returns>The same <see cref="UITestContext"/> instance.</returns>
-    public static Task TestBasicOrchardFeaturesAsync(this UITestContext context, string setupRecipeId) =>
-        context.TestBasicOrchardFeaturesAsync(new OrchardCoreSetupParameters(context)
-        {
-            RecipeId = setupRecipeId,
-        });
+    public static Task TestBasicOrchardFeaturesAsync(
+        this UITestContext context,
+        string setupRecipeId,
+        Func<UITestContext, Task> customPageHeaderCheckAsync = null) =>
```
DemeSzabolcs, line 35: I used `Func<UITestContext, Task>` instead of `Action<UITestContext>`. This way the check can be async too and if not, we can put `return Task.CompletedTask;` at the end. I don't think it would work the other way around (using await in a non async function without any warning).

Piedone: Yeah, that's good.

Piedone (approved):

PR #263, opened 2023-03-03
OSOE-464: Replacing IRecipeHarvester implementations

dministro:
[OSOE-464](https://lombiq.atlassian.net/browse/OSOE-464)

[OSOE-464]: https://lombiq.atlassian.net/browse/OSOE-464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/Services/OrchardCoreHosting/FakeRecipeHarvester.cs
+++ b/Lombiq.Tests.UI/Services/OrchardCoreHosting/FakeRecipeHarvester.cs
@@ -1,0 +1,56 @@
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using OrchardCore.Environment.Extensions;
+using OrchardCore.Modules;
+using OrchardCore.Recipes.Models;
+using OrchardCore.Recipes.Services;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Lombiq.Tests.UI.Services.OrchardCoreHosting;
+
+public class FakeRecipeHarvester : IRecipeHarvester
+{
+    private readonly IRecipeReader _recipeReader;
+    private readonly IExtensionManager _extensionManager;
+    private readonly IHostEnvironment _hostingEnvironment;
+    private readonly ILogger _logger;
+
+    public FakeRecipeHarvester(
+        IRecipeReader recipeReader,
+        IExtensionManager extensionManager,
+        IHostEnvironment hostingEnvironment,
+        ILogger<FakeRecipeHarvester> logger)
+    {
+        _recipeReader = recipeReader;
+        _extensionManager = extensionManager;
+        _hostingEnvironment = hostingEnvironment;
+        _logger = logger;
+    }
+
+    public virtual Task<IEnumerable<RecipeDescriptor>> HarvestRecipesAsync() =>
+        _extensionManager.GetExtensions().InvokeAsync(HarvestRecipes, _logger);
+
+    private Task<IEnumerable<RecipeDescriptor>> HarvestRecipes(IExtensionInfo extension)
+    {
+        var folderSubPath = PathExtensions.Combine(extension.SubPath, "Recipes");
+        return HarvestRecipesAsync(folderSubPath);
+    }
+
+    protected async Task<IEnumerable<RecipeDescriptor>> HarvestRecipesAsync(string path)
+    {
+        var recipeFiles = _hostingEnvironment.ContentRootFileProvider.GetDirectoryContents(path)
+            .Where(info => !info.IsDirectory && info.Name.EndsWith(".recipe.json", StringComparison.Ordinal))
+            .ToAsyncEnumerable();
+
+        return await recipeFiles.SelectAwait(
+            async recipeFile => await _recipeReader.GetRecipeDescriptor(
+                path,
+                recipeFile,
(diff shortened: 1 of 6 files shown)


Lombiq.Tests.UI/Services/OrchardCoreHosting/OrchardApplicationFactory.cs
```diff
@@ -140,6 +144,16 @@ private void AddFakeStore(IServiceCollection services)
     private static void AddFakeViewCompilerProvider(IServiceCollection services) =>
         services.AddSingleton<IViewCompilerProvider, FakeViewCompilerProvider>();
 
+    // We remove the existing IRecipeHarvester implementations and add a custom implementation that uses the same come
```
0liver:
```suggestion
    // We remove the existing IRecipeHarvester implementations and add a custom implementation that uses the same code
```

dministro: Thank you @0liver

Lombiq.Tests.UI/Services/OrchardCoreHosting/FakeRecipeHarvester.cs
```diff
@@ -0,0 +1,56 @@
...
+using OrchardCore.Recipes.Services;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Lombiq.Tests.UI.Services.OrchardCoreHosting;
+
+public class FakeRecipeHarvester : IRecipeHarvester
```
Piedone: Please don’t forget to create an OC issue about this, since this needs to actually be fixed there, and comment the issue link here too. Or do we need this solely for UI testing and nothing to change in OC? Also comment if there's anything special to take care of here during OC upgrades.

dministro: There is an open issue for this in OC: https://github.com/OrchardCMS/OrchardCore/issues/10329

Piedone: OK then!

Lombiq.Tests.UI/Services/OrchardCoreHosting/OrchardApplicationFactory.cs
```diff
@@ -140,6 +144,18 @@ private void AddFakeStore(IServiceCollection services)
     private static void AddFakeViewCompilerProvider(IServiceCollection services) =>
         services.AddSingleton<IViewCompilerProvider, FakeViewCompilerProvider>();
 
+    // We remove the existing IRecipeHarvester implementations and add a custom implementation that uses the same code
+    // as OC but with a fix in RecipeHarvester.HarvestRecipesAsync to avoid sync over async issue.
+    // This can be removed if the related issue in OC gets fixed and merged.
+    // OC issue: https://github.com/OrchardCMS/OrchardCore/issues/10329.
+    private static void AddFakeRecipeHarvester(IServiceCollection services)
+    {
+        services.RemoveAll<IRecipeHarvester>();
```
sarahelsaig: Shouldn't you specifically remove `RecipeHarvester` and `ApplicationRecipeHarvester` instead of all implementations? We don't know if others may add their own recipe harvesters, but if they do this will break. (we add our own template harvester in Lombiq.VueJs so that feels similar)

Lombiq.Tests.UI/Services/OrchardCoreHosting/OrchardApplicationFactory.cs
```diff
@@ -140,6 +144,18 @@ private void AddFakeStore(IServiceCollection services)
...
+
+        services.AddScoped<IRecipeHarvester, FakeApplicationRecipeHarvester>()
+            .AddScoped<IRecipeHarvester, FakeRecipeHarvester>();
```
sarahelsaig: I don't think we usually chain the service add methods like this.

Lombiq.Tests.UI/Services/OrchardCoreHosting/OrchardApplicationFactory.cs
```diff
@@ -140,6 +144,18 @@ private void AddFakeStore(IServiceCollection services)
     private static void AddFakeViewCompilerProvider(IServiceCollection services) =>
         services.AddSingleton<IViewCompilerProvider, FakeViewCompilerProvider>();
 
+    // We remove the existing IRecipeHarvester implementations and add a custom implementation that uses the same code
+    // as OC but with a fix in RecipeHarvester.HarvestRecipesAsync to avoid sync over async issue.
```
sarahelsaig, line 149: > that uses the same code as OC

Then we shouldn't call it a fake harvester since it actually functions and fake should be reserved to stub/dummy services. How about AsyncRecipeHarvester?

Lombiq.Tests.UI/Services/OrchardCoreHosting/FakeRecipeHarvester.cs
sarahelsaig: Since we already have the solution here, would you open an OCORE issue in Jira for contributing it back upstream?

sarahelsaig (changes requested):

PR #282, opened 2023-06-04
OSOE-606: Fixing new C# 11 analyzer violations

Piedone:
[OSOE-606](https://lombiq.atlassian.net/browse/OSOE-606)

[OSOE-606]: https://lombiq.atlassian.net/browse/OSOE-606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/Extensions/ElementRetrievalUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/ElementRetrievalUITestContextExtensions.cs
@@ -1,11 +1,11 @@
 using Atata;
+using Lombiq.HelpfulLibraries.Common.Utilities;
 using Lombiq.Tests.UI.Services;
 using OpenQA.Selenium;
 using Shouldly;
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
-using System.Globalization;
 using System.Linq;
 
 namespace Lombiq.Tests.UI.Extensions;
@@ -127,7 +127,7 @@ public static class ElementRetrievalUITestContextExtensions
             .Where(index => index >= 0)
             .ToList();
         var target = toMatch
-            .Select(item => item == null ? null : string.Create(CultureInfo.InvariantCulture, $"{item}"))
+            .Select(item => item == null ? null : StringHelper.CreateInvariant($"{item}"))
             .Select(item => item?.Trim())
             .ToArray();
 
--- a/Lombiq.Tests.UI/Extensions/ElementRetrievalUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/ElementRetrievalUITestContextExtensions.cs
@@ -5,6 +5,7 @@ using Shouldly;
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Globalization;
 using System.Linq;
 
 namespace Lombiq.Tests.UI.Extensions;
@@ -126,7 +127,7 @@ public static class ElementRetrievalUITestContextExtensions
             .Where(index => index >= 0)
             .ToList();
         var target = toMatch
-            .Select(item => item == null ? null : FormattableString.Invariant($"{item}"))
+            .Select(item => item == null ? null : string.Create(CultureInfo.InvariantCulture, $"{item}"))
             .Select(item => item?.Trim())
             .ToArray();
 
--- a/Lombiq.Tests.UI/Helpers/ReliabilityHelper.cs
+++ b/Lombiq.Tests.UI/Helpers/ReliabilityHelper.cs
@@ -1,4 +1,5 @@
 using Atata;
+using Lombiq.HelpfulLibraries.Common.Extensions;
 using OpenQA.Selenium;
 using System;
 using System.Threading.Tasks;
@@ -66,7 +67,8 @@ public static class ReliabilityHelper
         {
             throw new TimeoutException(
                 "The process didn't succeed with retries before timing out " +
-                $"(timeout: {result.Wait.Timeout}, polling interval: {result.Wait.PollingInterval}).");
+                $"(timeout: {result.Wait.Timeout.ToTechnicalString()}, polling interval: " +
+                $"{result.Wait.PollingInterval.ToTechnicalString()}).");
         }
     }
 
@@ -99,7 +101,7 @@ public static class ReliabilityHelper
         {
             throw new TimeoutException(
                 "The process didn't succeed with retries before timing out " +
-                $"(timeout: {result.Wait.Timeout}, polling interval: {result.Wait.PollingInterval}).");
+                $"(timeout: {result.Wait.Timeout:c}, polling interval: {result.Wait.PollingInterval.ToTechnicalString()}).");
         }
     }
 
--- a/Lombiq.Tests.UI/Helpers/ReliabilityHelper.cs
+++ b/Lombiq.Tests.UI/Helpers/ReliabilityHelper.cs
@@ -1,7 +1,7 @@
 using Atata;
-using Lombiq.HelpfulLibraries.Common.Extensions;
 using Lombiq.HelpfulLibraries.Common.Utilities;
 using OpenQA.Selenium;
+using OpenQA.Selenium.Support.UI;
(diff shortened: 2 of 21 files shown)


Lombiq.Tests.UI/Extensions/ElementRetrievalUITestContextExtensions.cs
```diff
@@ -126,7 +127,7 @@ public static void VerifyElementTexts(this UITestContext context, By by, params
             .Where(index => index >= 0)
             .ToList();
         var target = toMatch
-            .Select(item => item == null ? null : FormattableString.Invariant($"{item}"))
+            .Select(item => item == null ? null : string.Create(CultureInfo.InvariantCulture, $"{item}"))
```
0liver: Use the new extension method from HL (hopefully) in this PR too.

Lombiq.Tests.UI/Helpers/ReliabilityHelper.cs
```diff
@@ -66,7 +67,8 @@ public static class ReliabilityHelper
         {
             throw new TimeoutException(
                 "The process didn't succeed with retries before timing out " +
-                $"(timeout: {result.Wait.Timeout}, polling interval: {result.Wait.PollingInterval}).");
+                $"(timeout: {result.Wait.Timeout.ToTechnicalString()}, polling interval: " +
+                $"{result.Wait.PollingInterval.ToTechnicalString()}).");
```
0liver: You can use the new approach here, too:
```suggestion
                $"(timeout: {result.Wait.Timeout.ToTechnicalString()}, polling interval: " +
                string.Create(
                    CultureInfo.InvariantCulture,
                    $"(timeout: {result.Wait.Timeout}, polling interval: {result.Wait.PollingInterval})."));
```

0liver (changes requested):

Lombiq.Tests.UI/Helpers/ReliabilityHelper.cs
```diff
@@ -99,7 +101,7 @@ public static class ReliabilityHelper
         {
             throw new TimeoutException(
                 "The process didn't succeed with retries before timing out " +
-                $"(timeout: {result.Wait.Timeout}, polling interval: {result.Wait.PollingInterval}).");
+                $"(timeout: {result.Wait.Timeout:c}, polling interval: {result.Wait.PollingInterval.ToTechnicalString()}).");
```
0liver: We now have three different ways of displaying a `TimeSpan` in one file - that seems like overkill:

image

You don't need the `c` format specifier here, because it seems to be the default anyway:

image

I suggest to use the same approach here as in lines 69-71, and remove the new extension method on `TimeSpan`.

Lombiq.Tests.UI/MonkeyTesting/UrlFilters/NotStartsWithMonkeyTestingUrlFilter.cs
```diff
@@ -17,5 +16,5 @@ public class NotStartsWithMonkeyTestingUrlFilter : IMonkeyTestingUrlFilter
         _relativeUrlNotStartsWith = relativeUrlNotStartsWith;
 
     public bool AllowUrl(UITestContext context, Uri url) =>
-       !_relativeUrlNotStartsWith.Any(filterUrl => url.PathAndQuery.StartsWithOrdinalIgnoreCase(filterUrl));
+       !Array.Exists(_relativeUrlNotStartsWith, url.PathAndQuery.StartsWithOrdinalIgnoreCase);
```
0liver: Looking at this (again), I shudder due to the inconsistent code style we'll be using:

image

I'd be heavily in favor of providing the necessary methods (or all) as extension methods on the the `Array` class and use them as instance methods instead of these unintuitive static invocations.

Lombiq.Tests.UI/Models/SafeWaitAsync.cs
0liver, line 1: R# stumbled over two more improvements in this file:

image

They seem worth to also accept.

Lombiq.Tests.UI/Services/SmtpService.cs
```diff
@@ -72,10 +75,14 @@ public async Task<SmtpServiceRunningContext> StartAsync()
...
 
         var smtp4devConfig = (manifest["tools"] as JObject)?["rnwood.smtp4dev"];
+        // We don't need to access smtp4devConfig otherwise, this is just a null check, so using a coalesce expression
+        // would make the code less readable.
+#pragma warning disable IDE0270 // Use coalesce expression
         if (smtp4devConfig == null)
         {
             throw new InvalidOperationException("There was no smtp4dev configuration in the .NET CLI local tool manifest file.");
         }
+#pragma warning restore IDE0270 // Use coalesce expression
```
0liver: Thinking about this again, it seems to me that a simple comment would make this code simpler:

```suggestion
        // Verify that an smtp4dev configuration is in place.
        if ((manifest["tools"] as JObject)?["rnwood.smtp4dev"] == null)
        {
            throw new InvalidOperationException("There was no smtp4dev configuration in the .NET CLI local tool manifest file.");
        }
```

0liver (changes requested):

Lombiq.Tests.UI/Helpers/ReliabilityHelper.cs
Piedone: Yep, `StringHelper` should be used as well. I've used `c` before adding `ToTechnicalString()` and apparently this one usage I forgot to change.

Otherwise, `ToTechnicalString()` is necessary, since not all `ToString()`s of `TimeSpan` are in formatted strings.

Lombiq.Tests.UI/MonkeyTesting/UrlFilters/NotStartsWithMonkeyTestingUrlFilter.cs
Piedone: It's not too intuitive, but [the rule docs](https://rules.sonarsource.com/csharp/RSPEC-6605) suggests it, so it's not a big leap to figure it out. I guess there's some reason why there's no such extension in `System.Linq`. Anyway, I added them.

I really don't have in me to go through the whole of `Array` to add everything we might ever need, but I added the ones we already use.

0liver: I think you've covered the most important ones. We can always add more later.

Lombiq.Tests.UI/Helpers/ReliabilityHelper.cs
0liver: Currently, `TimeSpanExtensions.ToTechnicalString()` is not used anywhere in OSOCE, so I feel an urge to have it removed. `{timeSpan:c}` will already produce a culture-invariant string representation, and judging from the lack of analyzer warnings on the `timeSpan.ToString("c");` method body, that shouldn't pose a problem.

In fact, `timeSpan.ToString();` will produce the same result. Docs: https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-timespan-format-strings#the-constant-c-format-specifier.

Lombiq.Tests.UI/Models/UITestManifest.cs
```diff
@@ -38,6 +37,5 @@ private static T GetValueOfType<T>(object instance)
         instance
             .GetType()
             .GetFields(BindingFlags.NonPublic | BindingFlags.Instance)
-            .FirstOrDefault(field => field.FieldType == typeof(T))
-            ?.GetValue(instance) as T;
+            .Find(field => field.FieldType == typeof(T))?.GetValue(instance) as T;
```
0liver: I missed this formatting change before, please fix:
```suggestion
            .Find(field => field.FieldType == typeof(T))?
            .GetValue(instance) as T;
```

0liver (changes requested):

0liver (approved):

PR #290, opened 2023-06-19
OSOE-626: UI test crashes if the web project doesn't reference Lombiq.Tests.UI.Shortcuts

### LINKED ISSUE #279: UI test crashes if the web project doesn't reference Lombiq.Tests.UI.Shortcuts (OSOE-626)
Even if it's not yet used. It throws the exception:

``InvalidOperationException: Unable to resolve service for type 'Microsoft.AspNetCore.Identity.ILookupNormalizer' while attempting to activate 'Microsoft.AspNetCore.Identity.RoleManager`1[OrchardCore.Security.IRole]'.``

This dependency should be optional, and UI tests should work without it if no shortcuts are actually used.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-626)

DemeSzabolcs:
[OSOE-626](https://lombiq.atlassian.net/browse/OSOE-626)
Fixes #279

[OSOE-626]: https://lombiq.atlassian.net/browse/OSOE-626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Extensions/OrchardCoreBuilderExtensions.cs
```diff
@@ -36,7 +36,10 @@ public static class OrchardCoreBuilderExtensions
                 .AddSingleton<IModuleStaticFileProvider>(serviceProvider =>
                     new ModuleEmbeddedStaticFileProvider(serviceProvider.GetRequiredService<IApplicationContext>())));
 
-        if (enableShortcutsDuringUITesting) builder.AddTenantFeatures("Lombiq.Tests.UI.Shortcuts", "OrchardCore.Roles");
+        if (enableShortcutsDuringUITesting) builder.AddTenantFeatures(
+            "Lombiq.Tests.UI.Shortcuts",
+            "OrchardCore.Roles",
+            "OrchardCore.Users");
```
sarahelsaig: Why is this necessary when `OrchardCore.Users` is already a dependency of `Lombiq.Tests.UI.Shortcuts`? ([see the manifest](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI.Shortcuts/Manifest.cs#L11-L25))

If we really need it, then I think it would be good to have the manifest and this call in sync to avoid any weird edge cases. Include `OrchardCore.Roles` in the manifest as a dependency of `Lombiq.Tests.UI.Shortcuts`, and include all dependencies from the manifest in this call. It would be even neater, if you could explicitly reference the `FeatureAttribute` to fetch all dependencies instead of duplicating the list here, but I'm not certain if that's possible...

sarahelsaig (changes requested):

Lombiq.Tests.UI/Extensions/OrchardCoreBuilderExtensions.cs
DemeSzabolcs: >Why is this necessary when OrchardCore.Users is already a dependency of Lombiq.Tests.UI.Shortcuts? ([see the manifest](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI.Shortcuts/Manifest.cs#L11-L25))

To avoid the ```InvalidOperationException: Unable to resolve service for type 'Microsoft.AspNetCore.Identity.ILookupNormalizer' while attempting to activate 'Microsoft.AspNetCore.Identity.RoleManager`1[OrchardCore.Security.IRole]'.``` error message.

It's in the Manifest, yes. But when `Lombiq.Tests.UI.Shortcuts` is turned on (this is the default) **BUT** not referenced in the web project's csproj, the Manifest will not be "included", so neither `OrchardCore.Users`. By activating  `OrchardCore.Users` here we can avoid the error (`InvalidOperationException`), and the only problem will be, when the UI tests are trying to use `Lombiq.Tests.UI.Shortcuts` features, like going to a specific shortcut URL (they will get a 404). In my opinion that is more telling, than the previous `Unable to resolve...` message.

> If we really need it, then I think it would be good to have the manifest and this call in sync to avoid any weird edge cases. Include OrchardCore.Roles in the manifest as a dependency of Lombiq.Tests.UI.Shortcuts, and include all dependencies from the manifest in this call. It would be even neater, if you could explicitly reference the FeatureAttribute to fetch all dependencies instead of duplicating the list here, but I'm not certain if that's possible...

Okay I will take a look.

jtkech: Just for info as I'm not fully aware of the context.

This is `OC.Users` that registers identity services including an `ILookupNormalizer` that for some reasons is registered as a tenant singleton. So if `UI.Shortcuts` depends on `OC.Users` no problem.

But yes if `UI.Shortcuts` is not referenced by the web app and you are using the following, neither `UI.Shortcuts` nor `OC.Users` will be enabled, here only `OC.Roles` will be enabled.

    .AddTenantFeatures("Lombiq.Tests.UI.Shortcuts", "OrchardCore.Roles");

First, maybe something missing on OC side, maybe `OC.Roles` should also try to register identity services, at least `ILookupNormalizer` which is injected by the aspnetcore `RoleManager`.

That said if `OC.Roles` is really needed, as @DAud-IcI said better to add it to the `UI.Shortcuts` deps, and then just use the following so that if `UI.Shortcuts` is not there nothing is registered.

    .AddTenantFeatures("Lombiq.Tests.UI.Shortcuts");

---

Also just for info, depending on the need, `AddTenantFeatures()` registers features that will be enabled but not during the setup, `AddSetupFeatures()` registers features that will be enabled during the setup but not after, `AddGlobalFeatures()` registers features that will be enabled during the setup and also after.

Hope this will help.

**Edited**: Also, if `UI.Shortcuts` is not referenced by the web app but by the test app, would need to be tried but maybe your web app factory could register an `IModuleNamesProvider` so that the related assemblies could be loaded, as we do in OC in `SiteStartup.cs`.

DemeSzabolcs: Thanks for the help, yes you are right!

sarahelsaig: Thanks for the insights @jtkech ! When you said

> if UI.Shortcuts is not referenced by the web app and you are using the following, neither UI.Shortcuts nor OC.Users will be enabled, here only OC.Roles will be enabled.

do I understand correctly that this means the `AddTenantFeatures` couldn't enable `UI.Shortcuts` so it just gets skipped? If so, does it log some kind of warning about it?

jtkech: `UI.Shortcuts` will be persisted in the `ShellDescriptor` but because it is not referenced it is not a loadable feature. So when building the shell container, in `CompositionStrategy.ComposeAsync()` (see below) it will be part of `featureNames` but not part of the `features` (`FeatureEntry` collection) returned by the extension manager and that we use to build the `ShellBlueprint`.

            ...
            var featureNames = descriptor.Features.Select(x => x.Id).ToArray();

            var features = await _extensionManager.LoadFeaturesAsync(featureNames);

            var entries = new Dictionary<Type, FeatureEntry>();

            foreach (var feature in features)
            ...

> If so, does it log some kind of warning about it?

Currently no warning is logged, maybe we should and could be a good suggestion.

PR #298, opened 2023-07-27
OSOE-591: Test doesn't fail if the visual verification baseline image is not found but the test is retried

### LINKED ISSUE #264: Test doesn't fail if the visual verification baseline image is not found but the test is retried (OSOE-591)
When visual verification testing can't find the baseline image, it'll create it, which is good. However, in conjunction with test retries, this makes those tests passing that don't have a corresponding baseline image. This should be handled somehow, to make the tests fail even on retries.

The development-time experience here should be what we have today, but when tests are executed "for real" (for example, but not necessarily during CI builds), the newly created files shouldn't be considered for retries.

An approach would be to check if the current run is in GitHub Actions, and not create baseline images then, but this ties this fix to GHA. Perhaps better would be not to create baseline images if `MaxRetryCount` > 0, and update the exception message to indicate when a baseline images was not created and why (since during development you shouldn't really use retries but otherwise you should).

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-591)

MZole:
[OSOE-591](https://lombiq.atlassian.net/browse/OSOE-591)
Fixes #264

[OSOE-591]: https://lombiq.atlassian.net/browse/OSOE-591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Exceptions/VisualVerificationBaselineImageNotFoundException.cs
```diff
@@ -7,14 +8,22 @@ namespace Lombiq.Tests.UI.Exceptions;
...
             + " (like a NuGet package) UI testing assembly. If you run the test again, this newly created verification"
             + " file will be asserted against and the assertion will pass (unless the display of the app changed in the"
-            + " meantime).",
-            innerException)
-    {
-    }
+            + " meantime)."
+        : string.Create(
+            CultureInfo.InvariantCulture,
+            $"Baseline image file was not found under the path {path} and MaxRetryCount is set to {maxRetryCount}, "
```
DemeSzabolcs:
```suggestion
            $"Baseline image file was not found under the path {path} and maxRetryCount is set to {maxRetryCount}, "
```

Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
```diff
@@ -389,8 +389,18 @@ public static class VisualVerificationUITestContextExtensions
...
-                context.SaveSuggestedImage(element, approvedContext.BaselineImagePath, approvedContext.BaselineImageFileName);
-                throw new VisualVerificationBaselineImageNotFoundException(approvedContext.BaselineImagePath);
+                if (context.Configuration.MaxRetryCount == 0)
+                {
+                    context.SaveSuggestedImage(
+                        element,
+                        approvedContext.BaselineImagePath,
+                        approvedContext.BaselineImageFileName);
+                    throw new VisualVerificationBaselineImageNotFoundException(
+                        approvedContext.BaselineImagePath, context.Configuration.MaxRetryCount);
```
DemeSzabolcs: We don't need this, we are throwing anyway outside the `if`.

Lombiq.Tests.UI/Exceptions/VisualVerificationBaselineImageNotFoundException.cs
```diff
@@ -7,14 +8,22 @@ namespace Lombiq.Tests.UI.Exceptions;
...
+        maxRetryCount == 0 ? $"Baseline image file not found, thus it was created automatically under the path {path}."
             + " Please set its \"Build action\" to \"Embedded resource\" if you want to deploy a self-contained"
             + " (like a NuGet package) UI testing assembly. If you run the test again, this newly created verification"
             + " file will be asserted against and the assertion will pass (unless the display of the app changed in the"
-            + " meantime).",
-            innerException)
-    {
-    }
+            + " meantime)."
+        : string.Create(
```
DemeSzabolcs: Why do you need `string.Create`? You can convert `int`s to `string`s with `ToTechnicalString()`.

DemeSzabolcs (changes requested):

DemeSzabolcs (approved):

PR #313, opened 2023-09-16
OSOE-683: Update all NuGet dependencies

Psichorex:
[OSOE-683](https://lombiq.atlassian.net/browse/OSOE-683)

[OSOE-683]: https://lombiq.atlassian.net/browse/OSOE-683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/Extensions/IWebApplicationInstanceExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/IWebApplicationInstanceExtensions.cs
@@ -50,7 +50,10 @@ public static class IWebApplicationInstanceExtensions
             httpContextAccessor.HttpContext = shellScope.ShellContext.CreateHttpContext();
             var httpContext = httpContextAccessor.HttpContext;
 
+            // Seems to be a false positive warning. IDE thinks that left hand side can't be null.
+#pragma warning disable S2583 // Conditionally executed code should be reachable
             httpContext.Request.PathBase = "/" + shellHost.GetSettings(tenant).RequestUrlPrefix ?? string.Empty;
+#pragma warning restore S2583 // Conditionally executed code should be reachable
             httpContext.Features.Set(new RecipeEnvironmentFeature());
 
             await shellScope.UsingAsync(execute, activateShell);
--- a/Lombiq.Tests.UI/Extensions/IWebApplicationInstanceExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/IWebApplicationInstanceExtensions.cs
@@ -49,11 +49,7 @@ public static class IWebApplicationInstanceExtensions
             // Creating a fake HttpContext like in ModularBackgroundService.
             httpContextAccessor.HttpContext = shellScope.ShellContext.CreateHttpContext();
             var httpContext = httpContextAccessor.HttpContext;
-
-            // Seems to be a false positive warning. IDE thinks that left hand side can't be null.
-#pragma warning disable S2583 // Conditionally executed code should be reachable
-            httpContext.Request.PathBase = "/" + shellHost.GetSettings(tenant).RequestUrlPrefix ?? string.Empty;
-#pragma warning restore S2583 // Conditionally executed code should be reachable
+            httpContext.Request.PathBase = "/" + shellHost.GetSettings(tenant).RequestUrlPrefix;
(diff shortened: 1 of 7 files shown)


Lombiq.Tests.UI/Extensions/IWebApplicationInstanceExtensions.cs
```diff
@@ -50,7 +50,10 @@ public static class IWebApplicationInstanceExtensions
             httpContextAccessor.HttpContext = shellScope.ShellContext.CreateHttpContext();
             var httpContext = httpContextAccessor.HttpContext;
 
+            // Seems to be a false positive warning. IDE thinks that left hand side can't be null.
+#pragma warning disable S2583 // Conditionally executed code should be reachable
             httpContext.Request.PathBase = "/" + shellHost.GetSettings(tenant).RequestUrlPrefix ?? string.Empty;
+#pragma warning restore S2583 // Conditionally executed code should be reachable
```
sarahelsaig: It's not a false positive, but a precedence issue. It turns out the `+` is evaluated before the `??` while I think the intention for this code was to do `"/" + (shellHost.GetSettings(tenant).RequestUrlPrefix ?? string.Empty)`.
image

Also it turns out that the `?? string.Empty` is pointless, because adding null to a string is already the same as adding an empty string to a string:
image

Lombiq.Tests.UI/Models/SafeWaitAsync.cs
```diff
@@ -132,7 +132,7 @@ protected virtual bool DoesConditionResultSatisfy<TResult>(TResult result)
     {
         if (typeof(TResult) == typeof(bool))
         {
-            if (result is true) return true;
+            if (result.Equals(true)) return true;
         }
```
sarahelsaig: If you use the `is` operator instead of type checking then none of this is necessary:
```csharp
        if (result is bool resultBool)
        {
            if (resultBool) return true;
        }
```

Lombiq.Tests.UI/Services/AtataFactory.cs
```diff
@@ -92,10 +89,13 @@ Task<T> FromAsync<T>(Func<BrowserConfiguration, TimeSpan, Task<T>> factory)
...
+                // This warning seems to be a false positive as there is nothing that indicates its validity.
+#pragma warning disable S2589 // Boolean expressions should not be gratuitous
                 if (!ex.Message.ContainsOrdinalIgnoreCase("Cannot start the driver service on") || currentTry >= maxTryCount)
                 {
                     throw;
                 }
+#pragma warning restore S2589 // Boolean expressions should not be gratuitous
 
                 currentTry++;
                 var retryCount = maxTryCount - currentTry;
```
sarahelsaig: Yeah it's false positive, but it can be safely worked around by moving the `currentTry++` before the check. e.g.:
```csharp
                currentTry++;
                var retryCount = maxTryCount - currentTry;

                if (!ex.Message.ContainsOrdinalIgnoreCase("Cannot start the driver service on") || retryCount <= 0)
                {
                    throw;
                }
```

sarahelsaig (changes requested):

Lombiq.Tests.UI/Services/AtataFactory.cs
Psichorex: In this case shouldn't we adjust the start value of currentTry to 0?

Psichorex: If I adjust it to 0 the warning comes back which to me a non-senser :D 
image

sarahelsaig: If you adjust it to zero then you also have to add a `- 1` to the `retryCount` to retain the same behavior as before.
And in your pic change the `>=` to `>` but I think the `retryCount <= 0` from my example above better communicates it.

Psichorex: I missed the `retry <=0` change as it was out of the window. rekt.

sarahelsaig (approved):

PR #316, opened 2023-10-07
OSOE-430: Improve stability directly after a page load

sarahelsaig:
[OSOE-430](https://lombiq.atlassian.net/browse/OSOE-430)

[OSOE-430]: https://lombiq.atlassian.net/browse/OSOE-430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/Extensions/ExtendedLoggingExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/ExtendedLoggingExtensions.cs
@@ -1,5 +1,6 @@
 using Atata;
 using Lombiq.Tests.UI.Services;
+using Microsoft.SqlServer.Management.Dmf;
 using OpenQA.Selenium;
 using System;
 using System.Threading.Tasks;
@@ -8,6 +9,11 @@ namespace Lombiq.Tests.UI.Extensions;
 
 public static class ExtendedLoggingExtensions
 {
+    /// <summary>
+    /// Used for edge cases like when a scope becomes stale
+    /// </summary>
+    private const int StabilityRetryCount = 3;
+
     public static Task ExecuteLoggedAsync(
         this UITestContext context, string operationName, IWebElement element, Func<Task> functionAsync) =>
         context.ExecuteSectionAsync(GetLogSection(operationName, element), functionAsync);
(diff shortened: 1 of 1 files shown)


Lombiq.Tests.UI/Extensions/ExtendedLoggingExtensions.cs
```diff
@@ -87,10 +92,41 @@ public static class ExtendedLoggingExtensions
...
+            {
+                var notLast = i < StabilityRetryCount - 1;
+                try
+                {
+                    action();
+                    return;
+                }
+                catch (StaleElementReferenceException) when (notLast)
+                {
+                    Task.Delay(TimeSpan.FromSeconds(1)).Wait();
```
dministro: Why not `Thread.Sleep(TimeSpan.FromSeconds(1));`?

Lombiq.Tests.UI/Extensions/ExtendedLoggingExtensions.cs
```diff
@@ -87,10 +92,41 @@ public static class ExtendedLoggingExtensions
...
+            for (int i = 0; i < StabilityRetryCount; i++)
+            {
+                var notLast = i < StabilityRetryCount - 1;
+                try
+                {
+                    return function();
+                }
+                catch (StaleElementReferenceException) when (notLast)
+                {
+                    Task.Delay(TimeSpan.FromSeconds(1)).Wait();
```
dministro: Why not `Thread.Sleep(TimeSpan.FromSeconds(1));`?

Lombiq.Tests.UI/Extensions/ExtendedLoggingExtensions.cs
```diff
@@ -87,10 +92,41 @@ public static class ExtendedLoggingExtensions
...
+                try
+                {
+                    action();
+                    return;
+                }
+                catch (StaleElementReferenceException) when (notLast)
+                {
+                    Task.Delay(TimeSpan.FromSeconds(1)).Wait();
+                }
+            }
```
dministro, line 110: Why not `throw new InvalidOperationException("Impossible to reach.");` here? Like below.

Lombiq.Tests.UI/Extensions/ExtendedLoggingExtensions.cs
```diff
@@ -104,12 +140,25 @@ public static class ExtendedLoggingExtensions
...
+                context.Scope.AtataContext.Log.EndSection();
+                return result;
+            }
+            catch (StaleElementReferenceException) when (notLast)
+            {
+                await Task.Delay(TimeSpan.FromSeconds(1));
+            }
+        }
+
+        throw new InvalidOperationException("Impossible to reach.");
```
dministro, line 163: A custom exception would be better or a constant for the message at least.

dministro (changes requested):

Lombiq.Tests.UI/Extensions/ExtendedLoggingExtensions.cs
sarahelsaig: As the text says, that's impossible to reach. The other methods need it because they expect to return a value so if I don't throw at the end then I get a compile time error even though it's really not a possible execution path.

Lombiq.Tests.UI/Extensions/ExtendedLoggingExtensions.cs
sarahelsaig: That would be an unreasonable effort for a theoretically impossible scenario.

dministro (approved):

PR #322, opened 2023-11-14
OSOE-351: Offer security checks

### LINKED ISSUE #128: Offer security checks (OSOE-351)
Perhaps integrate with some tool that can test a form for XSS vulnerability or something like that. Need to look around. 

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-351)

Piedone:
[OSOE-351](https://lombiq.atlassian.net/browse/OSOE-351)
Fixes #128

[OSOE-351]: https://lombiq.atlassian.net/browse/OSOE-351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Services/UITestContext.cs
```diff
@@ -61,6 +62,13 @@ public class UITestContext
     /// </summary>
     public AzureBlobStorageRunningContext AzureBlobStorageRunningContext { get; }
 
+    /// <summary>
+    /// Gets the service to manage  <see href="https://www.zaproxy.org/">Zed Attack Proxy (ZAP)</see> instances for
```
BenedekFarkas:
```suggestion
    /// Gets the service to manage <see href="https://www.zaproxy.org/">Zed Attack Proxy (ZAP)</see> instances for
```

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -0,0 +1,216 @@
...
+    /// </summary>
+    /// <param name="userName">The name of the user to sign in with directly.</param>
+    public SecurityScanConfiguration SignIn(string userName = DefaultUser.UserName)
+    {
+        SignInUserName = userName;
+        return this;
+    }
+
+    /// <summary>
+    /// Excludes a given URL from the scan completely.
```
BenedekFarkas:
```suggestion
    /// Excludes URLs from the scan that are matched by the supplied regex.
```

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -0,0 +1,216 @@
...
+    /// </param>
+    public SecurityScanConfiguration ExcludeUrlWithRegex(string excludedUrlRegex)
+    {
+        ExcludedUrlRegexPatterns.Add(excludedUrlRegex);
+        return this;
+    }
+
+    /// <summary>
+    /// Disable a certain active scan rule for the whole scan. If you only want to disable a rule for a given page, use
+    /// <see cref="ExcludeUrlWithRegex(string)"/> instead.
```
BenedekFarkas: Because `ExcludeUrlWithRegex(string)` disables the scan altogether for the URLs matched.
```suggestion
    /// Disable a certain active scan rule for the whole scan. If you only want to disable a rule for specific pages
    /// matched by a regex, use <see cref="DisableScanRuleForUrlWithRegex(string, int)"/> instead.
```

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -0,0 +1,216 @@
...
+    /// </param>
+    public SecurityScanConfiguration DisableActiveScanRule(int id, string name = "")
+    {
+        DisabledActiveScanRules.Add(new ScanRule(id, name));
+        return this;
+    }
+
+    /// <summary>
+    /// Disable a certain passive scan rule for the whole scan. If you only want to disable a rule for a given page, use
+    /// <see cref="ExcludeUrlWithRegex(string)"/> instead.
```
BenedekFarkas:
```suggestion
    /// Disable a certain passive scan rule for the whole scan. If you only want to disable a rule for specific pages
    /// matched by a regex, use <see cref="DisableScanRuleForUrlWithRegex(string, int)"/> instead.
```

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -0,0 +1,216 @@
...
+        SignInUserName = userName;
+        return this;
+    }
+
+    /// <summary>
+    /// Excludes a given URL from the scan completely.
+    /// </summary>
+    /// <param name="excludedUrlRegex">
+    /// The regex pattern to match URLs against to exclude them. These should be patterns that match the whole absolute
+    /// URL, so something like ".*blog.*" to match /blog, /blog/my-post, etc.
```
BenedekFarkas:
```suggestion
    /// The regex pattern to match URLs against. It will be matched against the URL path, e.g., ".*blog.*" will match
    /// /blog, /blog/my-post, etc.
```

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -0,0 +1,216 @@
...
+        return this;
+    }
+
+    /// <summary>
+    /// Disables a rule (can be any rule, including e.g. both active or passive scan rules) for just URLs matching the
+    /// given regular expression pattern.
+    /// </summary>
+    /// <param name="urlRegex">
+    /// A regular expression pattern to match URLs against. This should be a regex pattern that matches the whole
+    /// absolute URL, so something like ".*blog.*" to match /blog, /blog/my-post, etc.
```
BenedekFarkas:
```suggestion
    /// The regex pattern to match URLs against. It will be matched against the URL path, e.g., ".*blog.*" will match
    /// /blog, /blog/my-post, etc.
```

Lombiq.Tests.UI/SecurityScanning/ZapManager.cs
```diff
@@ -0,0 +1,244 @@
...
+#pragma warning restore S103 // Lines should not be too long
+
+        var cliParameters = new List<object> { "run" };
+
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+        {
+            cliParameters.Add("--network");
+            cliParameters.Add("host");
+        }
+        else
```
BenedekFarkas, line 127: Does this work with Mac too? I'd guess that it rather needs the same as Linux, so probably need to swap the control flow and check against Windows OS instead.

Lombiq.Tests.UI/SecurityScanning/YamlDocumentExtensions.cs
```diff
@@ -0,0 +1,405 @@
...
+    /// </exception>
+    public static YamlDocument SetStartUrl(this YamlDocument yamlDocument, Uri startUri)
+    {
+        // Setting includePaths in the context is not necessary because by default everything under "urls" will be
+        // scanned.
+
+        var urls = yamlDocument.GetUrls();
+        var urlsCount = urls.Count();
+
+        if (urlsCount > 1)
```
BenedekFarkas: Is this due to a ZAP limitation? If I call AddUrl first and then call SetStartUrl, that will cause an error. Some URLs need to be added manually, if they aren't accessible via links, which might be case with registration, password reset, etc.

BenedekFarkas (changes requested):

BenedekFarkas:
This is awesome, I'm super happy we're taking this step!

Piedone:
Thank you!

Lombiq.Tests.UI/SecurityScanning/ZapManager.cs
```diff
@@ -0,0 +1,244 @@
...
+        }
+
+        if (jsonReports.Count != 1)
+        {
+            throw new SecurityScanningException(
+                "No SARIF JSON report was generated for the ZAP scan. This indicates that the scan couldn't finish. " +
+                "Check the test output for details.");
+        }
+
+        return new SecurityScanResult(reportsDirectoryPath, SarifLog.Load(jsonReports[0]));
```
BenedekFarkas, line 183: Not sure if this is the right spot, but somewhere around here the used image should also be deleted (not a problem for an ephemeral CI runner, but interesting in local dev environment for example.).

BenedekFarkas (changes requested):

BenedekFarkas:
BTW worked like a charm from VS Text Explorer too!

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
Piedone: It's important that not just the path part of the URL should match, so I applied this differently. Also for your comment on `DisableScanRuleForUrlWithRegex`.

Lombiq.Tests.UI/SecurityScanning/ZapManager.cs
Piedone: It doesn't, only under Linux, [see here](https://docs.docker.com/network/drivers/host/#:~:text=The%20host%20networking%20driver%20only%20works%20on%20Linux%20hosts%2C%20and%20is%20not%20supported%20on%20Docker%20Desktop%20for%20Mac%2C%20Docker%20Desktop%20for%20Windows%2C%20or%20Docker%20EE%20for%20Windows%20Server.). I don't see any platform restrictions for `--add-host` (perhaps it could be used under Linux too, but from what I've read, `--network host` seems like the recommended approach).

Lombiq.Tests.UI/SecurityScanning/YamlDocumentExtensions.cs
Piedone: It's for us to be able to offer compatibility with ZAP AF plans that are used otherwise for manual testing too. I thought that enforcing a single URL will makes things more explicit but we don't actually need that, we can simply clear all existing URLs too.

Piedone:
Awesome!

Lombiq.Tests.UI/SecurityScanning/ZapManager.cs
Piedone: Yeah, and also the container.

BenedekFarkas: Sorry, I meant that we should delete the containers, but IMO keeping the image is OK, so you don't have to download it for every run (it's 2GB).

Piedone: Ah OK, sure.

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
BenedekFarkas: Yep, OK, I should've cross-checked with the actual code, but I see the updated comment has matching examples now, thanks!

Lombiq.Tests.UI/SecurityScanning/ZapManager.cs
BenedekFarkas: Gotcha, thanks!

Lombiq.Tests.UI/SecurityScanning/YamlDocumentExtensions.cs
```diff
@@ -0,0 +1,425 @@
...
+public static class YamlDocumentExtensions
+{
+    /// <summary>
+    /// Sets the start URL under the app where to start the scan from in the current context of the ZAP Automation
+    /// Framework plan.
+    /// </summary>
+    /// <param name="startUri">The absolute <see cref="Uri"/> to start the scan from.</param>
+    /// <exception cref="ArgumentException">
+    /// Thrown when the ZAP Automation Framework plan contains more than a single URL in the "urls" section.
+    /// </exception>
```
BenedekFarkas: It doesn't throw that exception anymore.
```suggestion
```

Lombiq.Tests.UI/SecurityScanning/YamlDocumentExtensions.cs
```diff
@@ -0,0 +1,425 @@
...
+    public static YamlDocument SetStartUrl(this YamlDocument yamlDocument, Uri startUri)
+    {
+        // Setting includePaths in the context is not necessary because by default everything under "urls" will be
+        // scanned.
+
+        var urls = yamlDocument.GetUrls();
+        urls.Children.Clear();
+        urls.Add(startUri.ToString());
```
BenedekFarkas, line 22: Adding the start URI at the first position (without clearing the rest) would be more intuitive. This way you'd have to be aware that calling SetStartUrl and AddUrl in a different order would give a different result. Sorry, I wanted to suggest this in advance in my earlier comment, but forgot.

BenedekFarkas (changes requested):

Lombiq.Tests.UI/SecurityScanning/YamlDocumentExtensions.cs
Piedone: Clearing is necessary if we want to allow the usage of AF plans that you can manually execute from ZAP too. Due to the general recommendation that you should start with ZAP from the UI, including building custom AF plans, that's a common use case.

For this, you need the URL statically added, and it'll be the one you run the app under (like with IIS Express). This can't be the same as the one used for UI testing, since that'll use a random port. So, we can't include both of them, and you can't pre-add any other URL to the plan for the same reason either.

We could require an empty URLs section instead and throw an exception, but that's less convenient.

BenedekFarkas: Oh, I see. With these abstraction layers it comes off a bit weird, but it's necessary, OK!

BenedekFarkas (approved):

PR #325, opened 2023-11-20
OSOE-675: Set stopOnFail flag to true for Lombiq.Tests.UI.Samples

### LINKED ISSUE #309: Fail-fast test execution if any test fails (OSOE-675)
When you break something that causes multiple UI tests to fail, then it can take a lot of time until test execution finishes, since all failing tests will be retried until the configured try count. Instead, we could have an option to fail the whole test execution immediately if any tests fail.

XUnit has the [`stopOnFail` config](https://xunit.net/docs/configuration-files#stopOnFail) for this. Instead of a config file, a command line option might be better, but for that [we need to upgrade `xunit.runner.visualstudio` to the upcoming v2.5.1](https://github.com/xunit/visualstudio.xunit/issues/189) in all UI test projects. 

This might need an option in [Lombiq GitHub Actions](https://github.com/Lombiq/GitHub-Actions) as well. Setting it in the config file is no issue either, since we already do that in _Set-XUnitMaxParallelThreads.ps1_.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-675)

tteguayco:
[OSOE-675](https://lombiq.atlassian.net/browse/OSOE-675)
Fixes #309

[OSOE-675]: https://lombiq.atlassian.net/browse/OSOE-675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI.Samples/xunit.runner.json
+++ b/Lombiq.Tests.UI.Samples/xunit.runner.json
@@ -2,5 +2,6 @@
   "$schema": "https://xunit.net/schema/current/xunit.runner.schema.json",
   "parallelizeAssembly": false,
   "parallelizeTestCollections": true,
+  "maxParallelThreads": 3,
   "stopOnFail": true
 }
--- a/Lombiq.Tests.UI.Samples/xunit.runner.json
+++ b/Lombiq.Tests.UI.Samples/xunit.runner.json
@@ -2,5 +2,5 @@
   "$schema": "https://xunit.net/schema/current/xunit.runner.schema.json",
   "parallelizeAssembly": false,
   "parallelizeTestCollections": true,
-  "maxParallelThreads": 3
+  "stopOnFail": true
 }
--- a/Lombiq.Tests.UI/Docs/Configuration.md
+++ b/Lombiq.Tests.UI/Docs/Configuration.md
@@ -20,6 +20,8 @@ Note that since the tests are xUnit tests you can configure general parameters o
 </ItemGroup>
 ```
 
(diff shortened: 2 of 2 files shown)


Lombiq.Tests.UI.Samples/xunit.runner.json
```diff
@@ -2,5 +2,5 @@
   "$schema": "https://xunit.net/schema/current/xunit.runner.schema.json",
   "parallelizeAssembly": false,
   "parallelizeTestCollections": true,
-  "maxParallelThreads": 3
```
DemeSzabolcs, line 5: You can leave `"maxParallelThreads": 3` since this was here before. I wanted you to remove it from the other project because we set the value there from a GH action.

Lombiq.Tests.UI/Docs/Configuration.md
```diff
@@ -20,6 +20,8 @@ Note that since the tests are xUnit tests you can configure general parameters o
 </ItemGroup>
 ```
 
+Note also that some projects may included the flag [stopOnFail](https://xunit.net/docs/configuration-files#stopOnFail) set to `true`, which makes further tests stop once a failing test is encountered.
```
DemeSzabolcs:
```suggestion
Note also that some projects' _xunit.runner.json_ files may include the flag [stopOnFail](https://xunit.net/docs/configuration-files#stopOnFail) set to `true`, which makes further tests stop once a failing test is encountered.
```

just to be more specific

DemeSzabolcs (changes requested):

DemeSzabolcs (approved):

Issue #206, opened 2022-10-17
Custom logging instead of NLog-based one (OSOE-402)


Piedone:
Currently, the UI Testing Toolbox depends on NLog. In `OrchardApplicationFactory` we configure NLog for logging, and `OrchardCoreInstance.GetLogs()` we load the log files. Instead, while retaining support for the app under test to use NLog logging for its own reasons, we could have our custom in-memory logger just for testing. That'd collect log messages about errors and warnings, and we'd use that to check for log entries to fail the test.

Decoupling the project from NLog would also prevent issues like this when Orchard updates its NLog version: https://github.com/OrchardCMS/OrchardCore/pull/11194#issuecomment-1281160575

Related: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/264#discussion_r1016836343

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-402)

rikbosch:
Could something like https://github.com/horsdal/InMemoryLogger/blob/master/src/InMemoryLogger.cs be a solution?

Piedone:
Yep, looks suitable. Through `InMemoryLogger.RecordedLogs` we could also access the log entries.

In any case, the FailureDump should still contain a full log file.

rikbosch:
What about using OpenTelemetry for capturing logs / traces and metrics of a specific failed test?

Piedone:
That could be useful for tracing tests in general, though well beyond the scope of this issue, which is about substituting the `ILogger` implementation.

See also https://github.com/OrchardCMS/OrchardCore/issues/12251.

Piedone:
Perhaps the best would be to use [`FakeLogger`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.testing.fakelogger-1). Then [`FakeLogCollector`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.testing.fakelogcollector) can be used to retrieve the log messages.

Piedone:
This would also allow us to better filter log messages. Thus, we wouldn't log `ErrorController`'s exception, to begin with, instead of filtering it out during assertion.

Piedone:
It's important that we need to log URLs too, since this greatly improved troubleshooting from a failure dump.

Piedone:
For the record:

[...]

PR #334, opened 2023-12-27
TDEAL-16: ZAP improvements

sarahelsaig:
[TDEAL-16](https://lombiq.atlassian.net/browse/TDEAL-16)

[TDEAL-16]: https://lombiq.atlassian.net/browse/TDEAL-16?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Refactor**
  - Improved configuration handling for security scanning features.
  - Streamlined logic for temporary disabling of app log assertions during security scans.
  - Enhanced YAML document manipulation with a new method for node retrieval and addition.

- **New Features**
  - Added capability to mark certain security scan rules as false positives.
  - Introduced new methods for customizing the security scanning plan.


Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
```diff
@@ -25,7 +27,14 @@ public static class SecurityScanningUITestContextExtensions
         Action<SarifLog> assertSecurityScanResult = null) =>
         context.RunAndAssertSecurityScanAsync(
             AutomationFrameworkPlanPaths.BaselinePlanPath,
-            configure,
+            configuration =>
+            {
+                // Make sure to visit at least one page that throws an exception.
```
Piedone: Why is this necessary? Add a comment about it.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
```diff
@@ -107,6 +116,11 @@ public static class SecurityScanningUITestContextExtensions
         Action<SecurityScanConfiguration> configure = null,
         Action<SarifLog> assertSecurityScanResult = null)
     {
+        // Verify that the app logs are fine right now, then disable app log assertion for the duration of this scan.
+        await context.Configuration.AssertAppLogsAsync(context.Application);
+        var assertAppLogsAsync = context.Configuration.AssertAppLogsAsync;
+        context.Configuration.AssertAppLogsAsync = _ => Task.CompletedTask;
```
Piedone: Why would we do this? We need to know if something breaks in the app during a security scan.

Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs
```diff
@@ -65,7 +65,7 @@ public SecurityScanningTests(ITestOutputHelper testOutputHelper)
                 configuration => configuration
                     ////.UseAjaxSpider() // This is quite slow so just showing you here but not running it.
                     .ExcludeUrlWithRegex(".*blog.*")
-                    .DisablePassiveScanRule(10037, "Server Leaks Information via \"X-Powered-By\" HTTP Response Header Field(s)")
+                    .DisablePassiveScanRule(10020, "The response does not include either Content-Security-Policy with 'frame-ancestors' directive.")
```
Piedone, line 79: Adjust the comment above too.

Lombiq.Tests.UI/SecurityScanning/AutomationFrameworkPlans/Baseline.yml
```diff
@@ -42,8 +42,14 @@ jobs:
     enableTags: false
     disableAllRules: false
   rules:
-  - id: 10035
-    name: "Strict-Transport-Security Header"
+  # Some stock Orchard Core shapes and pages (including the setup page) contain inline script blocks, so this directive
+  # must be permitted for OC to work at all.
+  - id: 10055
+    name: "script-src includes unsafe-inline"
+    threshold: "off"
```
Piedone: Also on the frontend? Because then we should rather fix those in OC. If indeed, then please open an OC issue (https://lombiq.atlassian.net/wiki/spaces/OCORE/overview#Creating-issues) and link it from where in the specific test suites we have this rule disabled. But we shouldn't blanket disable this for every scan.

Also for the other YMLs.

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -161,7 +161,30 @@ public SecurityScanConfiguration DisablePassiveScanRule(int id, string name = ""
...
     public SecurityScanConfiguration DisableScanRuleForUrlWithRegex(string urlRegex, int ruleId, string ruleName = "")
     {
-        DisabledRulesForUrls[urlRegex] = new ScanRule(ruleId, ruleName);
+        _disabledRulesForUrls.Add((urlRegex, ruleId, ruleName));
+        return this;
+    }
+
+    /// <summary>
+    /// Marks a rule (can be any rule, including e.g. both active or passive scan rules) for just URLs matching the
+    /// given regular expression pattern.
```
Piedone:
```suggestion
    /// Marks a rule (can be any rule, including e.g. both active or passive scan rules) as false positivefor just URLs
    /// matching the given regular expression pattern.
```

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
```diff
@@ -25,7 +27,14 @@ public static class SecurityScanningUITestContextExtensions
         Action<SarifLog> assertSecurityScanResult = null) =>
         context.RunAndAssertSecurityScanAsync(
             AutomationFrameworkPlanPaths.BaselinePlanPath,
-            configure,
+            configuration =>
+            {
+                // Make sure to visit at least one page that throws an exception.
+                var errorPageRelativeUrl = context.GetRelativeUrlOfAction<ErrorController>(error => error.Index());
+                configuration.ModifyZapPlan(plan => plan.AddUrl(context.GetAbsoluteUri(errorPageRelativeUrl)));
```
Piedone: Do this for all the other built-in scans too.

Lombiq.Tests.UI/SecurityScanning/YamlDocumentExtensions.cs
```diff
@@ -204,20 +196,23 @@ public static YamlDocument DisablePassiveScanRule(this YamlDocument yamlDocument
...
+        });
 
-        ((YamlSequenceNode)policyDefinition["rules"]).Add(newRule);
+        return yamlDocument;
+    }
 
+    /// <summary>
+    /// Adds a regular expression used to exclude paths in the current context.
+    /// </summary>
+    public static YamlDocument AddGlobalExcludePath(this YamlDocument yamlDocument, string pathRegex)
```
Piedone: `AddExcludePathsRegex()` already does this.

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -161,7 +161,30 @@ public SecurityScanConfiguration DisablePassiveScanRule(int id, string name = ""
...
+    /// </summary>
+    /// <param name="urlRegex">
+    /// The regex pattern to match URLs against. It will be matched against the whole absolute URL, e.g., ".*blog.*"
+    /// will match https://example.com/blog, https://example.com/blog/my-post, etc.
+    /// </param>
+    /// <param name="ruleId">The ID of the rule. In the scan report, this is usually displayed as "Plugin Id".</param>
+    /// <param name="justification">
+    /// A human-readable explanation of why the alert is false positive.
+    /// </param>
+    public SecurityScanConfiguration MarkScanRuleAsFalsePositiveForUrlWithRegex(string urlRegex, int ruleId, string justification)
```
Piedone: Add docs about what happens when a false positive is reported, i.e. the comment from `AddAlertFilter()`.

Piedone (changes requested):

Lombiq.Tests.UI/SecurityScanning/AutomationFrameworkPlans/Baseline.yml
```diff
@@ -42,8 +42,14 @@ jobs:
     enableTags: false
     disableAllRules: false
   rules:
-  - id: 10035
-    name: "Strict-Transport-Security Header"
```
Piedone: This needs to be disabled because during UI testing, because a local app needn't use HSTS. It's also something commonly configured outside the app, like in Cloudflare.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: As it was, if anything breaks then the app log assertion kicks you out before the ZAP report is finished. So we actually never get any useful information about the safety of error handling. Even though a lot of security vulnerabilities are related to (incorrect) error handling.
image

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: Why? I figure we always want to do the baseline scan so it's enough to verify this once.

Lombiq.Tests.UI/SecurityScanning/AutomationFrameworkPlans/Baseline.yml
sarahelsaig: I see. Added the explanation as comment.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
Piedone: I see that as a smaller issue than not having app log assertions at all.

So, if I understand correctly, currently you get an exception and test failed if ZAP causes log entries, but you won't get the ZAP report. So, you need to fix the logged issues before you can progress with fixing whatever else ZAP may discover (ZAP shouldn't cause exceptions in your app, only input validation errors at most, but those needn't be logged). I don't see how this is survivorship bias (in the end, you need to fix both the logged errors and the ZAP alerts, it's just that won't see both at the same time). With this change, you won't see the error details from the logs at all (keep in mind that some errors are logged but aren't surfaced on the UI).

It would be nice if you'd still get the ZAP report, but then that needs to be fixed, not disabling log assertions for everything that ZAP does.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
Piedone: The Baseline Full scans are mutually exclusive since Full also runs what the Baseline does. The two API scans are unique but you wouldn't necessarily run Baseline or Full for the same app if it's just a bunch of APIs.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: I still think it's a big deal, but I don't have the energy to debate this (I'm sick). Anyway, [we can have an app log assertion at the end](https://github.com/Lombiq/UI-Testing-Toolbox/blob/6fa5d43849c1b8bef9489437b611ad55e352ab87/Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs#L138). This way the logs are still verified if the scan passes. If the scan doesn't pass, you are supposed to look through the log file in the failure dump anyway.

Lombiq.Tests.UI/SecurityScanning/AutomationFrameworkPlans/Baseline.yml
sarahelsaig: On the setup screen, for sure. Now that I think about it, is there any way to identify that OC is in setup? because if so, we could simply not do inject the content security policy header until that has concluded.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: Doesn't matter now, since I've removed it form here as well.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
Piedone: I don't want to debate for the debate's sake, but I honestly don't understand what benefits your original changes bring.

Now what's there is that the security scan failing has a higher priority than the log assertion failing, right? So, if the scan fails and there are log entries, then the test will fail with a `SecurityScanningAssertionException` instead of a Shouldy exception (coming from the log assertion). Once you fix the issues found the scan, only then will the test fail on the log assertion, if any log entries remain.

I see the appeal of having the scan first, but I'm afraid that this can lead to people configuring the spider to not visit certain pages, causing no errors to be logged.

Get well!

Lombiq.Tests.UI/SecurityScanning/AutomationFrameworkPlans/Baseline.yml
Piedone: I guess you can check for `SetupController` in the `HttpContext`'s routes, as well as `ShellSettings.TenantState` should be something else than `Running`. And yes if this breaks the setup screen than rather the header shouldn't be injected.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
Piedone: Why though? It's useful to check an error page, that's a good addition, just it's needed for all scans.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: > Now what's there is that the security scan failing has a higher priority than the log assertion failing, right?

Correct.

> I'm afraid that this can lead to people configuring the spider to not visit certain pages, causing no errors to be logged.

At least for my first use the opposite was true. I made this change initially, to avoid having to blacklist error pages when I just wanted to see the zap report already.

> Get well!

Thanks!

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: Always including the error page means that you _have to_ disable app log assertion (which you didn't want),  because the error page will guaranteed result in an error log. Also we enforce "OrchardCore.Diagnostics" so it's kinda redundant.

Piedone: You don't have to disable error log assertions, but you need to change its rule to exclude such expected test errors. But OK.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
Piedone: What kind of error pages? How did the ZAP report help with fixing the errors?

sarahelsaig: I mean when ZAP does weird stuff (e.g. fuzzing) that causes errors when normal human behavior doesn't. It makes things harder if I can't see what ZAP was doing (e.g. the request it sent is in the report). This has already happened during this issue, so I speak from real life experience when I say that I found the current exception priority better.

Lombiq.Tests.UI/SecurityScanning/AutomationFrameworkPlans/Baseline.yml
sarahelsaig: I managed to do that using `ShellSettings`. Unfortunately that only let me get rid of the `script-src unsafe-eval` rule.

We still need `script-src unsafe-inline`, because both Lombiq's and OC's codebase is absolutely littered with the stuff. After all, `script-src unsafe-inline` is needed for any script block, so if you do a global search for `<script at="Foot"`, there are 185 results in OC and 43 in OSOCE. All of those will have to be refactored into JS files before this rule can be re-enabled. We had some discussion about this in the past on Teams, I'm not convinced this is the way to go.

An alternative is to add a [cryptographic nonce](https://en.wikipedia.org/wiki/Cryptographic_nonce) (not to be confused with a [British nonce](https://en.wikipedia.org/wiki/Glossary_of_British_terms_not_widely_used_in_the_United_States#N:~:text=to%20his%20team.%22-,nonce,-a%20slang%20term), seriously wtf) to each script block that's delivered to that page. It could be done in Razor by extending the `ScriptTagHelper` to handle the `nonce` attribute. There is already an open issue in OC (https://github.com/OrchardCMS/OrchardCore/issues/13389) so I added a comment explaining my idea regarding the tag helper. For the time being we can't do anything else.

Also at least for TDEAL we still need `style-src unsafe-eval` because of [Slick](https://kenwheeler.github.io/slick/). It uses jQuery to add `style` attributes which violates the directive. This is a known problem, here is the open issue: https://github.com/kenwheeler/slick/issues/3799 We may replace Slick with a different carousel library in the future, but let's not scope creep in this issue.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: It's only expected to happen if manually add the URL there.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
Piedone: I see the DX problem, but I don't see it in the code. Looking into it, in the current `dev` code, an assertion exception will be thrown [here](https://github.com/Lombiq/UI-Testing-Toolbox/blob/2aca656f7ae9cd2478a8a93f337619c07b7b6d70/Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs#L118) if the scan fails. This happens before the logs are asserted [here](https://github.com/Lombiq/UI-Testing-Toolbox/blob/2aca656f7ae9cd2478a8a93f337619c07b7b6d70/Lombiq.Tests.UI/Services/UITestExecutionSession.cs#L116). So if both the scan and the log assertion would fail, then even in the `dev` code, the test will fail with a `SecurityScanningAssertionException`, and both the ZAP report and the logs will be in the failure dump. I just tested this.

What can happen though is that the ZAP scan actually doesn't fail, but still, it causes exceptions in the background, which causes the log assertion to fail. And in that case, it'd be useful to have the ZAP report.

There's actually [a config to still generate the ZAP report](https://github.com/Lombiq/UI-Testing-Toolbox/blob/2aca656f7ae9cd2478a8a93f337619c07b7b6d70/Lombiq.Tests.UI/SecurityScanning/SecurityScanningConfiguration.cs#L20), but I didn't think of such a use case, and for the original intended one (to generate a report even if the test passes) it's broken. But for this it's useful, so you can make it public and enable it during the test in question (or during debugging).

Lombiq.Tests.UI/SecurityScanning/AutomationFrameworkPlans/Baseline.yml
Piedone: I see. The nonce would definitely be the desirable solution, I added that issue to our tracked issues. Then use the OC-compatible header with a comment linking the OC issue, please, and link in the YMLs too.

We don't speak Bri'ish here.

image

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: I get `PageChangeAssertionException` while the security scan is running. It's not caused by the assertion at the end of the test that you linked. Also that's not even relevant here, after all I restore the assertion action [here](https://github.com/Lombiq/UI-Testing-Toolbox/pull/334/files#diff-b980812aff2c19ef91f60a5773448be935f40b4dd8fb2c7e785f2ba7440ee07cR137) before the code reaches [that part](https://github.com/Lombiq/UI-Testing-Toolbox/blob/2aca656f7ae9cd2478a8a93f337619c07b7b6d70/Lombiq.Tests.UI/Services/UITestExecutionSession.cs#L116).

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
Piedone: Let's take a step back: The point of using the Diagnostics module is to have a global error handler that prevents showing detailed error pages (when not in the Development environment, since [it allows them then](https://github.com/OrchardCMS/OrchardCore/blob/29d6ba2aa1e5c2aaff9133968391f9bc110faeac/src/OrchardCore.Modules/OrchardCore.Diagnostics/DiagnosticsStartupFilter.cs#L25)), and that allows the easy theming of "friendly" error pages. The goal of adding `ErrorController` here was to check if for an 5xx error you indeed don't get a detailed error page. Do I understand correctly?

Since while recommended, we can't assume everyone will use our security defaults, such a check seems useful. However, preventing logging would need a different default log assertion, a log filter (https://github.com/Lombiq/UI-Testing-Toolbox/issues/206) or an action that just returns an HTTP 5xxx (but that's not the same as an unhandled exception). And the Diagnostics module wouldn't work anyway (unless we run the app with a Test environment config or something). This is too much, so OK to not have it.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
Piedone: Where do you get the `PageChangeAssertionException` from, exactly? I don't see how that would happen in the current `dev`, since there's no page change between the scan starting and its assertion.

Piedone:
@coderabbitai review

I'm testing the AI code review tool [CodeRabbit](https://coderabbit.ai/) here. Please bear with me. And if it comments some useful feedback, please address it.

Piedone:
@coderabbitai review

Sorry about that.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: I can no longer reproduce it, so for now I have deleted this log suppression. (actually I've converted it into an extension method because I found another use for it, more below)

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: In the end I made a second call to the security scan a second time with modified values. This second scan has no recursion (so it shouldn't start wandering around) and the app log assertions are disabled just for its duration.

Lombiq.Tests.UI/SecurityScanning/ZapManager.cs
```diff
@@ -81,6 +70,11 @@ public sealed class ZapManager : IAsyncDisposable
 
         var mountedDirectoryPath = DirectoryPaths.GetTempSubDirectoryPath(context.Id, "Zap");
         var reportsDirectoryPath = Path.Combine(mountedDirectoryPath, _zapReportsDirectoryName);
+
+        // If there is already a Zap directory, delete it before recreating the necessary directory structure. This is
+        // only needed if a scan has already been executed once for the current context. Deleting the existing "Zap"
+        // directory ensures a clean slate.
+        if (Directory.Exists(mountedDirectoryPath)) Directory.Delete(mountedDirectoryPath, recursive: true);
```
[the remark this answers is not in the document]
sarahelsaig: Not necessary, This code is used in unit tests, so any exception will cause the test to fail and automatically get logged which is the expected behavior.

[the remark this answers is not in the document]
Piedone: You could use [`SafelyDeleteDirectoryIfExists`](https://github.com/Lombiq/UI-Testing-Toolbox/blob/261b5001e998b34290446823784fddde2d3b1fa7/Lombiq.Tests.UI/Helpers/DirectoryHelper.cs#L9) though.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
Piedone: Great!

Lombiq.Tests.UI/SecurityScanning/AutomationFrameworkPlans/Baseline.yml
```diff
@@ -42,9 +42,24 @@ jobs:
...
+  # This is required for <script> blocks which OC uses extensively. The rule may be removed when OC starts to provide
+  # cryptographic nonce for these script blocks (see https://github.com/OrchardCMS/OrchardCore/issues/13389).
+  - id: 10055
+    name: "script-src includes unsafe-inline"
+    threshold: "off"
+  # This is a low-risk alert, enforcing it is undesirable for branding.
+  - id: 10037
+    name: "Server Leaks Information via 'X-Powered-By' HTTP Response Header Field(s)"
+    threshold: "off"
+  # This rule generates false positives on UUIDs or random numeric sequences.
```
Piedone: Please open a ZAP issue about it.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
```diff
@@ -47,6 +51,49 @@ public static class SecurityScanningUITestContextExtensions
...
+    /// test. It also replaces the app log assertion logic with the specialized version for security scans, <see
+    /// cref="UseAssertAppLogsForSecurityScan"/>. The scan is configured to ignore the admin dashboard, optionally log
+    /// in as admin, and use the provided time limits for the "active scan" portion of the security scan.
+    /// </para></remarks>
+    public static Task RunAndConfigureAndAssertFullSecurityScanForAutomationAsync(
+        this UITestContext context,
+        Action<SecurityScanConfiguration> additionalConfiguration = null,
+        Action<SarifLog> assertSecurityScanResult = null,
+        bool doSignIn = true,
+        int maxScanDurationInMinutes = 10,
```
Piedone:
```suggestion
        int maxActiveScanDurationInMinutes = 10,
```

Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs
```diff
@@ -45,8 +45,8 @@ public SecurityScanningTests(ITestOutputHelper testOutputHelper)
     //   usually not just unnecessary for a website that's not an SPA, but also slows the scan down by a lot. However,
     //   if you have an SPA, you need to use it.
     // - Excludes certain URLs from the scan completely. Use this if you don't want ZAP to process certain URLs at all.
-    // - Disables the "Server Leaks Information via "X-Powered-By" HTTP Response Header Field(s)" alert of ZAP's passive
-    //   scan for the whole scan. This is because by default, Orchard Core sends an "X-Powered-By: OrchardCore" header.
+    // - Disables the The response does not include either Content-Security-Policy with 'frame-ancestors' directive."
+    //   alert of ZAP's passive scan for the whole scan. This is because by default, Orchard Core sends an "X-Powered-By: OrchardCore" header.
```
Piedone: The second half of the comment is outdated.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
```diff
@@ -47,6 +51,49 @@ public static class SecurityScanningUITestContextExtensions
...
+    /// <param name="maxScanDurationInMinutes">Time limit for the active scan altogether.</param>
+    /// <param name="maxRuleDurationInMinutes">Time limit for the individual rules in the active scan.</param>
+    /// <remarks><para>
+    /// This extension method makes changes to the normal configuration of the test to be more suited for CI operation.
+    /// It changes the <see cref="UITestContext.Configuration"/> to not do any retries because this is a long running
+    /// test. It also replaces the app log assertion logic with the specialized version for security scans, <see
+    /// cref="UseAssertAppLogsForSecurityScan"/>. The scan is configured to ignore the admin dashboard, optionally log
+    /// in as admin, and use the provided time limits for the "active scan" portion of the security scan.
+    /// </para></remarks>
+    public static Task RunAndConfigureAndAssertFullSecurityScanForAutomationAsync(
```
Piedone: Mention this method in the sample test.

Piedone: Everything is with automation here :). Some more descriptive name?

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -177,6 +177,10 @@ public SecurityScanConfiguration DisableScanRuleForUrlWithRegex(string urlRegex,
     /// <param name="justification">
     /// A human-readable explanation of why the alert is false positive.
     /// </param>
+    /// <remarks><para>
+    /// Marking a rule as false positive helps the development of ZAP by collecting which rules have the highest false
+    /// positive rate (see <a href="https://www.zaproxy.org/faq/how-do-i-handle-a-false-positive/">the FAQ</a>).
```
Piedone:
```suggestion
    /// positive rate (see <see href="https://www.zaproxy.org/faq/how-do-i-handle-a-false-positive/">the FAQ</see>).
```

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
```diff
@@ -134,18 +205,27 @@ public static class SecurityScanningUITestContextExtensions
...
     /// <returns>
     /// A <see cref="SecurityScanResult"/> instance containing the SARIF (<see
     /// href="https://sarifweb.azurewebsites.net/"/>) report of the scan.
     /// </returns>
     public static Task<SecurityScanResult> RunSecurityScanAsync(
         this UITestContext context,
         string automationFrameworkYamlPath,
-        Action<SecurityScanConfiguration> configure = null)
+        Action<SecurityScanConfiguration> configure = null,
+        Uri startUri = null)
```
Piedone: This is redundant with `SecurityScanConfiguration.StartUri`. How about instead check for it being `null` in this method, and only set it if it is? So then you can preconfigure a different one if you want.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
```diff
@@ -116,32 +154,47 @@ public static class SecurityScanningUITestContextExtensions
...
+                throw new SecurityScanningAssertionException(ex);
+            }
         }
 
-        // Clear app logs before app log assertion is restored.
-        context.ClearLogs();
-        context.Configuration.AssertAppLogsAsync = assertAppLogsAsync;
+        await RunAndAssertSecurityScanInnerAsync(startUri: null, configure);
+
+        // Verify that error page handling also works by visiting a known error page with no logging.
```
Piedone: `RunAndAssertSecurityScanAsync()` is meant to just run the supplied plan and assert on it, with this, it does a bit too much (since you can't opt out of the error page). So, make this optional with a config in `SecurityScanConfiguration`.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
```diff
@@ -116,32 +154,47 @@ public static class SecurityScanningUITestContextExtensions
...
+
+                // Configure this scan to only visit the target page.
+                scanConfiguration.ModifyZapPlan(plan =>
+                {
+                    plan.SetSpiderParameter("maxDepth", 1.ToTechnicalString());
+                    plan.SetSpiderParameter("maxChildren", 2.ToTechnicalString());
+                    plan.SetActiveScanParameter("recurse", "false");
+                });
+            }));
     }
```
Piedone: Hmm, this is very useful, but I think we can do better with the implementation. Two security scans rather looks like a workaround. How about instead automatically applying a temporary log filter for just that controller, and add the URL as an additional URL via `SecurityScanConfiguration` (or if that doesn't work, add a `requester` to the plan dynamically)?

Lombiq.Tests.UI/SecurityScanning/ZapManager.cs
```diff
@@ -61,26 +62,18 @@ public sealed class ZapManager : IAsyncDisposable
         string automationFrameworkYamlPath,
         Func<YamlDocument, Task> modifyPlan = null)
     {
-        // Being able to run more than one scan in a test would complicate report generation and processing the SARIF
-        // report so rather just preventing it here (really nobody should want it anyway).
```
Piedone, line 65: I guess "really nobody should want it anyway" is just one of those famous last words :D.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningConfiguration.cs
```diff
@@ -30,7 +32,23 @@ public class SecurityScanningConfiguration
...
+                result.Level != FailureLevel.Note)
+            .Select(result => new
+            {
+                Kind = result.Kind.ToString(),
+                Level = result.Level.ToString(),
+                Details = result,
+            })
+            .ToList();
+
+        errors.ShouldBeEmpty(JsonConvert.SerializeObject(errors));
```
Piedone: Why is this convert to JSON, check that if empty trick necessary, why is the original check not enough? Comment it.

Also, let's use STJ in new code now.

Same in OCC.

Lombiq.Tests.UI/Services/OrchardCoreUITestExecutorConfiguration.cs
```diff
@@ -208,4 +208,32 @@ public void AssertBrowserLogMaybe(IList<LogEntry> browserLogs, Action<string> lo
             throw;
         }
     }
+
+    /// <summary>
+    /// Similar to <see cref="AssertAppLogsCanContainWarningsAsync"/>, but also permits certain <c>|ERROR</c> log
+    /// entries which represent correct reaction to incorrect or malicious user behavior during a security scan.
```
Piedone:
```suggestion
    /// entries which represent correct reactions to incorrect or malicious user behavior during a security scan.
```

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -161,7 +161,34 @@ public SecurityScanConfiguration DisablePassiveScanRule(int id, string name = ""
...
+    /// </param>
+    /// <param name="ruleId">The ID of the rule. In the scan report, this is usually displayed as "Plugin Id".</param>
+    /// <param name="justification">
+    /// A human-readable explanation of why the alert is false positive.
+    /// </param>
+    /// <remarks><para>
+    /// Marking a rule as false positive helps the development of ZAP by collecting which rules have the highest false
+    /// positive rate (see <a href="https://www.zaproxy.org/faq/how-do-i-handle-a-false-positive/">the FAQ</a>).
+    /// </para></remarks>
+    public SecurityScanConfiguration MarkScanRuleAsFalsePositiveForUrlWithRegex(string urlRegex, int ruleId, string justification)
```
Piedone: Add an optional `name` parameter like for `ConfigureActiveScanRule()` for readability, and use it where this method is used.

Piedone (changes requested):

Lombiq.Tests.UI/Docs/SecurityScanning.md
```diff
@@ -19,6 +19,23 @@ You can create detailed security scans of your app with [Zed Attack Proxy (ZAP)]
...
+
+This is because the Docker installation is configured to use Windows images, while the [ZAP docker image](https://hub.docker.com/r/softwaresecurityproject/zap-stable/tags) is only available for Linux. If you rely on our [Lombiq GitHub Actions](https://github.com/Lombiq/GitHub-Actions) then you can configure it like this to disable a test, in this case `SecurityScanningTests`:
+```
+  call-build-and-test-workflow-windows:
+    name: Build and Test
+    uses: Lombiq/GitHub-Actions/.github/workflows/build-and-test-orchard-core.yml@dev
+    with:
+      machine-types: '["windows-latest"]'
+      test-filter: "FullyQualifiedName!~SecurityScanningTests"
+```
```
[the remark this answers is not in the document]
Piedone: First sounded like BS but actually useful!

Lombiq.Tests.UI/Docs/SecurityScanning.md
```diff
@@ -19,6 +19,23 @@ You can create detailed security scans of your app with [Zed Attack Proxy (ZAP)]
...
+## Limitations
+
+On Windows-based GitHub runners the security tests always fail with the following error:
+
+> The `docker.exe pull softwaresecurityproject/zap-stable:2.14.0 --quiet` command failed with the output below.
+> no matching manifest for windows/amd64 10.0.20348 in the manifest list entries.
+
+This is because the Docker installation is configured to use Windows images, while the [ZAP docker image](https://hub.docker.com/r/softwaresecurityproject/zap-stable/tags) is only available for Linux. If you rely on our [Lombiq GitHub Actions](https://github.com/Lombiq/GitHub-Actions) then you can configure it like this to disable a test, in this case `SecurityScanningTests`:
+```
+  call-build-and-test-workflow-windows:
```
Piedone: Use the convention in the LGHA docs, this is an old one.

Lombiq.Tests.UI/SecurityScanning/AutomationFrameworkPlans/Baseline.yml
sarahelsaig: Done https://github.com/zaproxy/zaproxy/issues/8303

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: You mean the Lombiq.Tests.UI.Samples.Tests.SecurityScanningTests, right?

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: This parameter is used to start at a different URL [within the same run](https://github.com/Lombiq/UI-Testing-Toolbox/blob/a9a408350ad2b71733b9975ac13fdb92c4140f26/Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs#L183-L197), specifically `~/Lombiq.Tests.UI/Error/Index`. That's regardless of the configuration.

Piedone: You can also achieve that by copying the config object and changing its start URL. I don't think it's useful to have this parameter in the public API for that internal use case.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
Piedone: Yep.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningConfiguration.cs
```diff
@@ -1,7 +1,9 @@
 using Lombiq.Tests.UI.Services;
 using Microsoft.CodeAnalysis.Sarif;
+using Newtonsoft.Json;
```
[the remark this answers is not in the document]
Piedone: Ah, this was nice! I didn't know it also watches comments.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: giphy
but "temporary log filter" and "requester" aren't among them :(

I couldn't find a schema or complete documentation for this yaml configuration, I hunted the above settings together the from various sources (e.g. [here](https://gist.github.com/kmcquade/c548c84e370367234b3fdf30afcd04aa), [here](https://groups.google.com/g/zaproxy-users/c/YgIiZ7FEI78/m/ymJFFWjGEAAJ) and [here](https://www.zaproxy.org/docs/desktop/addons/automation-framework/job-ascan/)). I don't know if it's possible to turn off recursion only for a specific page. Based on those, the configs I set here are job level. Without that, if I just added the error page as an additional start URL then the scans will take twice as long because it will try to probe the same site (access all the menu links) from two largely equivalent starting points. Granted, I don't know a lot about ZAP so maybe I'm just missing something...

Lombiq.Tests.UI/SecurityScanning/SecurityScanningConfiguration.cs
sarahelsaig: If `errors` is not empty, this serializes the affected items into the custom message of the `ShouldAssertException`. This is amazing for debugging, because the exception delivers you the exact problem instead of the whole collection of junk that gets truncated in the output anyway. The overhead is minima, during normal operation, it only creates an empty list and then serializes it. That's pretty cheap and well worth the infinite improvement in terms of developer experience.

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
sarahelsaig: Already exists, just called `justification` to emphasize that a "why" is needed.

Lombiq.Tests.UI/Docs/SecurityScanning.md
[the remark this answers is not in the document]
sarahelsaig: Yeah, I don't even read the first sentence, jump to the "However" part directly.

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
Piedone: That's not really the same thing. E.g. here:

image

You see why the rules were disabled, but unless you look them up by ID, you won't see which rules were disabled. The `name` serves the latter.

Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs
Piedone, line 1: After using the security defaults, can't the assertions below adjusted? Isn't actually `BasicSecurityScanShouldPass` passing without blanking out the assertion?

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
Piedone: Sorry, to clarify: 

- Log filter: you can set the log level in the appsettings to disable logging for just one class. You can also do this from code with [`WithFilter`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.filterloggerfactoryextensions.withfilter), see [sample](https://gunnarpeipman.com/aspnet-core-log-filtering/). But: it now dawned on me that the logging is not actually happening in the controller, so I'm not sure. Please check out if there's some straightforward way to prevent that exception from being logged. If not, then we'll revisit this together with https://github.com/Lombiq/UI-Testing-Toolbox/issues/206 and the current code can stay.
- You can add the [request_o_r job](https://www.zaproxy.org/docs/desktop/addons/automation-framework/job-requestor/) to a YAML to instruct ZAP to visit a URL, any time in the plan. There's also `AddRequestor()` for this.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningConfiguration.cs
Piedone: I see, that's useful, just please comment it because it's not apparent.

sarahelsaig: I do you one better! I turned it into an extension method for future reuse in other tests.

Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs
sarahelsaig: Is this in reference to [the config](https://github.com/Lombiq/UI-Testing-Toolbox/blob/1d07bc8b4efb6592f45009d9c3cf4574dd8c1dc6/Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs#L68-L70) in `SecurityScanWithCustomConfigurationShouldPass`? Or the `ShouldBeLessThan`? (You added the comment to the document instead of a specific line so I can only guess.)

Currently `BasicSecurityScanShouldPass` passes without further adjustment. So anything in the other tests is redundant and just illustrative. It's useful in that way though, for example at first I struggled with `ExcludeUrlWithRegex` because I didn't realize you need the wildcard at the beginning and end because the regex must match the whole string.

The number in `ShouldBeLessThan` is kind of pointless right now, maybe it can be removed because it's not a practical demonstration.

EDIT: I've removed the unnecessary `ShouldBeLessThan` assertion and updated the comments to reflect the current status of the site.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningConfiguration.cs
Piedone: Ah, I wanted to ask that, great!

Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs
Piedone: Yep, but mostly this: https://github.com/Lombiq/UI-Testing-Toolbox/blob/1d07bc8b4efb6592f45009d9c3cf4574dd8c1dc6/Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs#L151 (you can't comment on unchanged lines).

For SecurityScanWithCustomConfigurationShouldPass() please adjust the assertion to be a greater/less than range (+-5 of the number you usually see currently, let's say), so it kind of verifies the scan surfacing alerts.

Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs
```diff
@@ -31,33 +31,36 @@ public SecurityScanningTests(ITestOutputHelper testOutputHelper)
...
-    //   alert here.
-    // - Also disables the "Content Security Policy (CSP) Header Not Set" rule but only for the /about page. Use this to
-    //   disable rules more specifically instead of the whole scan.
-    // - Configures sign in with a user account. This is what the scan will start with. With the Blog recipe it doesn't
-    //   matter too much, since nothing on the frontend will change, but you can use this to scan authenticated features
-    //   too. Note that since ZAP uses its own spider, not the browser accessed by the test, user sessions are not
-    //   shared, so such an explicit sign in is necessary.
-    // - The assertion on the scan results is custom. Use this if you (conditionally) want to assert on the results
-    //   differently from the global context.Configuration.SecurityScanningConfiguration.AssertSecurityScanResult. The
-    //   default there is "no scanning alert is allowed"; we expect some alerts here.
```
Piedone: Why did you remove this comment? The custom assertion is still there (as it should be, since this sample demonstrates that too).

Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs
Piedone: Please address my last comment.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
```diff
@@ -205,23 +193,17 @@ async Task RunAndAssertSecurityScanInnerAsync(Uri startUri, Action<SecurityScanC
...
+        Action<SecurityScanConfiguration> configure = null)
     {
-        var configuration = new SecurityScanConfiguration();
-
-        configuration.StartAtUri(startUri ?? context.GetCurrentUri());
+        var configuration = new SecurityScanConfiguration()
+            .StartAtUri(context.GetCurrentUri());
 
         // By default ignore /vendor/ or /vendors/ URLs. This is case-insensitive. We have no control over them, and
         // they may contain several false positives (e.g. in font-awesome)..
```
Piedone:
```suggestion
        // they may contain several false positives (e.g. in font-awesome).
```

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
Piedone: While the second ZAP scan for the error page is not there anymore, which is great, the "check for it being null in this method, and only set it if it is" part would still be useful.

Lombiq.Tests.UI/Docs/SecurityScanning.md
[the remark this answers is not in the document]
Piedone: Asked the CR team: https://discord.com/channels/1134356397673414807/1134374140397236334/1196183468430864424

Lombiq.Tests.UI/SecurityScanning/SecurityScanningConfiguration.cs
```diff
@@ -32,23 +31,22 @@ public class SecurityScanningConfiguration
     /// </summary>
     public Action<UITestContext, SarifLog> AssertSecurityScanResult { get; set; } = AssertSecurityScanHasNoAlerts;
 
+    /// <summary>
+    /// Gets a value indicating whether the security scan should not visit the <see cref="ErrorController"/> to test
+    /// for correct error handling. This is achieved by running the scan a second time without leaving that page.
+    /// </summary>
+    public bool DontScanErrorPage { get; private set; }
```
Piedone: This would be better in `SecurityScanConfiguration`. That's about configuring a single scan; this class is about configuring all scans and default configs.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningConfiguration.cs
```diff
@@ -32,23 +31,22 @@ public class SecurityScanningConfiguration
     /// </summary>
     public Action<UITestContext, SarifLog> AssertSecurityScanResult { get; set; } = AssertSecurityScanHasNoAlerts;
 
+    /// <summary>
+    /// Gets a value indicating whether the security scan should not visit the <see cref="ErrorController"/> to test
+    /// for correct error handling. This is achieved by running the scan a second time without leaving that page.
```
Piedone: It's not run a second time anymore.

Piedone (changes requested):

Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs
sarahelsaig: It's because I removed the assertion, then added it back again and forgot the comment. 🤦‍♀️

Lombiq.Tests.UI/Docs/SecurityScanning.md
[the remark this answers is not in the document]
sarahelsaig: It took ten million years to figure out that the discord link didn't work for me because I haven't joined the coderabbit-chat channel yet. I had to hunt down an invite link on coderabbit's web site first.  I don't even think this is a boomer moment on my end! I mean what UX genius thought that the error message should say "You find yourself in a strange place." instead of "You are not invited to this channel, idiot!" or something like that...

Lombiq.Tests.UI/SecurityScanning/SecurityScanningConfiguration.cs
sarahelsaig: I thought this setting should apply to all scans. But okay, moved.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: Sorry, I don't understand. Check for what being null? If you mean the start URL, we set it to the current URL by default, then it can be overridden in the `configure` action below so it's never null.

Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs
sarahelsaig: [I've changed the number to 22](https://github.com/Lombiq/UI-Testing-Toolbox/pull/334/commits/5151a15c136da7a0b1d8291bb22f6ea593a9c42e#diff-ac81919e335adbfa99c1349993c7841049ccf274ca7a120c29da74f39094368aR69) from [the 34 it was is dev](https://github.com/Lombiq/UI-Testing-Toolbox/blob/b9fb086c7b1892d65e8a76655ce5245b3dfa08ac/Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs#L71). Did you mean something else?

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -161,7 +167,44 @@ public SecurityScanConfiguration DisablePassiveScanRule(int id, string name = ""
...
+        string urlRegex,
+        int ruleId,
+        string ruleName,
+        string justification)
+    {
+        if (string.IsNullOrWhiteSpace(justification?.Trim()))
+        {
+            throw new InvalidOperationException(
+                "Please provide a detailed justification for marking this alert as a false positive, including " +
+                "context and reasoning.");
```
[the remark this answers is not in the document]
sarahelsaig: It was never valid. The Trim method is inside the `string.IsNullOrWhiteSpace`, applied to its parameter.

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -90,7 +96,7 @@ public SecurityScanConfiguration SignIn(string userName = DefaultUser.UserName)
     /// </param>
     public SecurityScanConfiguration ExcludeUrlWithRegex(string excludedUrlRegex)
     {
-        ExcludedUrlRegexPatterns.Add(excludedUrlRegex);
+        _excludedUrlRegexPatterns.Add(excludedUrlRegex);
```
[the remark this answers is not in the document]
sarahelsaig: We pass this configuration into a ZAP, which is a Java application. The regex between .NET and Java isn't 100% compatible. ZAP will return an error if the regex is invalid. Such error will fail the test and enter into the test log. So this check is redundant at best. Could be even harmful in case the expression is valid in Java's regex flavor but not in .NET's regex flavor.

Lombiq.Tests.UI/Docs/SecurityScanning.md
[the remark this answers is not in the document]
Piedone: Ten million years seems like an exaggeration to me, your argument is invalid.

Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs
Piedone: > For SecurityScanWithCustomConfigurationShouldPass() please adjust the assertion to be a **greater/less than range** (+-5 of the number you usually see currently, let's say), so it kind of verifies the scan surfacing alerts.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
Piedone: Ah indeed. I thought you can pass in a `SecurityScanConfiguration` object.

Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs
Piedone: Also, https://github.com/Lombiq/UI-Testing-Toolbox/blob/4bb9d4213dbbc3183942cb33b4b2d6f5148a2812/Lombiq.Tests.UI.Samples/Tests/SecurityScanningTests.cs#L157 and the comments above doesn't seem to be necessary. However, do point people to our security defaults in this sample test too.

Piedone (changes requested):
See my reply comments.

Lombiq.Tests.UI/Docs/SecurityScanning.md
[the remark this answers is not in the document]
sarahelsaig: Nah, it was like in Triangle except my time looping axe murderer was bad UI design. 😁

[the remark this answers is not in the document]
Piedone: That should have been exhausting :D.

Piedone (approved):

PR #351, opened 2024-03-09
LMBQ-267: Fixing that only a single security scan could run at the same time, extending configurability, adjusting full scan config

Piedone:
[LMBQ-267](https://lombiq.atlassian.net/browse/LMBQ-267)

[LMBQ-267]: https://lombiq.atlassian.net/browse/LMBQ-267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -279,6 +331,31 @@ internal async Task ApplyToPlanAsync(YamlDocument yamlDocument, UITestContext co
...
+                excludes.Add(new YamlScalarNode("MySQL"));
+                excludes.Add(new YamlScalarNode("PostgreSQL"));
+                excludes.Add(new YamlScalarNode("SQLite"));
+            }
+            else
+            {
+                // Assuming SQLite.
+                excludes.Add(new YamlScalarNode("Microsoft SQL Server"));
+                excludes.Add(new YamlScalarNode("MySQL"));
+                excludes.Add(new YamlScalarNode("PostgreSQL"));
```
DemeSzabolcs, line 355: Shouldn't we put these into a `readonly` list(s) or something like that?

Lombiq.Tests.UI/SecurityScanning/ZapManager.cs
```diff
@@ -33,12 +33,21 @@ public sealed class ZapManager : IAsyncDisposable
...
     private static bool _wasPulled;
 
+    private int _zapPort;
+
+    static ZapManager()
+    {
+        var agentIndexTimesHundred = TestConfigurationManager.GetAgentIndexOrDefault() * 100;
+        _portLeaseManager = new PortLeaseManager(15000 + agentIndexTimesHundred, 15099 + agentIndexTimesHundred);
+    }
+
```
DemeSzabolcs, line 50: Maybe an explanation in a comment would be good here.

DemeSzabolcs (changes requested):

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
Piedone: Note that `excludes` here is not a `List<T>` or anything but a `YamlSequenceNode`. That doesn't have `AddRange()`, we can only add nodes one by one. Thus if we were to put these two triplets of `YamlScalarNode` into some static, then we'd need to iterate through them and them one by one.

This would result in more and more complex code, which would most possibly also be slower (so we're talking about nanoseconds here in any case), with perhaps the advantage of fewer allocations if more than one security scan is run. I don't think it's worth it.

Lombiq.Tests.UI/SecurityScanning/ZapManager.cs
Piedone: There is, below: https://github.com/Lombiq/UI-Testing-Toolbox/pull/351/files/84714cbaa18459e2b1d9ebff49adff11492707ec#diff-35fbe794126e4b604077192b3359f108db2b15d677bb08b7b8de6f573df86af5R134-R135

DemeSzabolcs (approved):

PR #333, opened 2023-12-21
NEST-113: Updating Configuration.md with info about changing HTML validation rules for specific tests

DemeSzabolcs:
[NEST-113](https://lombiq.atlassian.net/browse/NEST-113)

## Summary by CodeRabbit

- **Documentation**
	- Updated the Configuration guide with instructions for HTML validation setup.
	- Included guidance on multi-process test execution for improved testing practices.

[NEST-113]: https://lombiq.atlassian.net/browse/NEST-113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #361, opened 2024-05-02
OSOE-844: Don't include a file called .htmlvalidate.json by default in Lombiq.UITestingToolbox

### LINKED ISSUE #359: Don't include a file called .htmlvalidate.json by default (OSOE-844)
Right now we [include](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/Lombiq.Tests.UI.csproj#L43-L46) _.htmlvalidate.json_  in the project csproj file. This means if you place your own _.htmlvalidate.json_ file in a project that references the NuGet package, it won't be enough. You also have to remove the import from the package first. This is not intuitive and as far as I can see it's not documented either. 

Instead, I suggest renaming it to something like _default.htmlvalidate.json_ and [including](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/Docs/Configuration.md#:~:text=Then%20you%20can%20change%20the%20configuration%20to%20use%20that:) a `HtmlValidationOptions` out of the box that uses _.htmlvalidate.json_ if the file exists and _default.htmlvalidate.json_ if it's missing.

---

Also I suggest including `"form-dup-name": "off"` in the default file, because we we have to turn that rule off for every project that uses boolean fields or just stock ASP.NET Core checkbox tag helpers because it expands into stuff like this:
```html
<input name="OrderPart.IsCorporation.Value" type="hidden" value="false"><input name="OrderPart.BillingAndShippingAddressesMatch.Value" type="hidden" value="false">
```

---

Also backport [this](https://github.com/OrchardCMS/OrchardCore.Commerce/pull/434/files/19f6c8e86a1ee23bf1a2f12556d2e74ff9f36aab#r1585216804) into UITT.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-844)

sarahelsaig:
[OSOE-844](https://lombiq.atlassian.net/browse/OSOE-844)
Fixes #359

[OSOE-844]: https://lombiq.atlassian.net/browse/OSOE-844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Docs/Configuration.md
```diff
@@ -59,39 +59,36 @@ Recommendations and notes for such configuration:
 
 ### HTML validation configuration
 
-If you want to change some HTML validation rules from only a few specific tests, you can create a custom _.htmlvalidate.json_ file (e.g. _TestName.htmlvalidate.json_). For example:
+If you want to change some HTML validation rules from only a few specific tests, you can create a custom _.htmlvalidate.json_ file (e.g. _TestName.htmlvalidate.json_). This should extend the [default.htmlvalidate.json](../default.htmlvalidate.json) file (which is always copied into the build directory) by setting the value of `"extends"` to a relative path pointing to it. For example:
```
Piedone: After this, also remove the "Make sure to also include the root attribute and set it to true" note below, which is not generally applicable.

```suggestion
If you want to change some HTML validation rules from only a few specific tests, you can create a custom _.htmlvalidate.json_ file (e.g. _TestName.htmlvalidate.json_). This should extend the [default.htmlvalidate.json](../default.htmlvalidate.json) file (which is always copied into the build directory) by setting the value of `"extends"` to a relative path pointing to it and declaring `"root": true`. For example:
```

Lombiq.Tests.UI/Docs/Configuration.md
```diff
@@ -59,39 +59,36 @@ Recommendations and notes for such configuration:
...
+        "./default.htmlvalidate.json"
+    ],
+
+    "rules": {
+        "element-required-attributes": "off",
+        "no-implicit-button-type": "off"
+    },
 
-  "root":  true
+    "root":  true
```
Piedone: There's an extra space after the colon. Please also fix in the files.

Lombiq.Tests.UI/Docs/Configuration.md
```diff
@@ -59,39 +59,36 @@ Recommendations and notes for such configuration:
...
 }
 ```
 
 Then you can change the configuration to use that:
```
Piedone:
```suggestion
You can also create a completely standalone config file too, without any inheritance, but we'd recommend against that. 

You can change the configuration to use the above file as follows:
```

Lombiq.Tests.UI/Docs/Configuration.md
```diff
@@ -59,39 +59,36 @@ Recommendations and notes for such configuration:
...
 
 ```cs
- changeConfiguration: configuration => configuration.HtmlValidationConfiguration.HtmlValidationOptions =
-                configuration.HtmlValidationConfiguration.HtmlValidationOptions
-                    .CloneWith(validationOptions => validationOptions.ConfigPath =
-                        Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "TestName.htmlvalidate.json")));
+changeConfiguration: configuration => configuration.HtmlValidationConfiguration.HtmlValidationOptions =
+    configuration.HtmlValidationConfiguration.HtmlValidationOptions
+        .CloneWith(validationOptions => validationOptions.ConfigPath =
```
Piedone: Why do we actually need to clone? I think simply setting the path should be enough.

Lombiq.Tests.UI.Samples/.htmlvalidate.json
```diff
@@ -14,6 +14,7 @@
     "wcag/h36": "off",
     "wcag/h37": "off",
     "wcag/h67": "off",
-    "wcag/h71": "off"
+    "wcag/h71": "off",
+    "form-dup-name": "off"
```
Piedone: Why didn't you add this into the default config file in the end?

Lombiq.Tests.UI/Services/HtmlValidationConfiguration.cs
```diff
@@ -33,6 +34,14 @@ public class HtmlValidationConfiguration
...
         // https://github.com/atata-framework/atata-htmlvalidation/issues/5
         WorkingDirectory = "HtmlValidationTemp",
+        // If a consuming project adds a ".htmlvalidate.json" config file then use it, otherwise fall back to the
+        // "default.htmlvalidate.json" which always exists because Lombiq.Tests.UI copies it into the directory during
+        // build.
+        ConfigPath =
+            Path.Combine(AppDomain.CurrentDomain.BaseDirectory, ".htmlvalidate.json") is { } rootConfiguration &&
+            File.Exists(rootConfiguration)
+                ? rootConfiguration
+                : Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "default.htmlvalidate.json"),
```
Piedone: Easier to understand what's going on.

```suggestion
        ConfigPath =
            Path.Combine(AppDomain.CurrentDomain.BaseDirectory, ".htmlvalidate.json") is { } rootConfiguration && File.Exists(rootConfiguration)
            ? rootConfiguration
            : Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "default.htmlvalidate.json"),
```

Lombiq.Tests.UI/Docs/Configuration.md
```diff
@@ -59,39 +59,36 @@ Recommendations and notes for such configuration:
...
+            Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "TestName.htmlvalidate.json")));
+```
+
+Though if the file is in the base directory like above, then it can be simplified using the `WithRelativeConfigPath(params string[] pathSegments)` method:
+
+```cs
+changeConfiguration: configuration => configuration.HtmlValidationConfiguration.WithRelativeConfigPath("TestName.htmlvalidate.json");
 ```
 
 Make sure to also include the `root` attribute and set it to `true` inside the custom _.htmlvalidate.json_ file and include it in the test project like this:
```
Piedone:
```suggestion
If you want to do this for all tests in the project, just put an _.htmlvalidate.json_ file into the project root and it will be picked up without further configuration.

Make sure to also include the `root` attribute and set it to `true` inside the custom _.htmlvalidate.json_ file and include it in the test project like this:
```

Lombiq.Tests.UI/Services/HtmlValidationConfiguration.cs
```diff
@@ -61,6 +70,24 @@ public class HtmlValidationConfiguration
     public Predicate<UITestContext> HtmlValidationAndAssertionOnPageChangeRule { get; set; } =
         EnableOnValidatablePagesHtmlValidationAndAssertionOnPageChangeRule;
 
+    /// <summary>
+    /// Updates the <see cref="HtmlValidationOptions"/>.<see
+    /// cref="Atata.HtmlValidation.HtmlValidationOptions.ConfigPath"/> with a path relative to the <see
+    /// cref="AppDomain.BaseDirectory"/> of the <see cref="AppDomain.CurrentDomain"/> (i.e. the build directory).
+    /// </summary>
```
Piedone:
```suggestion
    /// <summary>
    /// Updates the <see cref="HtmlValidationOptions"/>.<see cref="HtmlValidationOptions.ConfigPath"/> with a path
    /// relative to the <see cref="AppDomain.BaseDirectory"/> of the <see cref="AppDomain.CurrentDomain"/> (i.e. the
    /// build directory).
    /// </summary>
```

Lombiq.Tests.UI/Services/TimeoutConfiguration.cs
```diff
@@ -27,13 +35,27 @@ public class TimeoutConfiguration
     /// </summary>
     public TimeSpan PageLoadTimeout { get; set; }
 
-    public static readonly TimeoutConfiguration Default = new()
+    /// <summary>
+    /// Gets or sets how long an individual test can run to prevent hanging indefinitely. Defaults to 600s.
```
Piedone:
```suggestion
    /// Gets or sets how long an individual test can run to prevent hanging indefinitely. Defaults to 10 minutes.
```

Lombiq.Tests.UI.Samples/.htmlvalidate.json
Piedone, line 1: This should also extend the default.

Piedone (changes requested):

Lombiq.Tests.UI.Samples/.htmlvalidate.json
sarahelsaig: omg I added it to the wrong file :scream_cat:

Lombiq.Tests.UI/Docs/Configuration.md
sarahelsaig: I don't know, it was already like this in the Configuration.md so I assumed it's necessary.
@DemeSzabolcs Do you remember the reason or where it came from? Since you added the docs in #333.

Lombiq.Tests.UI/Services/TimeoutConfiguration.cs
sarahelsaig: I'd prefer to keep the unit of the "Defaults" the same as the unit of the property.

Piedone: The unit of the property is `TimeSpan`, it's no an `int`.

Lombiq.Tests.UI/Docs/Configuration.md
Piedone: FYI I implemented a non-cloning way to set the path under https://github.com/Lombiq/UI-Testing-Toolbox/pull/362 too (will remove it) and nothing bad happened. The config object is unique to the test (ultimately due to `OrchardCoreUITestExecutorConfiguration` being instantiated for each test in `OrchardCoreUITestBase`) so changing its state isn't a problem.

Lombiq.Tests.UI/Services/TimeoutConfiguration.cs
sarahelsaig: The config key is `TimeoutConfiguration:TestRunTimeoutSeconds`.

Lombiq.Tests.UI/Docs/Configuration.md
DemeSzabolcs: > I don't know, it was already like this in the Configuration.md so I assumed it's necessary. @DemeSzabolcs Do you remember the reason or where it came from? Since you added the docs in #333.

I did that based on this:
https://github.com/atata-framework/atata-htmlvalidation/issues/8#issuecomment-1864839487

But yeah, `CloneWith` might not needed.

Lombiq.Tests.UI/Services/TimeoutConfiguration.cs
Piedone: I don't see how that's relevant in the docs of this property. As a developer interacting with this property, you don't need to care about a possible backing config key.

Lombiq.Tests.UI/Docs/Configuration.md
sarahelsaig: Thanks for the context! In this case it's indeed not necessary because the example was using `HtmlValidationOptions.Default`, but we use `new HtmlValidationOptions { ... }` so it's all unique instances.

Lombiq.Tests.UI/Services/TimeoutConfiguration.cs
sarahelsaig: If you configure it in one place using seconds but document it in other place using minutes, it's just needless inconsistency. But if you insist...

Piedone (approved):

PR #364, opened 2024-05-03
OSOE-770: Add support for structured html-validate output

### LINKED ISSUE #337: Add support for structured html-validate output (OSOE-770)
`html-validate`, what we use for HTML validation under the hood with `Atata.HtmlValidation`, [supports multiple formatters](https://html-validate.org/usage/cli.html#f-formatter) for the output, including JSON.

Based on this, add support for a structured HTML validation output, i.e. a collection of .NET objects (perhaps by deserializing the JSON output). This will be then useful to have custom assertions for.

Keep the simple text output too, but this output should from then on be the recommended way. Change all built-in features to use that.

Related: https://github.com/atata-framework/atata-htmlvalidation/issues/8.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-770)

AydinE:
[OSOE-770](https://lombiq.atlassian.net/browse/OSOE-770)
Fixes #337

[OSOE-770]: https://lombiq.atlassian.net/browse/OSOE-770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.editorconfig
DemeSzabolcs, line 1: I don't think we need this editroconfig file.

Lombiq.Tests.UI/Extensions/HtmlValidationResultExtensions.cs
DemeSzabolcs, line 1: Could you also add the line and rule id too?
Current:
image


Previous:
image

Otherwise looks good, the HtmlValidationReport file is okay too!

Lombiq.Tests.UI/Services/HtmlValidationConfiguration.cs
```diff
@@ -88,7 +92,18 @@ public HtmlValidationConfiguration WithRelativeConfigPath(params string[] pathSe
     public static readonly Func<HtmlValidationResult, Task> AssertHtmlValidationOutputIsEmptyAsync =
         validationResult =>
         {
-            validationResult.Output.ShouldBeEmpty();
+            // Keep supporting cases where output format is not set to JSON.
+            if (validationResult.Output.Trim().StartsWith('[') ||
+                validationResult.Output.Trim().StartsWith('{'))
```
DemeSzabolcs, line 96: I guess there isn't a fancier way to check this, I looked into it but I haven't came up with anything either. What do you think?

DemeSzabolcs (changes requested):

Lombiq.Tests.UI/Services/HtmlValidationConfiguration.cs
AydinE: Unfortunately we don't have access to the options here so we can't directly check, I haven't been able to think of a nicer way of doing it.

Lombiq.Tests.UI/Extensions/HtmlValidationResultExtensions.cs
AydinE: I have added a method in the latest commits to build the string pretty much exactly like it is in the example now

DemeSzabolcs: I will do another review soon.

DemeSzabolcs (approved):

Issue #355, opened 2024-04-22
Update testing dependencies (OSOE-838)


Piedone:
Update all of the dependencies (NuGet and NPM as well) we use for testing or otherwise referenced from the UI Testing Toolbox projects. This is especially important because [v3 of Atata was released](https://github.com/atata-framework/atata/releases/tag/v3.0.0) (see [migration guide](https://atata.io/upgrade/to-atata-3/)).

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-838)

Piedone:
We need this first: https://github.com/atata-framework/atata/issues/834. Maybe also https://github.com/atata-framework/atata/issues/835.

Piedone:
Once done, check if this perhaps fixes https://github.com/Lombiq/UI-Testing-Toolbox/issues/356.

PR #365, opened 2024-05-08
OSOE-847: Basic Orchard features testing should include Workflows and Audit Trail

### LINKED ISSUE #363: Basic Orchard features testing should include Workflows and Audit Trail (OSOE-847)
Extend `BasicOrchardFeaturesTestingUITestContextExtensions` so it also tests Workflows and Audit Trail.

- Do this similarly to how Media features are tested: a separate extension method just for Workflows, and one for Audit Trail (like `TestMediaOperationsAsync()`). Don't put all the code into the same class though.
- No need to use page classes, just write the test also similar to `TestMediaOperationsAsync()`.
- For Workflows, test the following: Add workflow, add Event and Task, test that workflow runs. You can e.g. display a notification on login or a content item publish.
- For Audit Trail, configure it to record events for a content type, test that content publishing for that type is correctly recorded.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-847)

DemeSzabolcs:
[OSOE-847](https://lombiq.atlassian.net/browse/OSOE-847)
Fixes #363

[OSOE-847]: https://lombiq.atlassian.net/browse/OSOE-847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Extensions/OrchardFeaturesTestingUITestContextExtensions.cs
Piedone, line 1: Hmm, this class will grow too, and it's three quite unrelated things in one (single responsibility principle). Since the next UITT release will be a major one with breaking changes due to https://github.com/Lombiq/UI-Testing-Toolbox/issues/355, let's make this viable long-term: please move this and `BasicOrchardFeaturesTestingUITestContextExtensions` to a new folder/namespace `Lombiq.Tests.UI.BasicOrchardFeaturesTesting`, and split this class into three.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -385,4 +386,19 @@ public static void ClickAndAcceptPrompt(this UITestContext context, By by, strin
...
+    /// A method to perform a drag and drop action.
+    /// </summary>
+    /// <param name="sourceElementBy">The source element, that should be dragged and dropped.</param>
+    /// <param name="destinationBy">The destination element, where the source element should be dropped.</param>
+    public static void DragAndDrop(this UITestContext context, By sourceElementBy, By destinationBy) =>
+        new Actions(context.Driver).DragAndDrop(context.Get(sourceElementBy), context.Get(destinationBy))
+            .Build()
+            .Perform();
+
+    public static void DragAndDropToOffset(this UITestContext context, By sourceElementBy, int offsetX, int offsetY) =>
```
Piedone, line 406: Add documentation.

Lombiq.Tests.UI/Extensions/OrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,199 @@
...
+        async () =>
+        {
+            var auditTrailPath = "/AuditTrail";
+
+            await context.EnableFeatureDirectlyAsync("OrchardCore.AuditTrail");
+            await context.GoToAdminRelativeUrlAsync("/Settings" + auditTrailPath);
+
+            await context.GoToEditorTabAsync("Content");
+
+            await context.SetCheckboxValueAsync(By.XPath("//input[@value='Page']"), isChecked: true);
```
Piedone: `SetCheckboxValueAsync()` always has `isChecked: true` in the codebase, so add that as the default for the parameter instead.

Lombiq.Tests.UI/Extensions/OrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,199 @@
...
+                    .Publish.ClickAndGo()
+                .AlertMessages.Should.Contain(message => message.IsSuccess);
+
+            await context.GoToAdminRelativeUrlAsync(auditTrailPath);
+
+            context.Exists(ByHelper.TextContains("of the Page"));
+
+            context.Exists(ByHelper.TextContains("was published"));
+
+            context.Exists(ByHelper.TextContains("was created"));
```
Piedone: These aren't unique to Audit Trail Test Page. Also check for "Audit Trail Test Page" to be included with these.

Lombiq.Tests.UI/Extensions/OrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,199 @@
...
+
+                context.DragAndDrop(
+                    By.XPath("//div[@class = 'jtk-endpoint jtk-endpoint-anchor jtk-draggable jtk-droppable']"), // #spell-check-ignore-line
+                    By.XPath(taskXPath));
+
+                await context.ClickReliablyOnAsync(By.XPath("//div[contains(@class, 'activity-event')]"));
+
+                // Waiting is necessary here because when we click on the event, we need to wait for the popup
+                // otherwise, the startup task button won't be clicked.
+                await Task.Delay(1000);
```
Piedone: Never do this. Wait for the popup's element to appear explicitly with `Exists()`.

Lombiq.Tests.UI/Extensions/OrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,199 @@
...
+
+                await context.ClickReliablyOnAsync(By.XPath("//a[@title='Startup task']").Visible());
```
Piedone: Why `Visible()`? That's the default.

Lombiq.Tests.UI/Extensions/OrchardFeaturesTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,199 @@
...
+                contentItemsPage
+                    .CreateNewPage()
+                        .Title.Set("Workflows Test Page")
+                        .Publish.ClickAndGo();
+
+                context.ShouldBeSuccess(contentItemPublishTestSuccessMessage);
+
+                // Checking if the workflow run was logged.
+                await context.GoToAdminRelativeUrlAsync(workflowsPath);
+                await context.ClickReliablyOnAsync(By.XPath("//a[contains(@href, 'Instances')]"));
```
Piedone: This opens the instances of Content Set Example Workflow.

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/OrchardFeaturesTestingUITestContextExtensions.cs
DemeSzabolcs: The elment already exists it's just not visible, if we add the visibility check too, it still won't work, since it exists, but I guess it's not enough time, to be properly loaded in. 
But anyway, I came up with a simple solution. Saving the workflow once, then clicking on the button works, then saving again.

Piedone: Exists() repeats until a timeout.

DemeSzabolcs: I understand, but it passes right away yet the button still cannot be clicked, that's why I went with the solution I wrote.

Lombiq.Tests.UI/BasicOrchardFeaturesTesting/OrchardAuditTrailFeatureTestingUITestContextExtensions.cs
Piedone, line 1: No need for the "Orchard" prefix in any of these classes, since they're already in the "Basic**Orchard**FeaturesTesting" namespace.

Lombiq.Tests.UI/BasicOrchardFeaturesTesting/OrchardAuditTrailFeatureTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+                    .Publish.ClickAndGo()
+                .AlertMessages.Should.Contain(message => message.IsSuccess);
+
+            await context.GoToAdminRelativeUrlAsync(auditTrailPath);
+
+            context.Exists(ByHelper.TextContains("of the Page"));
+
+            context.Exists(ByHelper.TextContains("was published"));
+
+            context.Exists(ByHelper.TextContains("was created"));
```
Piedone: As mentioned previously, these aren't unique to Audit Trail Test Page. Also check for "Audit Trail Test Page" to be included with **these texts here**, not separately. Because otherwise, this doesn't test for the two events of Audit Trail Test Page, rather it checks for three texts that can be satisfied by any random event independent of Audit Trail Test Page, and then checks for any random "Audit Trail Test Page" text. Check these exactly:

image

Lombiq.Tests.UI/BasicOrchardFeaturesTesting/OrchardWorkflowsFeatureTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+                var taskXPath = "//div[contains(@class, 'activity-task')]";
+
+                context.DragAndDropToOffset(By.XPath(taskXPath), 400, 0);
+
+                context.DragAndDrop(
+                    By.XPath("//div[@class = 'jtk-endpoint jtk-endpoint-anchor jtk-draggable jtk-droppable']"), // #spell-check-ignore-line
+                    By.XPath(taskXPath));
+
+                // We need to save the workflow early, because sometimes the editor, thus the startup task button can be
+                // buggy during UI testing. This way it's always clicked.
```
Piedone: Buggy how? Elaborate in the comment.

Lombiq.Tests.UI/BasicOrchardFeaturesTesting/OrchardWorkflowsFeatureTestingUITestContextExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+                contentItemsPage
+                    .CreateNewPage()
+                        .Title.Set("Workflows Test Page")
+                        .Publish.ClickAndGo();
+
+                context.ShouldBeSuccess(contentItemPublishTestSuccessMessage);
+
+                // Checking if the workflow run was logged.
+                await context.GoToAdminRelativeUrlAsync(workflowsPath);
+                await context.ClickReliablyOnAsync(By.XPath("//a[text()='Test workflow']/following-sibling::a[contains(@href, 'Instances')]"));
```
Piedone: I didn't know about this `following-sibling`, nice!

Piedone (changes requested):

Piedone (approved):

PR #367, opened 2024-05-10
OFFI-37: Adding SignInDirectlyAndGoToAdminRelativeUrlAsync extension method

porgabi:
[OFFI-37](https://lombiq.atlassian.net/browse/OFFI-37)

[OFFI-37]: https://lombiq.atlassian.net/browse/OFFI-37?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sarahelsaig (approved):
I question why we need all these `SignInDirectlyAnd...` methods in the first place, but it if we must have `SignInDirectlyAndGoToRelativeUrlAsync` then it's good to add its admin counterpart for consistency.

PR #368, opened 2024-05-11
LMBQ-323: Adding OS based visual verification option

DemeSzabolcs:
[LMBQ-323](https://lombiq.atlassian.net/browse/LMBQ-323)

[LMBQ-323]: https://lombiq.atlassian.net/browse/LMBQ-323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
@@ -53,6 +53,7 @@ to customize the name of the dump item.";
         IEnumerable<Size> sizes,
         Func<Size, By> getSelector,
         double pixelErrorPercentageThreshold = 0,
+        bool checkOS = false,
         Action<VisualVerificationMatchApprovedConfiguration> configurator = null)
     {
         context.HideScrollbar();
@@ -69,7 +70,16 @@ to customize the name of the dump item.";
                     pixelErrorPercentageThreshold: pixelErrorPercentageThreshold,
                     configurator: configuration =>
                     {
-                        configuration.WithFileNameSuffix(StringHelper.CreateInvariant($"{size.Width}x{size.Height}"));
+                        var fileNameSuffix = StringHelper.CreateInvariant($"{size.Width}x{size.Height}");
+
+                        if (checkOS)
+                        {
+                            var oS = OperatingSystem.IsWindows() ? "Windows" : "Unix";
+
+                            fileNameSuffix = fileNameSuffix + "-" + oS;
+                        }
+
+                        configuration.WithFileNameSuffix(fileNameSuffix);
                         configurator?.Invoke(configuration);
                     });
             }
--- a/Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
@@ -76,7 +76,7 @@ to customize the name of the dump item.";
                         {
                             var oS = OperatingSystem.IsWindows() ? "Windows" : "Unix";
 
-                            fileNameSuffix = fileNameSuffix + "-" + oS;
+                            fileNameSuffix = $"{fileNameSuffix}-{oS}";
                         }
 
                         configuration.WithFileNameSuffix(fileNameSuffix);


Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
```diff
@@ -69,7 +70,16 @@ public static class VisualVerificationUITestContextExtensions
...
                     configurator: configuration =>
                     {
-                        configuration.WithFileNameSuffix(StringHelper.CreateInvariant($"{size.Width}x{size.Height}"));
+                        var fileNameSuffix = StringHelper.CreateInvariant($"{size.Width}x{size.Height}");
+
+                        if (checkOS)
+                        {
+                            var oS = OperatingSystem.IsWindows() ? "Windows" : "Unix";
+
+                            fileNameSuffix = fileNameSuffix + "-" + oS;
```
MZole:
```suggestion
                            fileNameSuffix = $"{fileNameSuffix}-{oS}";
```

MZole (changes requested):

MZole (approved):

Lombiq.Tests.UI/Extensions/VisualVerificationUITestContextExtensions.cs
```diff
@@ -69,7 +70,16 @@ public static class VisualVerificationUITestContextExtensions
                     pixelErrorPercentageThreshold: pixelErrorPercentageThreshold,
                     configurator: configuration =>
                     {
-                        configuration.WithFileNameSuffix(StringHelper.CreateInvariant($"{size.Width}x{size.Height}"));
+                        var fileNameSuffix = StringHelper.CreateInvariant($"{size.Width}x{size.Height}");
+
+                        if (checkOS)
+                        {
+                            var oS = OperatingSystem.IsWindows() ? "Windows" : "Unix";
```
Piedone, line 77: There are more than two OSes ;). Also, didn't we have something similar already?

DemeSzabolcs: I will correct the variable name. We had something similiar, but insted of using this configuration, in multiple places, I added it here with a `bool` this way it's DRY. If you look at the OSOCE PR now it's just this `bool`, instead of the configuration, also in our private project.

Piedone: Ah OK, good! It's not about the variable name, but that this only supports Windows and *nix, both of which can have many different editions, and the latter can contain Linux and Mac OS (among others) too.

DemeSzabolcs: Okay, but we don't have a use case in any project, where we would differentiate between OS verions, or Mac OS, so that would just add unnecessary complexities for now.

Piedone: It's simply incorrect. It's not that every OS that is not Windows is Unix. This feature thus have a limited use case, because if we want to do anything more fine grained, which includes multiple distros of Linux, then we'll need to change this. Suffixing with the OS name would be better, for example.

DemeSzabolcs: I can check for the whole OS name and version, but in that case, the VV test will fail locally, if the user does not have the same version as the CI. For example, somebody uses Windows 11, and tests are running on Windows 10.

Piedone: Now I checked the code, and we have `WithUsePlatformAsSuffix()` for this (see https://github.com/Lombiq/UI-Testing-Toolbox/blob/798b78a33d5d877cedde27ed78d58c08975bc60f/Lombiq.Tests.UI.Samples/Tests/BasicVisualVerificationTests.cs#L81 for an example). That should be used and this new implementation removed.

If you want to have a shortcut for that, you can add something like an `AssertVisualVerificationApprovedWithPlatformSuffix()` that uses that by default, but there's no need for a new distinguishing logic.

DemeSzabolcs: Okay, I checked the generated suffixes with this. It matches with the local ones (Win32NT).

PR #378, opened 2024-06-08
OSOE-867: Addressing analyzer warnings

Psichorex:
[OSOE-867](https://lombiq.atlassian.net/browse/OSOE-867)

[OSOE-867]: https://lombiq.atlassian.net/browse/OSOE-867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI.Shortcuts/Controllers/InteractiveModeController.cs
```diff
@@ -2,11 +2,18 @@
...
 
 namespace Lombiq.Tests.UI.Shortcuts.Controllers;
 
 [AllowAnonymous]
 [DevelopmentAndLocalhostOnly]
+[SuppressMessage(
+    "Major Code Smell",
+    "S6934:A Route attribute should be added to the controller when a route template is specified at the action level",
+    Justification = "Using attribute routing on the Controller breaks the behavior and will result in a 404." +
+    "[controller]/{action=Index} on the controller will not hit the Index action method.")]
```
Piedone: This is most possibly because the pattern is wrong. See my comments on the Training Demo PR.

Lombiq.Tests.UI/Pages/OrchardCoreAdminPage.cs
```diff
@@ -1,8 +1,13 @@
 using Atata;
 using Lombiq.Tests.UI.Components;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Lombiq.Tests.UI.Pages;
 
+[SuppressMessage(
+    "Major Code Smell",
+    "S1144:Unused private types or members should be removed",
+    Justification = "Removing private setters result in failing UI Tests with NRE.")]
```
Piedone: For all of these "Removing private setters result in failing UI Tests with NRE." suppressions, figure out why this is the case, not just the symptom. Atata uses them?

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -328,10 +328,7 @@ internal async Task ApplyToPlanAsync(YamlDocument yamlDocument, UITestContext co
         }
 
         // False positive: https://github.com/SonarSource/sonar-dotnet/issues/8510.
```
Piedone: Remove this too.

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -42,7 +42,7 @@ public class SecurityScanConfiguration
     /// for correct error handling. This is achieved by adding the error page URL to the configuration with <see
     /// cref="YamlDocumentExtensions.AddRequestor"/>.
     /// </summary>
-    public bool DontScanErrorPage { get; private set; }
+    public bool DontScanErrorPage { get; }
```
Piedone: The setter should rather be public, because otherwise this property doesn't make sense.

Piedone (changes requested):

Lombiq.Tests.UI/Pages/OrchardCoreAdminPage.cs
Psichorex: Atata.PageObject<T>() requires the private setters I am quite sure about that. More here: https://atata.io/examples/page-object-inheritance/ So the rule wants these private setters to be removed because they are not directly called in the class itself. However if you remove the setters you will face NullReference exceptions most probably due to the removed setter thus they can't aquire their values in the background.

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
Psichorex: It can't be public as far as the property itself is public as needs to be less accessbile. The best solution in my opinion here is to reconsider this whole rule in these cases. Follow up here https://github.com/Lombiq/UI-Testing-Toolbox/pull/378#discussion_r1634748833

Lombiq.Tests.UI/Pages/OrchardCoreAdminPage.cs
Piedone: OK, then please in the justification instead tell that Atata needs this and link to the docs page.

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
Piedone: I don't understand. The setter of this property _can_ be public. And it should be, because this is a piece of configuration in a configuration class. The only reference to it only reads it, so if the setter is private, then this won't do anything.

Psichorex: Oh yes. My bad. You meant `{ get; set;}` and not `{ get; public set; }`. Brainmelt on my side.

Piedone (approved):

PR #386, opened 2024-07-07
OFFI-78: Support different base path for OC app

wAsnk:
[OFFI-78](https://lombiq.atlassian.net/browse/OFFI-78)

[OFFI-78]: https://lombiq.atlassian.net/browse/OFFI-78?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/MediaOperationsTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/MediaOperationsTestingUITestContextExtensions.cs
@@ -32,7 +32,7 @@ public static class MediaOperationsTestingUITestContextExtensions
                 context.Exists(By.XPath($"//span[contains(text(), '{imageName}')]"));
 
                 await context
-                    .Get(By.CssSelector($"a[href*=\"/media/{imageName}\"]").OfAnyVisibility())
+                    .Get(By.CssSelector($"a[href=\"{context.UrlPrefix}/media/{imageName}\"]").OfAnyVisibility())
                     .ClickReliablyAsync(context);
                 // Closing the newly opened tab with the image, so the browser doesn't continue to switch the UI back
                 // and forth.
@@ -56,7 +56,7 @@ public static class MediaOperationsTestingUITestContextExtensions
                     .ClickReliablyAsync(context);
 
                 await context
-                    .Get(By.CssSelector($"a[href*=\"/media/{documentName}\"]"))
+                    .Get(By.CssSelector($"a[href=\"{context.UrlPrefix}/media/{documentName}\"]"))
                     .ClickReliablyAsync(context);
                 context.SwitchToLastWindow();
                 context.Driver.Close();
--- a/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/MediaOperationsTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/MediaOperationsTestingUITestContextExtensions.cs
@@ -32,7 +32,7 @@ public static class MediaOperationsTestingUITestContextExtensions
                 context.Exists(By.XPath($"//span[contains(text(), '{imageName}')]"));
 
                 await context
-                    .Get(By.CssSelector($"a[href=\"/media/{imageName}\"]").OfAnyVisibility())
+                    .Get(By.CssSelector($"a[href*=\"/media/{imageName}\"]").OfAnyVisibility())
                     .ClickReliablyAsync(context);
                 context.SwitchToFirstWindow();
 
@@ -52,7 +52,7 @@ public static class MediaOperationsTestingUITestContextExtensions
                     .ClickReliablyAsync(context);
 
                 await context
-                    .Get(By.CssSelector($"a[href=\"/media/{documentName}\"]"))
+                    .Get(By.CssSelector($"a[href*=\"/media/{documentName}\"]"))
                     .ClickReliablyAsync(context);
                 context.SwitchToFirstWindow();
 
--- a/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/MediaOperationsTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/MediaOperationsTestingUITestContextExtensions.cs
@@ -32,7 +32,7 @@ public static class MediaOperationsTestingUITestContextExtensions
                 context.Exists(By.XPath($"//span[contains(text(), '{imageName}')]"));
 
                 await context
-                    .Get(By.CssSelector($"a[href=\"{context.UrlPrefix}/media/{imageName}\"]").OfAnyVisibility())
+                    .Get(By.CssSelector($"a[href^=\"{context.UrlPrefix}/media/{imageName}\"]").OfAnyVisibility())
                     .ClickReliablyAsync(context);
                 // Closing the newly opened tab with the image, so the browser doesn't continue to switch the UI back
                 // and forth.
@@ -56,7 +56,7 @@ public static class MediaOperationsTestingUITestContextExtensions
                     .ClickReliablyAsync(context);
 
                 await context
-                    .Get(By.CssSelector($"a[href=\"{context.UrlPrefix}/media/{documentName}\"]"))
+                    .Get(By.CssSelector($"a[href^=\"{context.UrlPrefix}/media/{documentName}\"]"))
                     .ClickReliablyAsync(context);
                 context.SwitchToLastWindow();
                 context.Driver.Close();
--- a/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/MediaOperationsTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/MediaOperationsTestingUITestContextExtensions.cs
@@ -34,6 +34,10 @@ public static class MediaOperationsTestingUITestContextExtensions
                 await context
                     .Get(By.CssSelector($"a[href*=\"/media/{imageName}\"]").OfAnyVisibility())
                     .ClickReliablyAsync(context);
+                // Closing the newly opened tab with the image, so the browser doesn't continue to switch the UI back
+                // and forth.
+                context.SwitchToLastWindow();
+                context.Driver.Close();
                 context.SwitchToFirstWindow();
 
                 context.WaitForPageLoad();
@@ -54,6 +58,8 @@ public static class MediaOperationsTestingUITestContextExtensions
                 await context
                     .Get(By.CssSelector($"a[href*=\"/media/{documentName}\"]"))
                     .ClickReliablyAsync(context);
+                context.SwitchToLastWindow();
+                context.Driver.Close();
                 context.SwitchToFirstWindow();
 
                 context.WaitForPageLoad();
--- a/Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
+++ b/Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
@@ -327,11 +327,11 @@ public class SecurityScanConfiguration
             //   pollPostData: ""
         }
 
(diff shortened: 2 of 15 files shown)


Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -481,7 +481,7 @@ private async Task SetupAsync()
 
             _dockerConfiguration = TestConfigurationManager.GetConfiguration<DockerConfiguration>();
 
-            var resultUri = await _currentSetupSnapshotManager.RunOperationAndSnapshotIfNewAsync(async () =>
```
Piedone: Please keep this. The setup operation needs to be able to create the URL the test should start from.

wAsnk: Okay, and in what form is it used currently? How can UI tests be started on different URLs currently? I can't seem to find it.

Piedone: It's this feature you removed here. The URL returned by the setup operation will be the starting URL of the test.

wAsnk: I luled, obviously this is what it was using it, but it is not anymore. So let me rephrase my question. Why is it needed to use this URL to navigate to the relative URL at the end of the setup? Is there a reason for that? If so, I don't think `resultUri.PathAndQuery` is the right way of doing it.

So e.g. what you say is if you set a URL of https://example.com/path/ to start the test on, you would get the relative Url `https://example.com/path/path` at line `await _context.GoToRelativeUrlAsync(resultUri.PathAndQuery);` which is obviously not okay.
So the correct navigation step is to navigate to `/`, which means `resultUri` isn't needed.

Piedone: Why would it be correct to always navigate to `/`? The reason we have the setup operation return a URL is precisely because that's not necessarily the case: the URL it returns is the start URL of the test. This is the URL that'll be open when control arrives at the beginning of a test's code.  The setup operation is not necessarily just filling out the setup screen, it can be anything to initialize the app to a testable state.

Lombiq.Tests.UI/Services/OrchardCoreInstance.cs
```diff
@@ -25,6 +25,13 @@ namespace Lombiq.Tests.UI.Services;
 public class OrchardCoreConfiguration
 {
     public string SnapshotDirectoryPath { get; set; }
+
+    /// <summary>
+    /// Gets or sets the base path of the application if mapped to a different URL. For example, if the Orchard Core
+    /// application is mapped to "/cms" then this should be "/cms/". The / is important at the end, otherwise during
```
Piedone: You can make sure this happens with `TrimEnd('/') + "/'"`.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
wAsnk: Did you understand my example?

wAsnk: > the URL it returns is the start URL of the test. This is the URL that'll be open when control arrives at the beginning of a test's code. The setup operation is not necessarily just filling out the setup screen, it can be anything to initialize the app to a testable state.

Where do you set this URL currently?

Lombiq.Tests.UI/SecurityScanning/SecurityScanConfiguration.cs
```diff
@@ -327,7 +327,11 @@ internal async Task ApplyToPlanAsync(YamlDocument yamlDocument, UITestContext co
             //   pollPostData: ""
         }
 
+        // We can do this or .ToArray() which results IDE0305: Collection initialization can be simplified. So let's use
+        // this.
+#pragma warning disable S3878 //S3878: Remove this array creation and simply pass the elements.
         yamlDocument.AddExcludePathsRegex([.. _excludedUrlRegexPatterns]);
+#pragma warning restore S3878
```
sarahelsaig: Could you use ToArray instead? When .NET 9 comes out, it will support `params` for all `IEnumerable` types so finding these by searching for ToArray will be much easier.

Lombiq.Tests.UI/Services/OrchardCoreInstance.cs
```diff
@@ -88,13 +95,13 @@ public async Task<Uri> StartUpAsync()
...
 
-        _reverseProxy = new TestReverseProxy(_url);
+        _reverseProxy = new TestReverseProxy(_url, _configuration.BasePath);
 
         await _reverseProxy.StartAsync();
 
         await StartOrchardAppAsync();
 
-        return new Uri(_url);
+        return new Uri(_url + _configuration.BasePath);
```
sarahelsaig: I think BasePath should work even if it's not prefixed with `/`. You can use [`new Uri(Uri baseUri, string relativeUri)`](https://learn.microsoft.com/en-us/dotnet/api/system.uri.-ctor?view=net-8.0#system-uri-ctor(system-uri-system-string)) instead.

sarahelsaig (changes requested):

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
Piedone: This URL is the URL to start the test on. The app may launch with a different URL, i.e. the _setup_ may start with one URL, but the _test_ can start with a second one. That's what the setup operation returns.

In any case, feel free to change whatever is necessary to make the prefixed base path work, but do not remove the ability for the setup operation to return a custom start URL for a test.

wAsnk: Could you show me an example where this is done? Because I don't understand how this could be done.

Piedone: It's in every setup operation, since otherwise you can't compile, e.g.: https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI.Samples/Helpers/SetupHelpers.cs#L46

wAsnk: Okay, now I understand what you mean, but it is really weird to do this:
image
Let's say you set the basepath to start the UI tests on for `/cms` and to operate with this basepath for everything. It means the setup will be started on that page and it will return `https://localhost:port/cms/` as the Uri (homePageUri variable here), and if I return that because I want to be on that page after the setup process then it will redirect to `https://localhost:port/cms/cms`. 

So I have to return a `https://localhost:port/` Uri, because in [GoToRelativeUrlAsync](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs#L25), the [GetAbsoluteUri](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs#L71) is called to create the Uri based on the `context.Scope.BaseUri` which is set to `https://localhost:port/cms/`. So I return  `https://localhost:port/` to get `https://localhost:port/cms` at `UITestExecutionSession` `await _context.GoToRelativeUrlAsync(resultUri.PathAndQuery);`

Or is this only weird for me? I can sleep with this in my mind, it's just weird.

Piedone: You can try `_context.GoToRelativeUrlAsync("/" + resultUri.PathAndQuery);` if `PathAndQuery` never contains a leading slash. That's also semantically more appropriate, since the assumption there is that the URL returned from the setup operation is always relative to the app root.

wAsnk: Yeah, that was my original thought. But that won't work because [GetAbsoluteUri](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs#L71) does `.TrimStart('/')` to the given Uri.

Lombiq.Tests.UI/BasicOrchardFeaturesTesting/MediaOperationsTestingUITestContextExtensions.cs
```diff
@@ -32,7 +32,7 @@ public static class MediaOperationsTestingUITestContextExtensions
                 context.Exists(By.XPath($"//span[contains(text(), '{imageName}')]"));
 
                 await context
-                    .Get(By.CssSelector($"a[href=\"/media/{imageName}\"]").OfAnyVisibility())
```
Piedone, line 35: Here and below rather utilize [`UrlPrefix`](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/Services/UITestContext.cs#L111).

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
Piedone: I discovered that we actually already have a mechanism for all of this, called `UrlPrefix`. So, rather do what [`SwitchCurrentTenant`](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/Services/UITestContext.cs#L246) does. Everything else should work without changes.

wAsnk: image

wAsnk: I swear I did not spend several hours on this.

wAsnk: Uhm this actually won't work because we also need a different base path during setup.

wAsnk: Okay nvm, I can still do it with this, but now it just seems like another solution, and not necessarily a better solution. But I can do it this way if you say so.

Piedone: Utilizing an existing mechanism, just also earlier, seems like the solution to go, yes.

Lombiq.Tests.UI/BasicOrchardFeaturesTesting/MediaOperationsTestingUITestContextExtensions.cs
```diff
@@ -32,7 +32,7 @@ public static class MediaOperationsTestingUITestContextExtensions
...
+                    .Get(By.CssSelector($"a[href=\"{context.UrlPrefix}/media/{imageName}\"]").OfAnyVisibility())
```
sarahelsaig: This breaks if the image has a version query string (for cache invalidation), which will be the default behavior starting with OC 2.0. (it's already like that in [`task/system-text-json-migration`](https://github.com/Lombiq/UI-Testing-Toolbox/blob/task/system-text-json-migration/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/MediaOperationsTestingUITestContextExtensions.cs#L35)) So instead of exact match you should check if the href starts with that path.
```suggestion
                    .Get(By.CssSelector($"a[href^=\"{context.UrlPrefix}/media/{imageName}\"]").OfAnyVisibility())
```

Lombiq.Tests.UI/BasicOrchardFeaturesTesting/MediaOperationsTestingUITestContextExtensions.cs
```diff
@@ -56,7 +56,7 @@ await context
                     .ClickReliablyAsync(context);
 
                 await context
-                    .Get(By.CssSelector($"a[href=\"/media/{documentName}\"]"))
+                    .Get(By.CssSelector($"a[href=\"{context.UrlPrefix}/media/{documentName}\"]"))
```
sarahelsaig: Same.
```suggestion
                    .Get(By.CssSelector($"a[href^=\"{context.UrlPrefix}/media/{documentName}\"]"))
```

sarahelsaig (changes requested):

PR #392, opened 2024-07-19
LMBQ-373: Fix visual verification tests instead of using Firefox and skipping them

DemeSzabolcs:
[LMBQ-373](https://lombiq.atlassian.net/browse/LMBQ-373)

[LMBQ-373]: https://lombiq.atlassian.net/browse/LMBQ-373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/UITestBase.cs
```diff
@@ -45,7 +45,7 @@ public abstract class UITestBase
 
             if (timeoutTask.IsCompleted)
             {
-                throw new TimeoutException($"The time allotted for the test ({timeout}) was exceeded.");
+                throw new TimeoutException($"The time allowed for the test ({timeout}) was exceeded.");
```
sarahelsaig: Revert this change. The phrase "the time allotted" is grammatically correct and appropriate for this specific context.
image

Lombiq.Tests.UI/Extensions/ScrollingUITestContextExtensions.cs
```diff
@@ -27,8 +27,21 @@ public static void ScrollTo(this UITestContext context, By by)
     /// <summary>
     /// Scrolls to a particular set of coordinates in the document.
     /// </summary>
-    public static void ScrollTo(this UITestContext context, int x, int y) =>
+    public static void ScrollTo(this UITestContext context, int x, int y)
+    {
         context.ExecuteScript("window.scrollTo(arguments[0], arguments[1], \"instant\");", x, y);
```
sarahelsaig: Since you are using apostrophes below as well:

```suggestion
        context.ExecuteScript("window.scrollTo(arguments[0], arguments[1], 'instant');", x, y);
```

Lombiq.Tests.UI/Extensions/ScrollingUITestContextExtensions.cs
```diff
@@ -27,8 +27,21 @@ public static void ScrollTo(this UITestContext context, By by)
...
+        context.TriggerHtmlReload();
+    }
+
+    /// <summary>
+    /// Sometimes after scrolling elements can't render properly, so we need to set the scale, to trigger the HTML
+    /// elements.
```
sarahelsaig:
```suggestion
    /// Sometimes elements don't render properly after scrolling. So we have to set the scale to re-render them.
```

sarahelsaig (changes requested):

Lombiq.Tests.UI/UITestBase.cs
DemeSzabolcs: Oh sorry 😅

PR #396, opened 2024-08-02
OSOE-365: Support for running tests without a browser

### LINKED ISSUE #79: Support for running tests without a browser (OSOE-365)
In certain cases, we need a running app but not a browser (or at least not a full one), like for testing web APIs. For such cases, we could start tests without a browser and then use something like RestEase for APIs or [ScrapySharp](https://github.com/rflechner/ScrapySharp "https://github.com/rflechner/ScrapySharp") as a barebone browser. The browser is launched when `AtataFactory.CreateDriver()` runs, this is the only thing necessary to switch off. However, a browser would still sometimes be needed for the initial setup (though we could add a feature to only open a browser when the setup is run but not for the test itself, or utilize Auto Setup). 

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-365)

Piedone:
[OSOE-365](https://lombiq.atlassian.net/browse/OSOE-365)
Fixes #79

[OSOE-365]: https://lombiq.atlassian.net/browse/OSOE-365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -77,35 +77,12 @@ public async Task<bool> ExecuteAsync(int retryCount, string dumpRootPath)
...
-                snapshotSubdirectory += "-" + setupConfiguration.SetupOperation!.GetHashCode().ToTechnicalString();
-
-                _snapshotDirectoryPath = Path.Combine(setupConfiguration.SetupSnapshotDirectoryPath, snapshotSubdirectory);
-
-                _configuration.OrchardCoreConfiguration.SnapshotDirectoryPath = _snapshotDirectoryPath;
-
-                _currentSetupSnapshotManager = UITestExecutionSessionsMeta.SetupSnapshotManagers.GetOrAdd(
-                    _snapshotDirectoryPath,
-                    path => new SynchronizingWebApplicationSnapshotManager(path));
-
```
Piedone, line 101: I just moved this due to the cognitive complexity of the method becoming too large (analyzer). Added `CaptureBrowserUsingDumpsAsync` for the same reason.

PR #402, opened 2024-08-11
LMBQ-178: Enhancing support for remote tests working with apps behind Cloudflare

Piedone:
[LMBQ-178](https://lombiq.atlassian.net/browse/LMBQ-178)

[LMBQ-178]: https://lombiq.atlassian.net/browse/LMBQ-178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Helpers/CloudflareHelper.cs
```diff
@@ -0,0 +1,218 @@
...
+                AuthorizationHeaderValueGetter = (_, _) => Task.FromResult(cloudflareApiToken),
+            });
+
+            _currentIp ??= _cloudflareApi != null ? await GetPublicIpAsync() : string.Empty;
+
+            if (_ipAccessRuleId == null)
+            {
+                Debug.WriteLine("Creating an IP Access Rule for the IP {0}.", (object)_currentIp);
+
+                // Delete any pre-existing rules for the current IP first.
```
DemeSzabolcs, line 46: Can't we somehow dry, the first deletion of the IP rules with the one at line 104? Are there any cases, where the rule is not deleted at the end on line 104 and we need this first one?

Lombiq.Tests.UI/CloudflareRemoteUITestBase.cs
```diff
@@ -0,0 +1,74 @@
...
+/// Base class for UI tests that run on a remote (i.e. not locally running) app behind Cloudflare. Gets around
+/// Cloudflare rejecting automated requests.
+/// </summary>
+public abstract class CloudflareRemoteUITestBase : RemoteUITestBase
+{
+    /// <summary>
+    /// Gets the Cloudflare account's ID, as indicated in the account homepage's URL on the Cloudflare dashboard. Note
+    /// that due to how the IP Access Rule management works, you may only have tests for apps behind a single Cloudflare
+    /// account in a given test project.
+    /// </summary>
```
DemeSzabolcs: Is it straight forward how to get the account's ID in CF? If not I would add a remark just as for the API token.

DemeSzabolcs (changes requested):

Lombiq.Tests.UI/Helpers/CloudflareHelper.cs
Piedone: We can't further, because yes, there is: the first deletion is about deleting any rules that are there by mistake. There shouldn't be any, if everything worked well, but in case a previous test failed in some way that prevented its rule deletion, then this is a safety net. If we don't do this, then the rule creation will fail.

And the second deletion is about the normal process, cleaning up after the test. This is what should normally happen, but there's no guarantee it'll succeed (like if the CF API is down for an hour).

DemeSzabolcs (approved):

PR #403, opened 2024-08-22
NEST-536: Add necessary security exceptions

sarahelsaig:
[NEST-536](https://lombiq.atlassian.net/browse/NEST-536)

[NEST-536]: https://lombiq.atlassian.net/browse/NEST-536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
@@ -92,7 +92,7 @@ public static class SecurityScanningUITestContextExtensions
                     ".*/(Login|ChangePassword)[?][rR]eturnUrl=.*", // #spell-check-ignore-line
                     6,
                     "Path Traversal",
-                    "Setting the ReturnUrl query parameter to a itself yields a false positive");
+                    "Setting the ReturnUrl query parameter to itself yields a false positive");
 
                 configuration.MarkScanRuleAsFalsePositiveForUrlWithRegex(
                     ".*/(Login|ChangePassword)[?][rR]eturnUrl=.*", // #spell-check-ignore-line
--- a/Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
@@ -85,6 +85,15 @@ public static class SecurityScanningUITestContextExtensions
                 // There is no need to security scan the admin dashboard.
                 configuration.ExcludeUrlWithRegex(@".*/Admin/.*");
 
+                // There is no need to security scan anything in Lombiq.Tests.UI.Shortcuts.
+                configuration.ExcludeUrlWithRegex(@".*/Lombiq.Tests.UI.Shortcuts/.*");
+
+                configuration.MarkScanRuleAsFalsePositiveForUrlWithRegex(
+                    ".*/(Login|ChangePassword)([?].*)?",
+                    6,
+                    "Path Traversal",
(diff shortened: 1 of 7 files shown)


Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
```diff
@@ -85,6 +85,21 @@ public static Task RunAndConfigureAndAssertFullSecurityScanForContinuousIntegrat
...
                 configuration.ExcludeUrlWithRegex(@".*/Admin/.*");
 
+                // There is no need to security scan anything in Lombiq.Tests.UI.Shortcuts.
+                configuration.ExcludeUrlWithRegex(@".*/Lombiq.Tests.UI.Shortcuts/.*");
+
+                configuration.MarkScanRuleAsFalsePositiveForUrlWithRegex(
+                    ".*/(Login|ChangePassword)[?][rR]eturnUrl=.*", // #spell-check-ignore-line
+                    6,
+                    "Path Traversal",
+                    "Setting the ReturnUrl query parameter to a itself yields a false positive");
```
Piedone:
```suggestion
                    "Setting the ReturnUrl query parameter to itself yields a false positive");
```

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
```diff
@@ -85,6 +85,21 @@ public static Task RunAndConfigureAndAssertFullSecurityScanForContinuousIntegrat
...
+
+                configuration.MarkScanRuleAsFalsePositiveForUrlWithRegex(
+                    ".*/(Login|ChangePassword)[?][rR]eturnUrl=.*", // #spell-check-ignore-line
+                    40018,
+                    "SQL Injection",
+                    "Setting the ReturnUrl query parameter to an SQL expression can't actually cause SQL Injection.");
```
Piedone: These three should rather be in the plans, so they're applied to every scan.

Piedone (changes requested):

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
Piedone: You didn't yet move `configuration.ExcludeUrlWithRegex(@".*/Lombiq.Tests.UI.Shortcuts/.*");` BTW.

sarahelsaig: I don't know how to do that for a filter with no rule ID. Also the `configuration.ExcludeUrlWithRegex(@".*/Admin/.*");` [right above it](https://github.com/Lombiq/UI-Testing-Toolbox/blob/53ec10c03ba3a010a52dc11f37b5817561a19479/Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs#L86) is the same kind, so I assumed if it was possible you would've already moved it into the plans.

Piedone: In the end, this is run, what should be possible in YML too: https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/SecurityScanning/YamlDocumentExtensions.cs#L107

I don't remember the reason for the admin config, but that perhaps makes more sense here: keep in mind that it's easier to override the C# config from a delegate, than modify the YML document. Maybe we want to override that. I can't imagine this for the shortcuts rule, though.

PR #405, opened 2024-08-29
OFFI-87: BE: Add endpoint testing

porgabi:
[OFFI-87](https://lombiq.atlassian.net/browse/OFFI-87)

[OFFI-87]: https://lombiq.atlassian.net/browse/OFFI-87?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Extensions/HttpClientUITestContextExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+        "Reliability",
+        "CA2000:Dispose objects before losing scope",
+        Justification = "Disposed by the HttpClient.")]
+public static class HttpClientUITestContextExtensions
+{
+    /// <summary>
+    /// Creates a new <see cref="HttpClient"/> and authorizes it with a Bearer token that is created based on the provided
+    /// <paramref name="clientId"/> and <paramref name="clientSecret"/>.
+    /// </summary>
+    public static async Task<HttpClient> CreateAndAuthorizeClientAsync(
```
wAsnk, line 31: Why don't you use the https://github.com/Lombiq/Orchard-Core-API-Client library of ours?

porgabi: Seemed simpler this way with no drawbacks. What would be the advantages of using the library?

wAsnk: I think you are reusing fairly big part of it already by copying code from it. Isn't it possible to DRY this with that and abstracting the already existing logic and doing the same from there?

porgabi: I only took the `HttpClientHandler` settings and its suppression messages from the library.

The issue I have with using the library is it also entails adding new interface(s) for our app's endpoints, if I understand correctly (just like the existing ones for e.g. [tenant-related operations](https://github.com/Lombiq/Orchard-Core-API-Client/blob/dev/Lombiq.OrchardCoreApiClient/Interfaces/IOrchardCoreTenantsApi.cs)). While I think we could use the ApiClient class instead of the generic HttpClient one to issue requests, I don't see the upside of making it work, especially as the solution I added is even more simple and doesn't require more submodules.

It's possible that I'm also biased here and not entirely objective (frankly, this task has become very tedious), so if you definitely think using the library would be a better approach, I'll do it, I just haven't found good enough reason to do so so far.

wAsnk: You don't necessarily need to create interfaces. You can create a new class like the ApiClient, but it would be for your usecase. But you also need to make sure about disposing the HTTP client like it is done in ApiClient. Because it seems like the memory usage is quite high in OFFI compared to OSOCE where we have 4x more UI tests, so that's weird.

wAsnk: Or if that just doesn't seem worth it, make sure you dispose the httpclient here as well.

porgabi: Disposing the `HttpClient`s doesn't seem to make a difference in memory usage, e.g. [build before](https://camo.githubusercontent.com/296e3c2b0239061aa539df475bc5140b5032986aa4630feb8e49e59101bbd2ef/68747470733a2f2f6170692e676c6f62616467652e636f6d2f76312f636861727467656e2f737461636b65642d617265612f74696d652f63686172745f737461636b65645f617265615f74696d655f37643334383034312d383966362d346235372d386561362d643965636366313633643664) vs [build after](https://camo.githubusercontent.com/f16c1ebad02f9f9a04fe4a304fcf2079b9a7802fbace378ad821a41db811f5f5/68747470733a2f2f6170692e676c6f62616467652e636f6d2f76312f636861727467656e2f737461636b65642d617265612f74696d652f63686172745f737461636b65645f617265615f74696d655f35353938366334382d626464372d346630322d396239352d306631323236326466306533). I think the usage is higher in the latter because one of the tests failed, but yeah.

wAsnk: Ok, revert disposing.

wAsnk (approved):

PR #406, opened 2024-09-05
OSOE-732: Generalize Failure Dump to Test Dump

### LINKED ISSUE #323: Generalize Failure Dump to Test Dump (OSOE-732)
Failure Dump is a useful feature to provide various output data in the form of files to the person inspecting tests. However, we sometimes want to provide such output even if the test passes. There are two cases already:

- Accessibility reports, see `AccessibilityCheckingConfiguration.CreateReportAlways`. This has special treatment now in [`UITestExecutor`](https://github.com/Lombiq/UI-Testing-Toolbox/blob/f6dab68c49529e708d5a71516d1cabfab8d8c082/Lombiq.Tests.UI/Services/UITestExecutor.cs#L46).
- Security scan reports, see `SecurityScanningConfiguration.CreateReportAlways`. This is not even working until this issue is completed.

So, rename Failure Dump (and everything related to it in the code, docs, and in [LGHA](https://github.com/Lombiq/GitHub-Actions)) to Test Dump and change how it works:

- Always generate it if `UITestContext.FailureDumpContainer` (which should be renamed) is non-empty.
- Remove the special treatment of  `AccessibilityCheckingConfiguration.CreateReportAlways`.
- Make `SecurityScanningConfiguration.CreateReportAlways` public and remove the note from its docs about this issue.

Once done, deprecate `SecurityScanningConfiguration.CreateReportOnTestFailAlways`.

Related: https://github.com/xunit/xunit/issues/2457

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-732)

Piedone:
[OSOE-732](https://lombiq.atlassian.net/browse/OSOE-732)
Fixes #323

[OSOE-732]: https://lombiq.atlassian.net/browse/OSOE-732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Exceptions/TestDumpItemAlreadyExistsException.cs
```diff
@@ -4,27 +4,27 @@ namespace Lombiq.Tests.UI.Exceptions;
...
+    public TestDumpItemAlreadyExistsException(string fileName, string customMessage)
         : this(fileName, customMessage, innerException: null)
     {
     }
 
-    public FailureDumpItemAlreadyExistsException(string fileName, string customMessage, Exception innerException)
+    public TestDumpItemAlreadyExistsException(string fileName, string customMessage, Exception innerException)
         : base(
-            $"A failure dump item with the same file name already exists. fileName: {fileName}"
+            $"A Test dump item with the same file name already exists. File name: {fileName}."
```
sarahelsaig: I guess the capitalization here was accidental?
```suggestion
            $"A test dump item with the same file name already exists. File name: {fileName}."
```

Lombiq.Tests.UI/Models/TestDumpItemGeneric.cs
```diff
@@ -4,21 +4,21 @@
...
-    public FailureDumpItemGeneric(
+    public TestDumpItemGeneric(
         TContent content,
         Func<TContent, Task<Stream>> getStream = null,
         Action<TContent> dispose = null)
     {
         if (content is not Stream && getStream == null)
         {
-            throw new ArgumentException($"{nameof(content)} is not a Stream, {nameof(getStream)} can't be null");
+            throw new ArgumentException($"{nameof(content)} is not a Stream and {nameof(getStream)} can't be null.");
```
sarahelsaig: It would be better to phrase the exception message as an instruction than a status report. Easier to understand during debugging,

```suggestion
            throw new ArgumentException(
                $"{nameof(content)} must be of type {nameof(Stream)} or {nameof(getStream)} must not be null.");
```

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -109,15 +111,19 @@ public async Task<bool> ExecuteAsync(int retryCount, string dumpRootPath)
...
+                dumpContainerPath => FailureTestDumpProcessAsync(dumpContainerPath, ex));
 
             if (_context?.IsFinalTry == true || retryCount >= _configuration.MaxRetryCount)
             {
                 var dumpFolderAbsolutePath = Path.Combine(AppContext.BaseDirectory, dumpRootPath);
 
                 _testOutputHelper.WriteLineTimestampedAndDebug(
                     "The test was attempted {0} time(s) and won't be retried anymore. You can see more details " +
-                        "on why it's failing in the FailureDumps folder: {1}",
+                        "on why it's failing in the TestDumps folder: {1}",
```
sarahelsaig: It's not necessary to re-state the folder name, it's already in the `dumpFolderAbsolutePath`.
```suggestion
                        "on why it's failing in the test dump folder: {1}",
```

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -221,16 +227,33 @@ private Exception PrepareAndLogException(Exception ex)
...
+    {
+        if (_context == null) return;
+
+        var debugInformationPath = GetDebugInformationPath(dumpContainerPath);
+
+        if (_context.IsBrowserRunning) await CaptureBrowserUsingDumpsAsync(debugInformationPath);
+
+        if (_dumpConfiguration.CaptureAppSnapshot) await CaptureAppSnapshotAsync(dumpContainerPath);
+
+        CaptureMarkupValidationResults(ex, debugInformationPath);
```
sarahelsaig: Each line in this method is separated with a blank line, which doesn't seem readable.
Can we group the `Capture*()` method calls together?
```suggestion
        if (_context.IsBrowserRunning) await CaptureBrowserUsingDumpsAsync(debugInformationPath);
        if (_dumpConfiguration.CaptureAppSnapshot) await CaptureAppSnapshotAsync(dumpContainerPath);
        CaptureMarkupValidationResults(ex, debugInformationPath);
```

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -221,16 +227,33 @@ private Exception PrepareAndLogException(Exception ex)
...
+    private async Task CreateTestDumpAsync(
         string dumpRootPath,
         int retryCount,
-        IDictionary<string, IFailureDumpItem> failureDumpContainer)
+        IDictionary<string, ITestDumpItem> testDumpContainer,
+        Func<string, Task> additionalDumpProcess = null)
     {
-        if (!_dumpConfiguration.CreateFailureDump) return;
+        if (!_dumpConfiguration.CreateTestDump ||
+            ((testDumpContainer == null || !testDumpContainer.Any()) && additionalDumpProcess == null))
```
sarahelsaig: imo it's easier to understand with less parentheses
```suggestion
        if (!_dumpConfiguration.CreateTestDump ||
            (testDumpContainer?.Any() != true && additionalDumpProcess == null))
```

sarahelsaig (changes requested):

Lombiq.Tests.UI/Exceptions/TestDumpItemAlreadyExistsException.cs
Piedone: Yep!

Lombiq.Tests.UI/Models/TestDumpItemGeneric.cs
Piedone: Yep.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
Piedone: Indeed.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
Piedone: Yeah, better like that.

PR #407, opened 2024-09-14
OFFI-101: Frontend server support and interactive mode improvements

sarahelsaig:
[OFFI-101](https://lombiq.atlassian.net/browse/OFFI-101)

[OFFI-101]: https://lombiq.atlassian.net/browse/OFFI-101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wAsnk (approved):

wAsnk (changes requested):
I believe this should be documented in the Readme that it is now possible to do such thing like frontend UI testing. Please document it.

sarahelsaig:
Instead, I think a sample should be created in a separate future OSOCE issue. Because just documenting that it's possible is not that useful since it's not a trivial process.

wAsnk (approved):

PR #409, opened 2024-09-23
OSOE-818: Upgrade to Orchard Core 2.0

sarahelsaig:
[OSOE-818](https://lombiq.atlassian.net/browse/OSOE-818)

[OSOE-818]: https://lombiq.atlassian.net/browse/OSOE-818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Piedone (approved):

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
Piedone, line 1: Check ln 597. Can't we use `LinkGenerator` or `UrlHelper`?

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
sarahelsaig: No, they still don't return anything. I also noticed that `serviceProvider.GetService<IActionContextAccessor>().ActionContext` is `null`. As far as I understand, both `LinkGenerator` and `UrlHelper` rely on that context.

Also `serviceProvider.GetRequiredService<LinkGenerator>()` returns a `DefaultLinkGenerator` with 0 endpoints.

Piedone: OK then. Updated the docs to remove the issue link, since that's not relevant anymore.

Piedone (approved):

PR #419, opened 2024-11-08
OSOE-402: Custom logging instead of NLog-based one

### LINKED ISSUE #206: Custom logging instead of NLog-based one (OSOE-402)

Piedone:
[OSOE-402](https://lombiq.atlassian.net/browse/OSOE-402)
Fixes #206

**To be added to the release notes on the next release:**

## Breaking changes in logging during UI tests

One of the implicit assertions that the UI Testing Toolbox does automatically is checking the Orchard Core application logs for any errors. By default, any error fails the test.

Previously, this relied on the tested web app using NLog for logging, and it parsed NLog log files. This not only required the app to use NLog over Serilog (that Orchard also supports out of the box) or any other logging framework, but it also depended on the structure of an otherwise freely configurable textual log format.

Now, the UI Testing Toolbox configures the app to use [`FakeLogger`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.testing.fakelogger) in addition to any logging framework it has set up, and uses it for assertions. `FakeLogger` not only doesn't need anything specific in the app, but it also lets you access and assert on log entries in a structured, statically typed way.

Using NLog is still supported, and log files will be included in the test dump. Any other logging framework is also supported, your app doesn't need to do anything special.

### What changed

- All log management methods are now async.
- Log assertions now don't happen on a string representing the whole log, but rather on `IApplicationLog` and `IApplicationLogEntry` instances.
- The `LogsShouldBeEmptyAsync` and `GetLogOutputAsync` methods have changed, see [the diff](https://github.com/Lombiq/UI-Testing-Toolbox/pull/419/files#diff-cf3bc5fe84fc8a945f13d316362a82540659e238e71462e8c37ad43509919ac8).
- `CreateAppLogAssertionForSecurityScan` and `CreateAppLogAssertionForSecurityScan` work the same but were moved into the `Lombiq.Tests.UI.SecurityScanning` namespace.
- A new `OrchardCoreConfiguration.AfterFakeLoggingConfiguration` event handler is exposed to change the default configuration of `FakeLogger`.
- `OrchardCoreUITestExecutorConfiguration.AssertAppLogsCanContainWarningsAsync` and `OrchardCoreUITestExecutorConfiguration.AssertAppLogsCanContainWarningsAndCacheFolderErrorsAsync` are no longer available. 

### How to adapt

- Use `OrchardCoreUITestExecutorConfiguration.AssertAppLogsCanContainCacheFolderErrorsAsync` instead of  `OrchardCoreUITestExecutorConfiguration.AssertAppLogsCanContainWarningsAndCacheFolderErrorsAsync` .
- `OrchardCoreUITestExecutorConfiguration.AssertAppLogsCanContainWarningsAsync` is not needed, since the new behavior is to not log warnings in the first place.
- The examples are updated in `Lombiq.Tests.UI.Samples.UITestBase`, see [the diff](https://github.com/Lombiq/UI-Testing-Toolbox/pull/419/files#diff-9576924746d4544c4ff5da0ea6aa77390c83a8548084a3eafb5c7a0ea840f8b1). Check them out if you have custom app log assertion code.

[OSOE-402]: https://lombiq.atlassian.net/browse/OSOE-402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/Extensions/ApplicationLogEnumerableExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/ApplicationLogEnumerableExtensions.cs
@@ -17,14 +17,12 @@ public static class ApplicationLogEnumerableExtensions
             return Environment.NewLine + await LogLinesToFormattedStringAsync(logsArray[0]);
         }
 
-        return string.Join(
-            Environment.NewLine + Environment.NewLine,
-            await Task.WhenAll(
-                logsArray.Select(async log =>
-                    $"# Log name: {log.Name}" +
-                    Environment.NewLine +
-                    Environment.NewLine +
-                    await LogLinesToFormattedStringAsync(log))));
+        // Parallelization with Task.WhenAll() isn't really necessary for performance here but would potentially change
+        // the order of the logs in the output.
+        var logContents = logsArray.AwaitEachAsync(async log =>
+            $"# Log name: {log.Name}" + Environment.NewLine + Environment.NewLine + await LogLinesToFormattedStringAsync(log));
+
+        return string.Join(Environment.NewLine + Environment.NewLine, logContents);
     }
 
     private static async Task<string> LogLinesToFormattedStringAsync(IApplicationLog log) =>
--- a/Lombiq.Tests.UI/Extensions/ApplicationLogEnumerableExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/ApplicationLogEnumerableExtensions.cs
@@ -13,8 +13,8 @@ public static class ApplicationLogEnumerableExtensions
             Environment.NewLine + Environment.NewLine,
             await Task.WhenAll(
                 logs.Select(async log =>
-                    log.Name +
+                    $"# Log name: {log.Name}" +
                     Environment.NewLine +
                     Environment.NewLine +
-                    (await log.GetContentAsync()).Select(logMessage => logMessage.ToString()))));
+                    string.Join(Environment.NewLine, (await log.GetEntriesAsync()).Select(logEntry => logEntry.ToString())))));
 }
--- a/Lombiq.Tests.UI/Extensions/ApplicationLogEnumerableExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/ApplicationLogEnumerableExtensions.cs
@@ -11,5 +11,10 @@ public static class ApplicationLogEnumerableExtensions
     public static async Task<string> ToFormattedStringAsync(this IEnumerable<IApplicationLog> logs) =>
         string.Join(
             Environment.NewLine + Environment.NewLine,
-            await Task.WhenAll(logs.Select(async log => log.Name + Environment.NewLine + Environment.NewLine + await log.GetContentAsync())));
+            await Task.WhenAll(
+                logs.Select(async log =>
+                    log.Name +
+                    Environment.NewLine +
+                    Environment.NewLine +
+                    (await log.GetContentAsync()).Select(logMessage => logMessage.ToString()))));
 }
--- a/Lombiq.Tests.UI/Extensions/ApplicationLogEnumerableExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/ApplicationLogEnumerableExtensions.cs
@@ -8,13 +8,25 @@ namespace Lombiq.Tests.UI.Extensions;
 
 public static class ApplicationLogEnumerableExtensions
 {
-    public static async Task<string> ToFormattedStringAsync(this IEnumerable<IApplicationLog> logs) =>
-        string.Join(
+    public static async Task<string> ToFormattedStringAsync(this IEnumerable<IApplicationLog> logs)
+    {
+        var logsArray = logs.ToArray();
+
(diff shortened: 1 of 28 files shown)


Lombiq.Tests.UI/Extensions/ApplicationLogEnumerableExtensions.cs
```diff
@@ -8,8 +8,25 @@ namespace Lombiq.Tests.UI.Extensions;
...
+
+        return string.Join(
             Environment.NewLine + Environment.NewLine,
-            await Task.WhenAll(logs.Select(async log => log.Name + Environment.NewLine + Environment.NewLine + await log.GetContentAsync())));
+            await Task.WhenAll(
+                logsArray.Select(async log =>
+                    $"# Log name: {log.Name}" +
+                    Environment.NewLine +
+                    Environment.NewLine +
+                    await LogLinesToFormattedStringAsync(log))));
```
dministro: This approach may change the order of the log records. I would be surprised if we will have more than a few hundred log records here. Do we need parallelization?

Lombiq.Tests.UI/Models/FakeLoggerLogApplicationLog.cs
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public Task RemoveAsync()
+    {
+        LogCollector.Clear();
+        return Task.CompletedTask;
+    }
+}
+
+public sealed class FakeLoggerApplicationLogEntry : IApplicationLogEntry
```
dministro: This can be a `record`.
```csharp
public record FakeLoggerApplicationLogEntry(FakeLogRecord LogRecord) : IApplicationLogEntry
{
    public LogLevel Level => LogRecord.Level;
    public EventId Id => LogRecord.Id;
    public Exception Exception => LogRecord.Exception;
    public string Message => LogRecord.Message;
    public string Category => LogRecord.Category;
    public DateTimeOffset Timestamp => LogRecord.Timestamp;

    public override string ToString() => FormatLogRecord(LogRecord);

    public static string FormatLogRecord(FakeLogRecord record) =>
        StringHelper.CreateInvariant($"{record.Timestamp:yyyy-MM-dd HH:mm:ss.fff} [{record.Level}] {record.Category}: {record.Message}") +
        (record.Exception != null ? record.Exception.ToString() : string.Empty);
}
```

Lombiq.Tests.UI/Services/SynchronizingWebApplicationSnapshotManager.cs
```diff
@@ -46,7 +46,6 @@ public async Task<Uri> RunOperationAndSnapshotIfNewAsync(AppInitializer appIniti
 
             var result = await appInitializer();
             await result.Context.Application.TakeSnapshotAsync(_snapshotDirectoryPath);
-            await result.Context.Application.ResumeAsync();
```
dministro, line 49: Note to myself: This was a question mark, but the instance is resumed in `TakeSnapshotInnerAsync`.

dministro (changes requested):

Lombiq.Tests.UI/Extensions/ApplicationLogEnumerableExtensions.cs
Piedone: Indeed, parallelization is not really needed here (and in the overwhelming majority of cases we'll have a single log anyway).

Lombiq.Tests.UI/Services/SynchronizingWebApplicationSnapshotManager.cs
Piedone: Yeah, previously `TakeSnapshotInnerAsync` lacked the resume, but since it stops the instance, it's proper to have the resume there and not force the consumer to do it (after discovering that the instance was unexpectedly stopped).

Lombiq.Tests.UI/Models/FakeLoggerLogApplicationLog.cs
Piedone: Good idea!

dministro (approved):

dministro (approved):

PR #421, opened 2024-11-17
OSOE-904: Create an example of using FrontendServer and context.ExecuteJavascriptTestAsync

### LINKED ISSUE #408: Create an example of using FrontendServer and context.ExecuteJavascriptTestAsync() (OSOE-904)
This is a followup of #407. As mentioned in https://github.com/Lombiq/UI-Testing-Toolbox/pull/407#issuecomment-2367748945 there should be a sample test with demonstration to make these features accessible.

- Create a setup method demonstrating `FrontendServer` using a simple-to-setup standalone web server. Ideally something we can initiate directly from CLI without any setup code. [`http-server`](https://www.npmjs.com/package/http-server) from NPM may be a good candidate.
- Add test interacting with this frontend with the usual comment documentations.
- The same, selenium node.js. (may need some code generalized and migrated from OFFI)
- Add mention and links to Readme.md.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-904)

sarahelsaig:
[OSOE-904](https://lombiq.atlassian.net/browse/OSOE-904)
Fixes #408

[OSOE-904]: https://lombiq.atlassian.net/browse/OSOE-904?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI.Samples/FrontendUITestBase.cs
```diff
@@ -0,0 +1,97 @@
...
+
+namespace Lombiq.Tests.UI.Samples;
+
+// Sometimes Orchard Core is used in a headless manner, as a web API server that the visitors reach through a web
+// frontend (for example a Vue or React single page application). In this scenario you can test the API directly and
+// test the frontend with a dummy backend separately, but that will only get you so far. You'll want some tests that
+// ensure the two work together. To make this happen, you need some custom logic to initialize the frontend process
+// after setup, with a unique port number to avoid clashes. Also the frontend may need the backend URL, whose port is
+// already randomized with every test.
+public abstract class FrontendUITestBase : UITestBase
```
Piedone, line 19: Explain how much of this is necessary in every case, and what's specific to this sample but can be different based on the use case.

Lombiq.Tests.UI.Samples/Tests/FrontendTests.cs
```diff
@@ -0,0 +1,46 @@
...
+    // served by frontend server the same way as usual. In this case we have an HTTP file server, so you can navigate
+    // these directories and files. If we had a large client application that interacts with the headless OC in the
+    // backend, we would be able to do something more interesting but that's outside the scope of this demo.
+    [Fact]
+    public Task FrontendServerShouldStartWithTest() =>
+        ExecuteFrontendTestAfterSetupAsync(
+            async context =>
+            {
+                // Don't forget that if you want to interact with the frontend manually, you can just switch the context
+                // back to the back end and use the interactive mode extension method.
```
Piedone: Can't you interact with the frontend in interactive mode too?

Lombiq.Tests.UI.Samples/Tests/JavascriptTests.cs
```diff
@@ -0,0 +1,67 @@
+using Lombiq.Tests.UI.Extensions;
+using Lombiq.Tests.UI.Services.GitHub;
+using System.IO;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Lombiq.Tests.UI.Samples.Tests;
+
+// Suppose you want to write UI tests in Javascript. Why would you want to do that? Unlikely if you are an Orchard Core
```
Piedone: Everywhere it should be spelled "JavaScript".

Lombiq.Tests.UI.Samples/Tests/FrontendTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        : base(testOutputHelper)
+    {
+    }
+
+    // The interesting details were in FrontendUITestBase, here we just show that you can freely interact with the pages
+    // served by frontend server the same way as usual. In this case we have an HTTP file server, so you can navigate
+    // these directories and files. If we had a large client application that interacts with the headless OC in the
+    // backend, we would be able to do something more interesting but that's outside the scope of this demo.
+    [Fact]
+    public Task FrontendServerShouldStartWithTest() =>
```
Piedone, line 21: This fails for me with the following output:

```
 Lombiq.Tests.UI.Samples.Tests.FrontendTests.FrontendServerShouldStartWithTest
   Source: FrontendTests.cs line 21
   Duration: 2 min

  Message: 
System.IO.IOException : The directory under E:\ShortPath\OSOCE2\test\Lombiq.UITestingToolbox\Lombiq.Tests.UI.Samples\bin\Debug\net8.0\Temp\4152376e-6c07-435c-960b-f873253fa83c couldn't be cleaned up even after 10 attempts.
---- System.IO.IOException : The process cannot access the file '\\?\E:\ShortPath\OSOCE2\test\Lombiq.UITestingToolbox\Lombiq.Tests.UI.Samples\bin\Debug\net8.0\Temp\4152376e-6c07-435c-960b-f873253fa83c\App' because it is being used by another process.

  Stack Trace: 
DirectoryHelper.SafelyDeleteDirectoryIfExists(String path, Int32 maxTryCount) line 37
UITestExecutionSession.ShutdownAsync() line 185
UITestExecutor.ExecuteOrchardCoreTestInnerAsync(WebApplicationInstanceFactory webApplicationInstanceFactory, UITestManifest testManifest, OrchardCoreUITestExecutorConfiguration configuration, String dumpRootPath) line 84
UITestBase.ExecuteOrchardCoreTestAsync(WebApplicationInstanceFactory webApplicationInstanceFactory, UITestManifest testManifest, OrchardCoreUITestExecutorConfiguration configuration) line 66
OrchardCoreUITestBase`1.ExecuteTestAsync(Func`2 testAsync, Browser browser, Func`2 setupOperation, Func`2 changeConfigurationAsync) line 383
--- End of stack trace from previous location ---
----- Inner Stack Trace -----
FileSystem.RemoveDirectoryRecursive(String fullPath, WIN32_FIND_DATA& findData, Boolean topLevel)
FileSystem.RemoveDirectory(String fullPath, Boolean recursive)
DirectoryHelper.SafelyDeleteDirectoryIfExists(String path, Int32 maxTryCount) line 19

  Standard Output: 
[...]
2024-11-17 22:51:35.4185 - Attempting to start the Orchard Core instance.
2024-11-17 22:51:36.9974 - http-server (test frontend): Starting up http-server, serving ./
2024-11-17 22:51:36.9976 - http-server (test frontend): 
2024-11-17 22:51:36.9977 - http-server (test frontend): http-server version: 14.1.1
2024-11-17 22:51:36.9977 - http-server (test frontend): 
2024-11-17 22:51:36.9977 - http-server (test frontend): http-server settings: 
2024-11-17 22:51:36.9978 - http-server (test frontend): CORS: disabled
2024-11-17 22:51:36.9978 - http-server (test frontend): Cache: 3600 seconds
2024-11-17 22:51:36.9978 - http-server (test frontend): Connection Timeout: 120 seconds
2024-11-17 22:51:36.9978 - http-server (test frontend): Directory Listings: visible
2024-11-17 22:51:36.9978 - http-server (test frontend): AutoIndex: visible
2024-11-17 22:51:36.9978 - http-server (test frontend): Serve GZIP Files: false
2024-11-17 22:51:36.9978 - http-server (test frontend): Serve Brotli Files: false
2024-11-17 22:51:36.9978 - http-server (test frontend): Default File Extension: none
2024-11-17 22:51:36.9979 - http-server (test frontend): 
2024-11-17 22:51:36.9979 - http-server (test frontend): Available on:
2024-11-17 22:51:36.9979 - http-server (test frontend):   http://192.168.0.17:9364
2024-11-17 22:51:36.9979 - http-server (test frontend):   http://127.0.0.1:9364
2024-11-17 22:51:36.9979 - http-server (test frontend):   http://172.24.64.1:9364
2024-11-17 22:51:36.9979 - http-server (test frontend): Hit CTRL-C to stop the server
2024-11-17 22:51:36.9979 - http-server (test frontend): Wait operation has concluded.
2024-11-17 22:51:36.9981 - The Orchard Core instance was started.
2024-11-17 22:51:36.9984 - http-server (test frontend): 
2024-11-17 22:51:36.9985 - Finished waiting for the setup operation.
2024-11-17 22:51:36.9996 - Attempting to stop the Orchard Core instance.
2024-11-17 22:51:36.9997 - The Orchard Core instance was stopped.
2024-11-17 22:52:37.5548 - An exception has occurred while interacting with the page https://localhost:9365/ (Blog - Lombiq's OSOCE - UI Testing).
2024-11-17 22:52:37.5609 - The test failed with the following exception: System.IO.IOException: The directory under E:\ShortPath\OSOCE2\test\Lombiq.UITestingToolbox\Lombiq.Tests.UI.Samples\bin\Debug\net8.0\Temp\4152376e-6c07-435c-960b-f873253fa83c\App couldn't be cleaned up even after 60 attempts.
 ---> System.IO.IOException: The process cannot access the file '\\?\E:\ShortPath\OSOCE2\test\Lombiq.UITestingToolbox\Lombiq.Tests.UI.Samples\bin\Debug\net8.0\Temp\4152376e-6c07-435c-960b-f873253fa83c\App' because it is being used by another process.
   at System.IO.FileSystem.RemoveDirectoryInternal(String fullPath, Boolean topLevel, Boolean allowDirectoryNotEmpty)
   at System.IO.FileSystem.RemoveDirectoryRecursive(String fullPath, WIN32_FIND_DATA& findData, Boolean topLevel)
   at System.IO.FileSystem.RemoveDirectory(String fullPath, Boolean recursive)
   at Lombiq.Tests.UI.Helpers.DirectoryHelper.SafelyDeleteDirectoryIfExists(String path, Int32 maxTryCount) in E:\ShortPath\OSOCE2\test\Lombiq.UITestingToolbox\Lombiq.Tests.UI\Helpers\DirectoryHelper.cs:line 19
   --- End of inner exception stack trace ---
   at Lombiq.Tests.UI.Helpers.DirectoryHelper.SafelyDeleteDirectoryIfExists(String path, Int32 maxTryCount) in E:\ShortPath\OSOCE2\test\Lombiq.UITestingToolbox\Lombiq.Tests.UI\Helpers\DirectoryHelper.cs:line 37
   at Lombiq.Tests.UI.Services.OrchardCoreInstance`1.DisposeAsync() in E:\ShortPath\OSOCE2\test\Lombiq.UITestingToolbox\Lombiq.Tests.UI\Services\OrchardCoreInstance.cs:line 146
   at Lombiq.Tests.UI.Services.UITestExecutionSession.ShutdownAsync() in E:\ShortPath\OSOCE2\test\Lombiq.UITestingToolbox\Lombiq.Tests.UI\Services\UITestExecutionSession.cs:line 157
   at Lombiq.Tests.UI.Services.UITestExecutionSession.SetupAsync() in E:\ShortPath\OSOCE2\test\Lombiq.UITestingToolbox\Lombiq.Tests.UI\Services\UITestExecutionSession.cs:line 510
   at Lombiq.Tests.UI.Services.UITestExecutionSession.ExecuteAsync(Int32 retryCount, String dumpRootPath) in E:\ShortPath\OSOCE2\test\Lombiq.UITestingToolbox\Lombiq.Tests.UI\Services\UITestExecutionSession.cs:line 81
2024-11-17 22:52:37.671 TRACE > Take page snapshot #01 TestDumpPageSnapshot
2024-11-17 22:52:41.950 TRACE < Take page snapshot #01 TestDumpPageSnapshot (4.278s) >> "01 Orchard Core Setup page - TestDumpPageSnapshot.mhtml"
2024-11-17 22:52:42.0149 - The test was attempted 1 time(s) and won't be retried anymore. You can see more details on why it's failing in the test dump folder: E:\ShortPath\OSOCE2\test\Lombiq.UITestingToolbox\Lombiq.Tests.UI.Samples\bin\Debug\net8.0\TestDumps\FrontendServerShouldStartWithTest-c47a144e764ea857b0538dbb05b29f6674d956b47f993eb30c74b8d32d16177f
2024-11-17 22:52:42.015 TRACE > Deinitialize AtataContext
2024-11-17 22:52:42.121 TRACE < Deinitialize AtataContext (0.106s)
2024-11-17 22:52:42.123 DEBUG Finished test
      Total time: 1m 28.676s
  Initialization:     0.046s |  0.1 %
       Test body: 1m 28.522s | 99.8 %
Deinitialization:     0.106s |  0.1 %
```

Lombiq.Tests.UI.Samples/Tests/test.mjs
Piedone, line 1: The file name would be better as something clearer like _ExampleJavascriptTestShouldWork.mjs_.

Lombiq.Tests.UI.Samples/Tests/JavascriptTests.cs
```diff
@@ -0,0 +1,67 @@
...
+    // Using this approach you only have to write minimal C# boilerplate, which you can see below.
+    [Fact]
+    public Task ExampleJavascriptTestShouldWork() =>
+        ExecuteTestAfterSetupAsync(context =>
+        {
+            // Don't forget to mark the script files as "Copy if newer", so they are available to the test. It's best to
+            // include something like the following in your csproj file:
+            // <None Update="Tests\*.mjs" CopyToOutputDirectory="PreserveNewest" />
+            var workingDirectory = "Tests";
+            var scriptPath = Path.Join(workingDirectory, "test.mjs");
```
Piedone: These two seem like values that can be computed by default apart from the test script name, and used as is in the majority of cases, instead of having to pass them in. Also in `Sandbox()`.

Lombiq.Tests.UI/Extensions/FrontendOrchardCoreUITestExecutorConfigurationExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Lombiq.Tests.UI.Extensions;
+
+public static class FrontendOrchardCoreUITestExecutorConfigurationExtensions
+{
+    private const string BackendUri = nameof(BackendUri);
+    private const string FrontendUri = nameof(FrontendUri);
+
+    /// <summary>
+    /// Returns the start URLs for the Vite frontend and the Orchard Core backend from the <see
```
Piedone: Is the Vite reference here and below deliberate?

Lombiq.Tests.UI/Extensions/FrontendUITestContextExtensions.cs
```diff
@@ -13,6 +13,34 @@ namespace Lombiq.Tests.UI.Extensions;
...
+    public static void SwitchToBackend(this UITestContext context) =>
+        context.SwitchCurrentTenant(
+            tenantName: null,
+            context.Configuration.GetFrontendAndBackendUris().BackendUri);
+
+    /// <summary>
+    /// Navigates to the frontend <see cref="Uri"/> returned by <see
+    /// cref="FrontendOrchardCoreUITestExecutorConfigurationExtensions.GetFrontendAndBackendUris"/> and presents it as
+    /// switching to a tenant named <see cref="FrontendPseudoTenantName"/> which is not a real Orchard Core tenant so
+    /// this information can only be used for information.
```
Piedone: What do you mean can be used only for information? BTW there's a duplicate of "information" there.

Lombiq.Tests.UI/Extensions/FrontendUITestContextExtensions.cs
```diff
@@ -65,4 +124,88 @@ await Cli.Wrap(command)
...
         }
     }
+
+    /// <summary>
+    /// Invokes <see cref="ShortcutsUITestContextExtensions.SwitchToInteractiveAsync"/> with a custom notification
+    /// message that contains instructions to invoke the Javascript test manually with <c>node</c>.
+    /// </summary>
+    /// <param name="scriptPath">The relative or absolute path pointing to the test script file.</param>
+    /// <param name="workingDirectory">The path where the test script should be executed, will be converted to absolute.</param>
+    public static Task SwitchToInteractiveWithJavascriptTestInfoAsync(
```
Piedone: Very cool!

Lombiq.Tests.UI/ui-testing-toolkit.mjs
```diff
@@ -0,0 +1,132 @@
...
+ * @param {function(WebDriver, string):Promise<void>} test
+ * @param {function(chrome.Options):chrome.Options} configureOptions Update the configuration before the driver is built.
+ * @returns {Promise<void>} Success if the driver is created and the test has run to completion.
+ */
+async function runTest(test, configureOptions = null) {
+    const args = process.argv.slice(2);
+    if (args.length !== 3) throw new Error('Usage: node script.js driverPath startUrl tempDirectory');
+    const [driverPath, startUrl, tempDirectory] = args;
+
+    let options = new chrome.Options()
```
Piedone: Doesn't this mean that the JS test-related C# methods should throw an exception if the current browser is not Chrome, because this won't work?

Lombiq.Tests.UI/ui-testing-toolkit.mjs
```diff
@@ -0,0 +1,132 @@
...
+        .addArguments('--ignore-certificate-errors')
+        .addArguments('--no-sandbox')
+        ;
+
+    if (process.env.GITHUB_ENV) options = options.addArguments('headless');
+    if (configureOptions) options = configureOptions(options) ?? options;
+
+    const service = new chrome.ServiceBuilder(driverPath).build();
+    const driver = chrome.Driver.createSession(options, service);
+    await driver.manage().setTimeouts({ implicit: 10000 });
```
Piedone, line 104: This isn't an implicit wait, right?

Lombiq.Tests.UI/ui-testing-toolkit.mjs
```diff
@@ -0,0 +1,132 @@
...
+async function runTest(test, configureOptions = null) {
+    const args = process.argv.slice(2);
+    if (args.length !== 3) throw new Error('Usage: node script.js driverPath startUrl tempDirectory');
+    const [driverPath, startUrl, tempDirectory] = args;
+
+    let options = new chrome.Options()
+        .addArguments('disable-dev-shm-usage')
+        .addArguments('unsafely-disable-devtools-self-xss-warnings')
+        .addArguments('disable-search-engine-choice-screen')
+        .addArguments('--lang=en-US')
```
Piedone: This should rather come from the config, like it is in `WebDriverFactory`.

Lombiq.Tests.UI/ui-testing-toolkit.mjs
```diff
@@ -0,0 +1,132 @@
...
+        .addArguments('disable-accelerated-2d-canvas')
+        .addArguments('disable-gpu')
+        .addArguments('force-color-profile=sRGB')
+        .addArguments('force-device-scale-factor=1')
+        .addArguments('high-dpi-support=1')
+        .addArguments('disable-smooth-scrolling')
+        .addArguments('ignore-certificate-errors')
+        .addArguments('--ignore-certificate-errors')
+        .addArguments('--no-sandbox')
```
Piedone: Shouldn't these all be the same as in `WebDriverFactory`? If they should, please fix. And in any case, comment in both places if these two need to be kept in sync in some way.

Piedone (changes requested):

Lombiq.Tests.UI.Samples/Tests/FrontendTests.cs
sarahelsaig: I don't understand the question.

Lombiq.Tests.UI.Samples/Tests/test.mjs
sarahelsaig: I'm referencing the same file in the sandbox, so that would be confusing. Also I don't want more JS tests than just this one so it's pointless to specify the test name in the file name.

Lombiq.Tests.UI.Samples/Tests/JavascriptTests.cs
sarahelsaig: In OFFI I did that with an additional "setup and test" method that derived the path from the caller's namespace. But that kind of stuff is too opinionated and unnecessary complexity.

Lombiq.Tests.UI/Extensions/FrontendUITestContextExtensions.cs
sarahelsaig: That you can't use it in methods that take a tenant name.

Lombiq.Tests.UI/ui-testing-toolkit.mjs
sarahelsaig: No, because they can provide their own driver config function if they want to. But I've updated the code to throw about it more explicitly in JS.

Lombiq.Tests.UI/ui-testing-toolkit.mjs
sarahelsaig: It is an [implicit wait](https://www.selenium.dev/documentation/webdriver/waits/). Why?

Lombiq.Tests.UI/ui-testing-toolkit.mjs
sarahelsaig: I don't want to over-complicate it because every such config has to be passed into the script through command line arguments

Lombiq.Tests.UI.Samples/Tests/FrontendTests.cs
sarahelsaig: It looks like the shutdown code doesn't work on Windows only? Typical Windows.

I've added a fallback "forceful cancellation" to the stop code. I believe this uses SIGKILL, which is like ending the process in Task Manager. Please check if it works for you now.

Lombiq.Tests.UI.Samples/Tests/FrontendTests.cs
Piedone:
```suggestion
                // Don't forget that if you want to interact with the backend manually, you can just switch the context
                // back to the backend and use the interactive mode extension method.
```

Piedone: I believe that this should be https://github.com/Lombiq/UI-Testing-Toolbox/pull/421/files/bdb345aa3710e59b52ce149c161b1eff6fa36e1e#r1846935116, otherwise I don't understand.

Lombiq.Tests.UI.Samples/Tests/JavascriptTests.cs
Piedone: _Tests_ here refers to the build output directory, i.e. the test's execution directory, which is not necessarily the directory of the current/JS code file. How about while keeping the option to specify the working directory open, default that to the current directory, and recommend copying mjs files to the root of the output directory instead? Then, this could be:

```c#
return context.SetupSeleniumAndExecuteJavascriptTestAsync(_testOutputHelper, /*scriptName parameter*/"test.mjs");
```

Lombiq.Tests.UI/ui-testing-toolkit.mjs
Piedone: Hmm, I remembered that as a wait before every interaction (i.e. that it'll click slowly). It's not that, so OK then.

Lombiq.Tests.UI/ui-testing-toolkit.mjs
Piedone: Nothing else from here needs to be passed in, however, this one I think should: this also affects the `Accept-Language` header, thus the UI behavior. I.e. without settings this to the same as in C#, the C# and JS tests may see a different UI (even if they test different parts of the app, e.g. the same date for the same content item may be displayed differently).

Lombiq.Tests.UI.Samples/Tests/test.mjs
Piedone: I see. Then _JavaScriptTests.msj_, rather.

Lombiq.Tests.UI.Samples/Tests/FrontendTests.cs
Piedone: {97F4D6DA-7615-4763-A0E3-B8D5265718F6}

Works now, thanks.

Lombiq.Tests.UI.Samples/Tests/FrontendTests.cs
sarahelsaig: Nope. The point is that you can explore the frontend in interactive mode, but you can't just call `await context.SwitchToInteractiveAsync();`. You have to switch the UI test context scope back to the backend first (`context.SwitchToBackend();`) or it won't work, because the `await context.SwitchToInteractiveAsync();` will try to visit the page on the frontend's URL.

Piedone:
```suggestion
                // back to the backend and use the interactive mode extension method.
```

Piedone: I see. Can't `SwitchToInteractiveAsync` detect where it's called from and change the URL accordingly? Having to remember this is exposing an implementation particularity. Or at least an overload that makes this easier to discover, something along the lines of `SwitchToInteractiveModeFromFrontendWithSwitchToBackendAsync` (this is terribly long, but you get the idea).

sarahelsaig: It's not just the SwitchToInteractiveAsync, but every method that uses Lombiq.Tests.UI.Shortcuts (because those controller actions are in the backend). I don't think it's viable nor worth the extra complexity.

Piedone: How about checking `UITestContext.TenantName` for `FrontendPseudoTenantName` and throw in something like an `EnsureBackend()` in `ShortcutsUITestContextExtensions`?

Lombiq.Tests.UI.Samples/Tests/JavaScriptTests.cs
```diff
@@ -1,67 +1,50 @@
...
-            },
-            configuration =>
-            {
-                // Since this is an interactive "test", make sure the browser is always displayed.
-                configuration.BrowserConfiguration.Headless = false;
-                return Task.CompletedTask;
-            });
-    }
+    public Task Sandbox() =>
+        SandboxAfterSetupAsync(async context =>
```
Piedone:
```suggestion
        OpenSandboxAfterSetupAsync(async context =>
```

Lombiq.Tests.UI.Samples/Tests/JavaScriptTests.cs
```diff
@@ -1,67 +1,50 @@
...
         ExecuteTestAfterSetupAsync(context =>
         {
-            // Don't forget to mark the script files as "Copy if newer", so they are available to the test. It's best to
-            // include something like the following in your csproj file:
+            // Don't forget to mark the script files as "Copy if newer", so they are available to the test. If you
+            // include something like the following in your csproj file, then you only have to do this once:
             // <None Update="Tests\*.mjs" CopyToOutputDirectory="PreserveNewest" />
-            var workingDirectory = "Tests";
-            var scriptPath = Path.Join(workingDirectory, "test.mjs");
+            var scriptPath = Path.Join("Tests", "JavaScriptTests.mjs");
```
Piedone, line 29: Under https://github.com/Lombiq/UI-Testing-Toolbox/pull/421#discussion_r1846951255 I also suggested to recommend copying mjs files to the root of the output directory instead. That would make this easier simpler too, with just `"JavaScriptTests.mjs"` needed for the path.

Lombiq.Tests.UI/Services/FrontendServer.cs
```diff
@@ -124,11 +125,18 @@ public void Configure(
...
+                    forcefulCancellation.CancelAfter(TimeSpan.FromMinutes(1));
+
+                    // Using Task.WhenAny() without unwrapping its output avoids the OperationCanceledException, which
+                    // is not relevant in disposal code.
+                    await Task.WhenAny(cliTask);
+
+                    gracefulCancellation.Dispose();
+                    forcefulCancellation.Dispose();
 
                     await context.PortLeaseManager.StopLeaseAsync(frontendPort);
```
Piedone, line 141: Shouldn't these three lines be in a `finally`?

Piedone (changes requested):

Lombiq.Tests.UI.Samples/Tests/JavaScriptTests.cs
sarahelsaig: I would not recommend that. It's simpler in this specific example, but would be less organized in a real scenario. If you have JS tests in multiple test fixtures, then you would structure them like Tests/TestFixtureName/TestName.mjs anyway. Copying those to TestFixtureName/TestName.mjs would not be an improvement, while flattening the whole structure to unceremoniously dump every mjs file into the build root would make a mess.

Also without the walkthrough comments this can be compacted a lot anyway:
```csharp
    public Task ExampleJavaScriptTestShouldWork() =>
        ExecuteTestAfterSetupAsync(context => context.SetupSeleniumAndExecuteJavaScriptTestAsync(
            _testOutputHelper, Path.Join("Tests", "JavaScriptTests.mjs")));
```
In this situation, making the msbuild code less intuitive just to save one `Path.Join` seems pointless to me.

Lombiq.Tests.UI/Services/FrontendServer.cs
sarahelsaig: Why? Nothing is this `StopAsync` is expected to throw.

Piedone: `await Task.WhenAny(cliTask);` can, no?

Lombiq.Tests.UI.Samples/Tests/JavaScriptTests.cs
Piedone: Fair enough.

Lombiq.Tests.UI/Services/FrontendServer.cs
sarahelsaig: Nope, as explained by the comment above it. `await await Task.WhenAny(cliTask);` or `await Task.WhenAny(cliTask).Unwrap();` could throw. `await Task.WhenAny(cliTask);` just finishes with success when an inner task ends.

Piedone: OK then.

Piedone (approved):

PR #431, opened 2024-12-19
LMBQ-429: Adding the ability to maintain more than one Cloudflare IP Access Rule for remote tests

Piedone:
[LMBQ-429](https://lombiq.atlassian.net/browse/LMBQ-429)

[LMBQ-429]: https://lombiq.atlassian.net/browse/LMBQ-429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Helpers/CloudflareHelper.cs
```diff
@@ -111,35 +125,37 @@ await ReliabilityHelper.DoWithRetriesAndCatchesAsync(
...
+                if (deleteSucceededResult.IsSuccess) _ipAccessRuleIds.TryRemove(currentIp, out _);
 
-                ThrowIfNotSuccess(deleteSucceededResult, _currentIp, "couldn't be deleted");
+                ThrowIfNotSuccess(deleteSucceededResult, currentIp, "couldn't be deleted");
 
                 testOutputHelper.WriteLineTimestampedAndDebug(
                     "Removed the Cloudflare IP Access Rule for the IP {0} (Rule ID: {1}) since this test had the last reference to it.",
-                    _currentIp,
+                    currentIp,
                     oldIpAccessRuleId);
```
DemeSzabolcs: Is there a point in using `oldIpAccessRuleId` in one place but `value` in another? Since here `oldIpAccessRuleId` is made equal to `value`.

DemeSzabolcs (changes requested):

Lombiq.Tests.UI/Helpers/CloudflareHelper.cs
Piedone: Not really, this is just a leftover. Fixed it.

DemeSzabolcs (approved):

PR #432, opened 2024-12-22
OSOE-935: Upgrade to xUnit 3 and handling test cancellations gracefully

### LINKED ISSUE #82: Handle cancellations gracefully (OSOE-362)
[This xUnit bug](https://github.com/xunit/xunit/issues/2008 "https://github.com/xunit/xunit/issues/2008") was fixed for this but will only be part of xUnit v3. Once that's released, `TestContext` [will contain a `CancellationToken`](https://github.com/xunit/xunit/issues/345#issuecomment-1451388078) which is accessible as `TestContext.Current.CancellationToken`. The latter should be passed down so canceled tests won't e.g. leave browser windows or ChromeDriver processes open.

[Hooking into the console cancel event](https://github.com/xunit/xunit/issues/2008#issuecomment-588321441 "https://github.com/xunit/xunit/issues/2008#issuecomment-588321441") won't work with `dotnet test`. 

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-362)

### LINKED ISSUE #430: Upgrade to xUnit 3 (OSOE-935)
See https://xunit.net/docs/getting-started/v3/whats-new and https://xunit.net/docs/getting-started/v3/migration.

https://github.com/OrchardCMS/OrchardCore/pull/17250 is about doing this for OC.

https://github.com/Lombiq/UI-Testing-Toolbox/issues/82 can be started once this is done.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-935)

Piedone:
[OSOE-935](https://lombiq.atlassian.net/browse/OSOE-935)
Fixes #430
Fixes #82

Test cancellation can be tried when running the exe of a test: run it from the CLI, wait like 10 seconds, hit Ctrl+C (only once), and you should see "Test execution was cancelled. Shutting down the test execution session." at one point (only once the tests actually started).

To be added to the release notes:

## Upgraded xUnit to v3 with breaking changes

This version of the UI Testing Toolbox uses v3 of xUnit, which brings [many updates](https://xunit.net/docs/getting-started/v3/whats-new). However, it's also a breaking version, requiring you to adapt your test projects, see [the official guide](https://xunit.net/docs/getting-started/v3/migration).

Migrating UI test projects consuming the UI Testing Toolbox should be simpler, though, with you requiring to do roughly the following steps:

1. Update the `xunit.runner.visualstudio` package reference to latest (currently 3.0.2).
2. Convert test projects to executables, [see docs](https://xunit.net/docs/getting-started/v3/migration#migrating-to-v3-packages:~:text=and%20your%20users.-,Convert%20to%20Executable%20Project,-Update%20your%20project) (this will most possibly only need you to add `<OutputType>Exe</OutputType>` to the first `PropertyGroup` in the test project's csproj). These are all the projects with `xunit.runner.visualstudio` references.
3. In all test projects, change `xunit` package references to `xunit.v3` with the latest version (currently 2.0.0). If a test project lacks such a reference then add it (currently `<PackageReference Include="xunit.v3" Version="2.0.0" />`). Only test projects should reference `xunit`, but they must reference it, otherwise tests won't be discovered.
4. Also update the `Microsoft.NET.Test.Sdk` references in the test projects to latest (currently 17.13.0) too while we're at it, or add it if it's missing (currently `<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.13.0" />`).
5. `ITestOutputHelper` is now in the `Xunit` namespace instead of `Xunit.Abstractions`. Fix the namespace imports in `UITestBase` classes first and then anywhere else; they'll show up as build errors.
6. Clean up unused namespace imports. In Visual Studio, you can do this by right-clicking on the solution -> Analyze and Code Cleanup -> Run Code Cleanup (Profile 1).
7. Run the tests to check if everything is still OK. If tests don't show up in the Visual Studio Test Explorer, or show up but don't start if you try to run them, check out the Tests pane of the Output Window for clues. Confirm that the Web project doesn't contain any xUnit references (even transitively, like by accidentally referencing UI test projects) but all test projects reference `xunit.v3`. You might need to do a recursive git clean to be sure to start with a clean slate.
8. Verify in CI that the tests are actually discovered and all run there too (the CI build being green is not enough, it might miss no tests being discovered).

[OSOE-935]: https://lombiq.atlassian.net/browse/OSOE-935?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI.Samples/Tests/RemoteTests.cs
```diff
@@ -13,13 +12,26 @@ namespace Lombiq.Tests.UI.Samples.Tests;
...
 // Toolbox also supports this. Check out the example below!
 
 // Note how the test derives from RemoteUITestBase this time, not UITestBase. This is a generic base class for any
 // remote test. However, if you're testing an app behind Cloudflare, you might see requests rejected with an HTTP 403
 // due to Cloudflare's Bot Fight Mode; in that case, derive from CloudflareRemoteUITestBase instead after checking out
 // its documentation on how to set it up.
+
+// While there may be some common code between local and remote tests, they cannot be the same, and the common code must
+// be in a project independent of the web app (which is not the case here for the sake of simplicity):
+// - In local tests, you have a much lower-level access to the app: E.g., you can access the Orchard Core log, use
```
dministro, line 26: Did you mean "higher-level"?

```suggestion
// - In local tests, you have much higher-level access to the app: E.g., you can access the Orchard Core log, use
```

Lombiq.Tests.UI/BasicOrchardFeaturesTesting/WorkflowsFeatureTestingUITestContextExtensions.cs
```diff
@@ -48,6 +48,8 @@ public static Task TestWorkflowsAsync(this UITestContext context) =>
                     By.XPath("//div[@class = 'jtk-endpoint jtk-endpoint-anchor jtk-draggable jtk-droppable']"), // #spell-check-ignore-line
                     By.XPath(taskXPath));
 
+                context.WaitElementToNotChange(By.ClassName("jtk-connector")); // #spell-check-ignore-line
```
dministro, line 51: Why this change is necessary?

Lombiq.Tests.UI/Models/UITestManifest.cs
```diff
@@ -11,32 +11,7 @@ namespace Lombiq.Tests.UI.Models;
 /// </summary>
 public class UITestManifest
 {
-    public ITest XunitTest { get; }
-    public string Name => XunitTest?.DisplayName;
+    public ITest XunitTest => TestContext.Current.Test;
+    public string Name => XunitTest?.TestDisplayName;
```
dministro: Can `TestContext.Current.Test` `null` here?

Lombiq.Tests.UI/Models/UITestManifest.cs
```diff
@@ -11,32 +11,7 @@ namespace Lombiq.Tests.UI.Models;
...
     public Func<UITestContext, Task> TestAsync { get; set; }
```
dministro: This can be a constructor parameter.
image

Lombiq.Tests.UI/OrchardCoreUITestBase.cs
```diff
@@ -369,7 +369,7 @@ protected virtual async Task ExecuteTestAsync(
         Func<UITestContext, Task<Uri>> setupOperation,
         Func<OrchardCoreUITestExecutorConfiguration, Task> changeConfigurationAsync)
     {
-        var testManifest = new UITestManifest(_testOutputHelper) { TestAsync = testAsync };
+        var testManifest = new UITestManifest { TestAsync = testAsync };
```
dministro: See my comment for `UITestManifest.TestAsync`.

Lombiq.Tests.UI/RemoteUITestBase.cs
```diff
@@ -61,7 +61,7 @@ async Task BaseUriVisitingTest(UITestContext context)
             await testAsync(context);
         }
 
-        var testManifest = new UITestManifest(_testOutputHelper) { TestAsync = BaseUriVisitingTest };
+        var testManifest = new UITestManifest { TestAsync = BaseUriVisitingTest };
```
dministro: See my comment for `UITestManifest.TestAsync`.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -61,6 +61,12 @@ public UITestExecutionSession(
 
     public async Task<bool> ExecuteAsync(int retryCount, string dumpRootPath)
     {
+        _configuration.TestCancellationToken.Register(() =>
```
dministro: I'm not sure, but I think `System.Threading.CancellationTokenRegistration` returned by `CancellationToken.Register` needs to be disposed.

Is this really necessary?

Testing `_configuration.TestCancellationToken.IsCancellationRequested` also wellcome before starting the whole ceremony.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -98,7 +104,30 @@ public async Task<bool> ExecuteAsync(int retryCount, string dumpRootPath)
...
+                    throw new TimeoutException($"The time allotted for the test ({timeout}) was exceeded.");
+                }
+
+                await testTask;
+            }
+
+            // Since the timeout task is not yet completed but the Task.WhenAny has finished, the test task is done in
+            // some way. So it's safe to await it here. It's also necessary to cleanly propagate any exceptions that may
+            // have been thrown inside it.
+            await testTask;
```
dministro: In case when a cancellation is requested on the `_configuration.TestCancellationToken`, the `timeoutTask.IsCompleted` will be `true`.

Lombiq.Tests.UI/Services/UITestExecutor.cs
```diff
@@ -75,11 +60,6 @@ private static async Task ExecuteOrchardCoreTestInnerAsync(
         {
```
dministro, line 60: The `_configuration.TestCancellationToken.IsCancellationRequested` can also stop the loop.

dministro (changes requested):

Lombiq.Tests.UI.Samples/Tests/RemoteTests.cs
Piedone: I did mean lower level: the app runs in the same process, so you can communicate with its container and types directly, whereas in prod, all you have is the public UI.

Lombiq.Tests.UI/BasicOrchardFeaturesTesting/WorkflowsFeatureTestingUITestContextExtensions.cs
Piedone: This waits for the connector to settle after the drag and drop. The test was flaky without this, since the connection might not be registered and thus saved before it.

Lombiq.Tests.UI/Models/UITestManifest.cs
Piedone: Not, actually.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
Piedone: It indeed needs to be disposed, so I added that.

I don't really understand the rest of your comment: xUnit 3 supports cancellations. If we want the tests to clean up after themselves when such a cancellation happens, hooking into it is necessary. Running `ShutdownAsync()` will do the clean-up. Thus after this, a cancelled test won't leave ChromeDriver processes or browser windows open.

Lombiq.Tests.UI.Samples/Tests/InteractiveModeTests.cs
```diff
@@ -34,57 +31,19 @@ public Task SampleTest() =>
...
-                                // The other thread will wait indefinitely if the button wasn't clicked in the end. So,
-                                // failing the test then.
-                                TestContext.Current.CancelCurrentTest();
-                            }
-                        },
-                        context.Configuration.TestCancellationToken));
-
-                // Ensure that the info tab is closed and the control handed back to the last tab.
-                context.Driver.Url.ShouldBe(currentUrl);
-            });
```
Piedone: I've spent half a day trying to make this reliable, i.e. not hang randomly (that hung the tests too), and be reliable with its interactions. I couldn't get there, it was always flaky. I don't see any point in trying harder, we don't really need this test.

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -98,7 +105,30 @@ public async Task<bool> ExecuteAsync(int retryCount, string dumpRootPath)
...
```
dministro: I think we dont need to `await` twice
```suggestion
            }

            // Since the timeout task is not yet completed but the Task.WhenAny has finished, the test task is done in
            // some way. So it's safe to await it here. It's also necessary to cleanly propagate any exceptions that may
            // have been thrown inside it.
            await testTask;
```

dministro (changes requested):

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
Piedone: Indeed.

dministro (approved):

PR #440, opened 2024-12-29
Updating all dependencies, fixing SMTP tests, changing browser logging to BiDi

Piedone:
[OSOE-815](https://lombiq.atlassian.net/browse/OSOE-815)

**To be included in the release notes:**

## Breaking changes in browser log management

The browser log, i.e. what you also see in the developer console, is useful to pinpoint unhandled JavaScript exceptions, network errors, and other client-side issues. This is what in Selenium can be accessed via `IWebDriver.Manage().Logs.GetLog(LogType.Browser)`. In the UI Testing Toolbox it was accessible via various other ways, and was also automatically asserted after every page load.

In this release, how you can access the browser log, and what objects represent the log, changed. This was necessary because the previous version actually only supported Chrome and Edge, it just silently failed under other browsers (notably Firefox). With a new Selenium version, it now fails with an exception (that's why we actually noticed it).

To overcome this, all browsers now use the new Selenium [BiDirectional API](https://www.selenium.dev/documentation/webdriver/bidi/) to access the logs.

### What changed

- `UITestContext.HistoricBrowserLog` is now called `CumulativeBrowserLog` to better reflect that it stores all browser log entries since the start of the test. Instead of containing `OpenQA.Selenium.LogEntry` objects, it contains `OpenQA.Selenium.BiDi.Modules.Log.Entry` ones. These have a slightly different interface, but work much the same: Most possibly, you'll only need to change references to `LogEntry.Message` to `Entry.Text`, and `LogEntry.Level` to `Entry.Level`, which uses a different, but very similar, enum. For the same reason, `ClearHistoricBrowserLog()` is now called `ClearCumulativeBrowserLog()` but behaves the same.
- Similar to the response log, see below, the browser log now has a filter, `UITestContext.BrowserLogFilter`, to determine what gets logged in the first place. If you need to ignore expected log entries (since a non-empty log fails the test by default), we recommend adjusting the filter instead of the assertion. The samples contain examples of this.
- `OrchardCoreUITestExecutorConfiguration.AssertBrowserLog` now works with `OpenQA.Selenium.BiDi.Modules.Log.Entry` objects too, as do all related helpers.
- `UITestContext.UpdateHistoricBrowserLogAsync()` was removed. It's not necessary anymore since the logs are now updated not explicitly, but with events, whenever a new entry is added.
- `UITestContext.AssertCurrentBrowserLogAsync()` and the `IWebDriver` extension `GetAndEmptyBrowserLog()` were also removed for the same reason.
- Unexpected response codes, like 404 responses, aren't logged to the browser log anymore. Thus, `LogEntry.IsNotFoundLogEntry()` was removed since it can't apply to `Entry` objects. Instead, check out `UITestContext.CumulativeResponseLog` and `UITestContext.ResponseLogFilter`. This new behavior is demonstrated in the samples.

### How to adapt

- The most important issues will show up as build errors. Fix them following the guidelines above.
- Check if you have assertion code for the browser log containing 404 or other network errors. These won't show up in the browser log anymore, so you'll have to adjust your assertions to use the response log, see above.
- Running your test suite will make any further issues apparent.
- We also encourage you to explore the BiDi APIs, since they provide functionality not seen in Selenium before:
```c#
var biDi = await context.Scope.Driver.AsBiDiAsync();
// Check out biDi.Network, for example.
```

[OSOE-815]: https://lombiq.atlassian.net/browse/OSOE-815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Piedone:
Will open a new PR from `issue/OSOE-815-combined`.

PR #463, opened 2025-02-23
LMBQ-440: Using IP ranges for allowing remote tests of apps behind Cloudflare

Piedone:
[LMBQ-440](https://lombiq.atlassian.net/browse/LMBQ-440)

[LMBQ-440]: https://lombiq.atlassian.net/browse/LMBQ-440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/Helpers/CloudflareHelper.cs
+++ b/Lombiq.Tests.UI/Helpers/CloudflareHelper.cs
@@ -203,7 +203,7 @@ internal static class CloudflareHelper
 
         throw new IOException(
             $"The Cloudflare IP Access Rule for allowing requests from this runner {messagePart}. There might be a " +
-            $"leftover rule for the IP range {currentIp} that needs to be deleted manually." +
+            $"leftover rule for the IP range {currentIpRange} that needs to be deleted manually." +
             (result.InnerException is ApiException ex ? $" Response: {ex.Content}" : string.Empty),
             result.InnerException);
     }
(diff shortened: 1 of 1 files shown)


Lombiq.Tests.UI/Helpers/CloudflareHelper.cs
```diff
@@ -193,7 +203,7 @@ private static void ThrowIfNotSuccess((bool IsSuccess, Exception InnerException)
 
         throw new IOException(
             $"The Cloudflare IP Access Rule for allowing requests from this runner {messagePart}. There might be a " +
-            $"leftover rule for the IP {currentIp} that needs to be deleted manually." +
+            $"leftover rule for the IP range {currentIp} that needs to be deleted manually." +
```
DemeSzabolcs: Shouldn't this be?
```suggestion
            $"leftover rule for the IP range {currentIpRange} that needs to be deleted manually." +
```

DemeSzabolcs (changes requested):

Lombiq.Tests.UI/Helpers/CloudflareHelper.cs
Piedone: Yep!

PR #464, opened 2025-02-25
OFFI-122: Adding many useful helper functions and adding SMTP client url to frontend

wAsnk:
[OFFI-122](https://lombiq.atlassian.net/browse/OFFI-122)

[OFFI-122]: https://lombiq.atlassian.net/browse/OFFI-122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/ui-testing-toolkit.mjs
```diff
@@ -1,40 +1,284 @@
...
+/**
+ * Waits for the URL to be the provided value.
+ * @param driver {WebDriver} - The WebDriver instance.
+ * @param url {string} - The URL to wait for.
+ * @returns {Promise<*>}
+ */
+async function waitForUrl(driver, url)
+{
+    return safeWait(driver, until.urlIs(url));
+}
```
sarahelsaig: Please include the `timeout` parameter here as well. It could be necessary in case of waiting for an async call that leads to a page change after taking a long time.

Lombiq.Tests.UI/ui-testing-toolkit.mjs
```diff
@@ -1,40 +1,284 @@
...
+ */
+async function waitUntilInputValueIsNotEmpty(driver, element, timeout = 5000) {
+    await safeWait(
+        driver,
+        async function () {
+            const currentValue = await element.getAttribute('value');
+            return currentValue?.length > 0;
+        },
+        timeout);
+}
```
sarahelsaig: If `await` is your last call then don't do `async`, just return the promise (`return safeWait(...)`). Same below.

Lombiq.Tests.UI/ui-testing-toolkit.mjs
```diff
@@ -1,40 +1,284 @@
...
+/**
+ * Finds an element using the provided selector.
+ * @param driver {WebDriver} - The WebDriver instance.
+ * @param by {By} - The By selector used to find the element.
+ * @returns {Promise<*>}
+ */
+async function findElementBy(driver, by)
+{
+    return driver.findElement(by);
+}
```
sarahelsaig: Remove that `async`, you don't await anything in the method. Same in `findElementByXpath` and `findElementByInputValue`.

Lombiq.Tests.UI/ui-testing-toolkit.mjs
```diff
@@ -1,40 +1,284 @@
...
+
+/**
+ * Finds an input element by its value.
+ * @param driver {WebDriver} - The WebDriver instance.
+ * @param value {string} - The value of the input element to find.
+ * @returns {Promise<*>}
+ */
+async function findElementByInputValue(driver, value)
+{
+    return findElementByXpath(driver, `//input[@value=\'${value}\']`);
```
sarahelsaig: Use `` `//input[@value=${JSON.stringify(value)}]` `` so this won't break if the value has an apostrophe in it.

Lombiq.Tests.UI/ui-testing-toolkit.mjs
```diff
@@ -1,40 +1,284 @@
...
+}
+
+/**
+ * Creates a By selector for finding an element by its tag and containing text.
+ * @param tag The tag of the element.
+ * @param searchAttribute The text to search for.
+ * @returns {!By} The By selector.
+ */
+function byTagAndText(tag, searchAttribute){
+    return By.xpath(`//${tag}[contains(.,"${searchAttribute}")]`);
```
sarahelsaig: Use `` `//${tag}[contains(., ${JSON.stringify(searchAttribute)})]` `` so this won't break if the text has a quote in it. Same in `byInputByLabel`.

Lombiq.Tests.UI/ui-testing-toolkit.mjs
```diff
@@ -1,40 +1,284 @@
...
+async function linkShouldOpenNewTab(driver, element)
+{
+    const targetAttr = await element.getAttribute("target");
+    return targetAttr === "_blank";
+}
+
+/**
+ * Creates a By selector for finding an element by its tag and containing text.
+ * @param tag The tag of the element.
+ * @param searchAttribute The text to search for.
```
sarahelsaig: Why is this parameter called `searchAttribute`? As I understand it's not referring to an attribute. Maybe it should be `text`? Same in `findElementByTagAndContainingText`.

Lombiq.Tests.UI/ui-testing-toolkit.mjs
```diff
@@ -1,40 +1,284 @@
...
+
+/**
+ * Checks if the provided link should open in a new tab by its target attribute.
+ * @param driver {WebDriver} - The WebDriver instance.
+ * @param element {WebElement} - The link element to check.
+ * @returns {Promise<boolean>} - True if the link should open in a new tab, false otherwise.
+ */
+async function linkShouldOpenNewTab(driver, element)
+{
+    const targetAttr = await element.getAttribute("target");
```
sarahelsaig: Use apostrophes instead of quotes consistently.

sarahelsaig (changes requested):

Lombiq.Tests.UI/ui-testing-toolkit.mjs
wAsnk: driver.findElement should be awaited, shouldn't it?

sarahelsaig: Nope. You already return the `Promise` result of `driver.findElement`, so adding `async` will only slow down the JS interpreter. (Technically it won't cause problems because in JS a promise-returning-promise automatically unwraps itself into a simple promise, but it's inefficient and less readable.) This is just like returning `Task` in C#, if you only have one asynchronous call at the end of the function then you can return it directly without the `async`/`await`.

wAsnk: Okay, done then.

PR #466, opened 2025-02-28
OCC-310: Fetch the log contents but only include the entries that will throw

sarahelsaig:
[OCC-310](https://lombiq.atlassian.net/browse/OCC-310)

[OCC-310]: https://lombiq.atlassian.net/browse/OCC-310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Models/MemoryApplicationLog.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Lombiq.Tests.UI.Models;
+
+/// <summary>
+/// An <see cref="IApplicationLog"/> implementation where the entries are stored in memory.
+/// </summary>
+public record MemoryApplicationLog(string Name, IList<IApplicationLogEntry> Entries)
```
Piedone, line 13: Why is this necessary, exactly? Is multiple `await`s really that costly to warrant copying the log into this type? Because keep in mind, that by default, the logs will be in-memory already with `FakeLogger` (see from `FakeLoggerLogApplicationLog`).

Piedone (changes requested):

Lombiq.Tests.UI/Models/MemoryApplicationLog.cs
sarahelsaig: We are [converting the whole log content to string](https://github.com/Lombiq/UI-Testing-Toolbox/blob/5657458855d2b2255db4d95b250cfff2b39b5fc3/Lombiq.Tests.UI/Extensions/WebApplicationInstanceExtensions.cs#L170) every time the logs are asserted anyway, so I don't think that matters. On the other hand this makes the code easier to debug, so overall I think it's worth the extra copy. When the source is already in-memory as with FakeLogger, then it's fine either way.

Lombiq.Tests.UI/Extensions/ApplicationLogEnumerableExtensions.cs
```diff
@@ -8,23 +9,27 @@ namespace Lombiq.Tests.UI.Extensions;
...
         if (logsArray.Length == 1)
         {
-            return Environment.NewLine + await LogLinesToFormattedStringAsync(logsArray[0]);
+            return Environment.NewLine + logsArray[0].ToFormattedString();
         }
 
-        // Parallelization with Task.WhenAll() isn't really necessary for performance here but would potentially change
-        // the order of the logs in the output.
-        var logContents = logsArray.AwaitEachAsync(async log =>
-            $"# Log name: {log.Name}" + Environment.NewLine + Environment.NewLine + await LogLinesToFormattedStringAsync(log));
```
sarahelsaig, line 23: Note that this original code was missing an `await`, so I don't think it ever worked when there were multiple log sources to list, meaning that `logContents` was `Task<IList<string>>` instead of `IList<string>`.

Lombiq.Tests.UI/Models/MemoryApplicationLog.cs
Piedone: How does it make easier to debug?

sarahelsaig: Since the logs is converted into this type, you can just explore the  `logs` in `AssertLogsAsync`, and easily do breakpoint conditions since the log entries are synchronously accessible.

Also going back to the fact that we filter the logs in `MemoryApplicationLog.FromLogAsync` so the Shouldly user message will only contain the the offending log entries that cause the assertion failure. This is especially relevant for tests with a lot of filtered out logs, such is the case with security tests. If we already load in the logs to make that filtering, why not keep the results so we don't have to do it again to generate the formatted string?

Piedone: OK, let's keep it, but please add the rationale to the class docs, because currently it seems that it's for perf.

> the Shouldly user message will only contain the the offending log entries that cause the assertion failure.

Yes, and that I'm definitely for, but it doesn't necessarily need this type.

sarahelsaig: > Yes, and that I'm definitely for, but it doesn't necessarily need this type.

I wouldn't say "need", but it makes things more organized and easier to reason about.

Piedone (approved):

PR #473, opened 2025-03-14
OSOE-1057: Reliability improvements necessary after Chrome 134.0.6998.89

Piedone:
[OSOE-1057](https://lombiq.atlassian.net/browse/OSOE-1057)

[OSOE-1057]: https://lombiq.atlassian.net/browse/OSOE-1057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Extensions/ExtendedLoggingExtensions.cs
Piedone, line 1: For why I removed the retries here see: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/issues/991#issuecomment-2722367087

wAsnk (approved):

Lombiq.Tests.UI/Extensions/EmailUITestContextExtensions.cs
```diff
@@ -27,20 +27,10 @@ public static async Task GoToSmtpWebUIAsync(this UITestContext context)
...
-        {
-            context
-                .Scope.AtataContext.Log
-                .Info($"The smtp4dev site didn't reload, so the the missing loading element was ignored: " +
-                    $"{exception.Message}");
-        }
+        // We are waiting for this exact element to appear, only with one class, that indicates that the loading is
+        // happening. The loading is not always happening that's why we do it with Safely(), making sure that the
+        // element either did not exist, or we waited for it to appear.
+        context.Get(By.XPath($"//div[@class='{LoadingMaskClass}']").Safely());
```
DemeSzabolcs, line 33: I also had problems here again, just for reference this is what I did: https://github.com/Lombiq/UI-Testing-Toolbox/blob/796b38f849e9da7dbca113d32c00561b5c31f9f7/Lombiq.Tests.UI/Extensions/EmailUITestContextExtensions.cs#L34-L56

Your solution seems nicer though.

Piedone: Yeah, we don't need to catch exceptions. Also, note that `Get` and friends already do retries for existence, so you don't need to retry for that outside of them.

Lombiq.Tests.UI.Samples/Tests/MonkeyTests.cs
```diff
@@ -60,35 +60,21 @@ public Task TestAdminPagesAsMonkeyRecursivelyShouldWorkWithAdminUser() =>
...
 
-                return context.TestAdminAsMonkeyRecursivelyAsync(monkeyTestingOptions);
-            },
-            // Requests to /api/graphql without further parameters will fail with HTTP 400, but that's OK, since some
-            // parameters are required.
-            configuration => configuration.ResponseLogFilter = e => e.IsNonSuccessResponseAndNotExpectedStatusResponse("/api/graphql", 400));
-
-    // Let's just test the background tasks management admin area.
-    [Fact]
-    public Task TestAdminBackgroundTasksAsMonkeyRecursivelyShouldWorkWithAdminUser() =>
```
Piedone, line 71: This test wasn't testing or showcasing anything new, but could try to save invalid cron settings, causing OC exceptions. So, incorporated the educational part into the test above and removed it.

Lombiq.Tests.UI/Extensions/ExtendedLoggingExtensions.cs
Piedone: I re-added the `StaleElementReferenceException` retries to more appropriate places, everywhere where these methods were used and where it was applicable.

DemeSzabolcs (approved):

PR #500, opened 2025-03-28
OCC-333: Fix randomly broken BasicOrchardFeaturesTests

sarahelsaig:
[OCC-333](https://lombiq.atlassian.net/browse/OCC-333)

[OCC-333]: https://lombiq.atlassian.net/browse/OCC-333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs
@@ -291,7 +291,7 @@ public static class BasicFeaturesTestingUITestContextExtensions
     /// <param name="setupParameters">The setup parameters.</param>
     /// <returns>The same <see cref="UITestContext"/> instance.</returns>
     public static Task TestSetupAsync(this UITestContext context, OrchardCoreSetupParameters setupParameters = null) =>
-        context.TestSetupAsync(setupParameters, "Test setup");
+        context.TestSetupAsync(setupParameters, "Test setup", shouldBeSuccess: true);
 
     /// <summary>
     /// <para>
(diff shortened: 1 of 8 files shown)


Lombiq.Tests.UI/BasicOrchardFeaturesTesting/WorkflowsFeatureTestingUITestContextExtensions.cs
```diff
@@ -62,7 +62,7 @@ public static Task TestWorkflowsAsync(this UITestContext context) =>
                 // We need to save the workflow early, because sometimes the editor, thus the startup task button can be
                 // buggy during UI testing (it won't be clicked, even if we check for its existence). This way it's
                 // always clicked.
-                await context.ClickReliablyOnSubmitAsync();
```
DemeSzabolcs, line 65: I know this is not strongly related, but this was also changed recently: https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/Pages/OrchardCoreLoginPage.cs#L70

Causing some tests to fail in other projects. The button can be "Sign in" on some websites, so it would be better to have a generic implementation like before.

If you can fix that here (as a scope change ofc) I would appreciate that, but if not, no problem. I thought I would mention it since there is already a UITT PR.

DemeSzabolcs: See these also: https://github.com/Lombiq/UI-Testing-Toolbox/pull/495#discussion_r2020054194

sarahelsaig: In the end I put the login text into `OrchardCoreSetupParameters.LoginButtonText` property so it's customizable at call. I didn't want to do static field Zoli suggested, because that will 110% will cause problems in DotNest or some other multi-tenant app sooner or later.

Since I was already there, I've refactored `BasicFeaturesTestingUITestContextExtensions` to use this existing configuration object instead of a bunch of separate `TestBasicOrchardFeaturesExcept*Async` methods. Those I have marked Obsolete, so in the future we will only need `TestBasicOrchardFeaturesAsync` and everything else is configured centrally to make it more maintainable and DRY.

DemeSzabolcs (approved):

Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs
```diff
@@ -330,20 +399,15 @@ public static Task TestLoginAsync(
...
-
-                (await context.GetCurrentUserNameAsync()).ShouldBeEmpty();
-            });
+            userName,
+            password + "WrongPass!",
+            signOut: true,
+            shouldBeSuccess: false,
+            logInButtonText);
 
     /// <summary>
```
DemeSzabolcs, line 412: I just realized that in the new code, on line 468: https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs#L468

The new login button text is not applied, so it fails if you test registration.

PR #501, opened 2025-04-01
OFFI-277: Add feature for Elasticsearch parallel use

sarahelsaig:
[OFFI-277](https://lombiq.atlassian.net/browse/OFFI-277)

[OFFI-277]: https://lombiq.atlassian.net/browse/OFFI-277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
```diff
@@ -0,0 +1,89 @@
...
+namespace Lombiq.Tests.UI.Models;
+
+public record ElasticsearchRunningContext(Guid Id, string Prefix)
+{
+    private IndexName IndexName => Indices.Index($"{Prefix}_*");
+
+    public Task BeforeTestAsync(UITestContext context) =>
+        context.Application.UsingScopeAsync(async provider =>
+        {
+            var index = IndexName;
```
DemeSzabolcs: Something is not right here, I think it's just that the naming is confusing. I implemented: https://lombiq.atlassian.net/browse/LMBQ-462 here.

Using `IndexName.Name` in the `await elasticIndexManager.GetLastTaskId()` function (which requires an index name) throws an exception, since it can't find the index.

So I had to use: 
```cs
 var indexSettings = await settingsService.GetSettingsAsync();
 var exactIndexName = indexSettings.FirstOrDefault()?.IndexName;
```
Which results in `search-{tenant-name}`. I don't know what is wrong here, my first thought is that you are using a wildcard index name, instead of an exact one, if I see it correctly. Then I think it should be renamed to `WildcardIndexName` or something like that. Another possibility is what I've done, which is to name my variable `exactIndexName`.

But I might be wrong, and I also need to put the prefix before the exactIndex name.

See for context: 

```cs
    public Task BeforeTestAsync(UITestContext context) =>
        context.Application.UsingScopeAsync(async provider =>
        {
            var index = IndexName;
            var cancellation = context.Configuration.TestCancellationToken;

            if (GetClient(provider) is not { } client)
            {
                throw new InvalidOperandException(
                    $"Couldn't resolve {nameof(IElasticClient)} while waiting for \"{index}\".");
            }

            (await client.Indices.FlushAsync(index, ct: cancellation)).ThrowIfFailed($"flush index \"{index}\"");
            (await client.Indices.RefreshAsync(index, ct: cancellation)).ThrowIfFailed($"refresh index \"{index}\"");

            // Elasticserch indexing sometimes takes longer, and the testing starts before it finishes. To prevent that,
            // we are checking if all of the indexing tasks are finished.
            var elasticIndexManager = provider.GetRequiredService<ElasticIndexManager>();
            var settingsService = provider.GetRequiredService<ElasticIndexSettingsService>();
            var indexingTaskManager = provider.GetRequiredService<IIndexingTaskManager>();
            var indexSettings = await settingsService.GetSettingsAsync();
            var exactIndexName = indexSettings.FirstOrDefault()?.IndexName;

            var startIndex = 0;
            const int batchSize = 1000;
            long lastTaskId = 0;
            var taskCount = 0;

            // We are getting the last indexing task (regardless of the state). This function works like a cursor, so
            // there is no way to get directly the last task in the list. Since we have to give a "count" parameter, we
            // are retrieving the indexing tasks by batches of 1000. Then if there is no more, we get the last one.
            do
            {
                var tasks = (await indexingTaskManager.GetIndexingTasksAsync(startIndex, batchSize)).ToList();
                taskCount = tasks.Count;

                if (taskCount > 0)
                {
                    lastTaskId = tasks[^1].Id;
                    startIndex += batchSize;
                }
            }
            while (taskCount > 0);

            long? lastFinishedTaskId = null;

            // We have the id of the last indexing task that should happen, so we are waiting here for that task to
            // complete, since "GetLastTaskId()" returns only completed tasks.
            while (lastTaskId != lastFinishedTaskId)
            {
                lastFinishedTaskId = await elasticIndexManager.GetLastTaskId(exactIndexName);

                // The indexing takes a couple of seconds, so there is no need to check them so fast: we are adding a
                // delay.
                await Task.Delay(500, cancellation);
            }
        });
```

Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
```diff
@@ -0,0 +1,89 @@
...
+            var cancellation = context.Configuration.TestCancellationToken;
+
+            if (GetClient(provider) is not { } client)
+            {
+                throw new InvalidOperandException(
+                    $"Couldn't resolve {nameof(IElasticClient)} while waiting for \"{index}\".");
+            }
```
DemeSzabolcs: Can't we DRY this? It's also on line 66.

Lombiq.Tests.UI/Services/OrchardCoreUITestExecutorConfiguration.cs
```diff
@@ -186,6 +186,13 @@ public Func<LogEntry, bool> BrowserLogFilter
     /// </summary>
     public bool UseAzureBlobStorage { get; set; }
 
+    /// <summary>
+    /// Gets or sets a value indicating whether Elasticsearch will be used. When enabled, the test will assign a unique
+    /// prefix to the test, ensure that Elasticserach caches are flushed before starting the test, and clean up the
```
DemeSzabolcs: Shouldn't this say: "the test will assign a unique prefix to the index", instead of "the test will assign a unique prefix to the test"?

DemeSzabolcs (changes requested):

Lombiq.Tests.UI/Services/OrchardCoreUITestExecutorConfiguration.cs
sarahelsaig: Sure.
```suggestion
    /// Gets or sets a value indicating whether Elasticsearch will be used. When enabled, the test will assign a unique
    /// prefix to the INDEX, ensure that Elasticserach caches are flushed before starting the test, and clean up the
```

Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
sarahelsaig: No, the exception text is meaningfully different.

DemeSzabolcs: Ah ok!

Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
sarahelsaig: The `IndexName` here is `$"{Prefix}_*"`, so all indexes that start with the `ElasticsearchRunningContext.Prefix`. This is only to be used with `Nest.IElasticClient` which provides low-level access the the Elasticsearch API.

On the other hand, OC services like `OrchardCore.Search.Elasticsearch.Core.Services.ElasticIndexManager` don't need this, because they implicitly add the prefix from configuration (and also the tenant ID if not on the default tenant) to the received name using `ElasticIndexManager.GetFullIndexName(indexName)`. This applies to `elasticIndexManager.GetLastTaskId` too, you don't need to prefix it explicitly. So if you pass this `IndexName` into `ElasticIndexManager`, then the name will be double-prefixed by the time it's sent to the API.
In other words, on the default tenant `IndexName.Name == elasticIndexManager.GetFullIndexName("*")`.

I've renamed the prop to `LowLevelIndexName` and added doc comment to explain its use.

DemeSzabolcs: Thanks for the clarification! Great then!

DemeSzabolcs (approved):

PR #506, opened 2025-04-04
NEST-498: Adapting SMTP test setup to be compatible with multiple providers

BenedekFarkas:
[NEST-498](https://lombiq.atlassian.net/browse/NEST-498)

[NEST-498]: https://lombiq.atlassian.net/browse/NEST-498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/Extensions/EmailUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/EmailUITestContextExtensions.cs
@@ -154,6 +154,11 @@ public static class EmailUITestContextExtensions
         {
             await context.ClickReliablyOnAsync(By.ClassName("save"));
             context.Get(By.ClassName("validation-summary-errors").Safely())?.Text?.Trim().ShouldBeNullOrEmpty();
+
+            // "SMTP" is wired in so that we don't have to reference the 'OrchardCore.Email.Smtp' package just to use
+            // 'SmtpEmailProvider.TechnicalName'.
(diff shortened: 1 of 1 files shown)


Piedone:
Use change-specified PR titles in submodules, please.

Lombiq.Tests.UI/Extensions/EmailUITestContextExtensions.cs
```diff
@@ -154,6 +154,11 @@ public static async Task ConfigureSmtpSettingsAsync(
         {
             await context.ClickReliablyOnAsync(By.ClassName("save"));
             context.Get(By.ClassName("validation-summary-errors").Safely())?.Text?.Trim().ShouldBeNullOrEmpty();
+
+            // "SMTP" is wired in so that we don't have to reference the 'OrchardCore.Email.Smtp' package just to use
+            // 'SmtpEmailProvider.TechnicalName'.
+            await context.SetDropdownByValueAsync(By.Id("ISite_EmailSettings_DefaultProvider"), "SMTP");
+            await context.ClickReliablyOnAsync(By.ClassName("save"));
```
Piedone, line 161: Isn't this only needed when there are multiple providers, so this should be conditional?

BenedekFarkas: The UI is still there (regardless of how many providers there are) and this step is quite simple, so I figured the extra code is not worth it.

Piedone (approved):

PR #519, opened 2025-04-24
LMBQ-462: Waiting for Elasticsearch index build before tests

DemeSzabolcs:
[LMBQ-462](https://lombiq.atlassian.net/browse/LMBQ-462)

[LMBQ-462]: https://lombiq.atlassian.net/browse/LMBQ-462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
+++ b/Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
@@ -92,6 +92,8 @@ public record ElasticsearchRunningContext(Guid Id, string Prefix)
                     // a delay.
                     await Task.Delay(500, cancellation);
                 }
+
+                stopWatch.Stop();
             }
         });
 
--- a/Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
+++ b/Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
@@ -6,6 +6,7 @@ using Nest;
 using OrchardCore.Indexing;
 using OrchardCore.Search.Elasticsearch.Core.Services;
 using System;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Threading;
@@ -72,10 +73,19 @@ public record ElasticsearchRunningContext(Guid Id, string Prefix)
 
                 long? lastFinishedTaskId = null;
 
+                var timeout = TimeSpan.FromSeconds(60);
+                var stopWatch = Stopwatch.StartNew();
+
                 // We have the id of the last indexing task that should happen, so we are waiting here for that task to
                 // complete, since "GetLastTaskId()" returns only completed tasks.
                 while (lastTaskId > lastFinishedTaskId || lastFinishedTaskId == null)
                 {
+                    if (stopWatch.Elapsed > timeout)
+                    {
+                        stopWatch.Stop();
+                        throw new TimeoutException($"Last finished tasked id did not match with last task id within {timeout}.");
+                    }
+
                     lastFinishedTaskId = await TryGetLastTaskIdAsync(elasticIndexManager, exactIndexName);
 
                     // The indexing takes a couple of seconds, so there is no need to check them so fast: we are adding
--- a/Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
+++ b/Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
@@ -74,7 +74,7 @@ public record ElasticsearchRunningContext(Guid Id, string Prefix)
 
                 // We have the id of the last indexing task that should happen, so we are waiting here for that task to
                 // complete, since "GetLastTaskId()" returns only completed tasks.
-                while (lastTaskId < lastFinishedTaskId)
+                while (lastTaskId > lastFinishedTaskId)
                 {
                     lastFinishedTaskId = await TryGetLastTaskIdAsync(elasticIndexManager, exactIndexName);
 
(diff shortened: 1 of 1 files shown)


Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
```diff
@@ -33,8 +36,68 @@ public Task BeforeTestAsync(UITestContext context) =>
...
+
+            // We are getting the last indexing task (regardless of the state). This function works like a cursor, so
+            // there is no way to get directly the last task in the list. Since we have to give a "count" parameter, we
+            // are retrieving the indexing tasks by batches of 1000. Then if there is no more, we get the last one.
+            if (exactIndexName != null)
+            {
+                do
+                {
+                    var tasks = (await indexingTaskManager.GetIndexingTasksAsync(startIndex, batchSize)).ToList();
+                    taskCount = tasks.Count;
```
sarahelsaig: You only really care about the last task, if there are any tasks. So I'd avoid converting an enumerable with potentially 1000 items into a list, as that could be pretty wasteful. Why not use `var lastTask = (...).LastOrDefault()` instead? Also instead of `int taskCount`, you could use `bool hasTask` because the actual amount of tasks does not matter.

Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
```diff
@@ -33,8 +36,68 @@ public Task BeforeTestAsync(UITestContext context) =>
...
+
+                    if (taskCount > 0)
+                    {
+                        lastTaskId = tasks[^1].Id;
+                        startIndex += batchSize;
+                    }
+                }
+                while (taskCount > 0);
```
sarahelsaig: This is a kind of batching iteration, instead of `do... while` maybe
`for (var startIndex = 0; hasTask; startIndex += batchSize)` would be more appropriate.

Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
```diff
@@ -33,8 +36,68 @@ public Task BeforeTestAsync(UITestContext context) =>
...
+                // We have the id of the last indexing task that should happen, so we are waiting here for that task to
+                // complete, since "GetLastTaskId()" returns only completed tasks.
+                while (lastTaskId != lastFinishedTaskId)
+                {
+                    lastFinishedTaskId = await TryGetLastTaskIdAsync(elasticIndexManager, exactIndexName);
+
+                    // The indexing takes a couple of seconds, so there is no need to check them so fast: we are adding
+                    // a delay.
+                    await Task.Delay(500, cancellation);
+                }
```
sarahelsaig: Shouldn't there be a timeout or max retry count here, to avoid hanging?

Also while unlikely, could another thread add a new indexing task a meantime, causing a hang because `lastTaskId < lastFinishedTaskId`?
So maybe it would be better to write `while (lastTaskId > lastFinishedTaskId)`.

Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
```diff
@@ -33,8 +36,68 @@ public Task BeforeTestAsync(UITestContext context) =>
...
+    // It takes time to "LastTaskId" to get set (0,5 - 1 seconds), until then it throws "InvalidOperationException", so
+    // we are retrieving it safely. This can happen if there are just a couple indexes, so we don't leave time for it to
+    // initialize.
+    private static async Task<long?> TryGetLastTaskIdAsync(ElasticIndexManager elasticIndexManager, string indexName)
+    {
+        try
+        {
+            return await elasticIndexManager.GetLastTaskId(indexName);
+        }
+        catch
```
sarahelsaig: This should not catch unrelated exceptions.
```suggestion
        catch (InvalidOperationException)
```

Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
```diff
@@ -33,8 +36,68 @@ public Task BeforeTestAsync(UITestContext context) =>
...
+                    // The indexing takes a couple of seconds, so there is no need to check them so fast: we are adding
+                    // a delay.
+                    await Task.Delay(500, cancellation);
+                }
+            }
         });
 
+    // It takes time to "LastTaskId" to get set (0,5 - 1 seconds), until then it throws "InvalidOperationException", so
+    // we are retrieving it safely. This can happen if there are just a couple indexes, so we don't leave time for it to
+    // initialize.
```
sarahelsaig: I feel like you are over-explaining technical details. Also if you want to comment on a method you should use `<summary>` even if it's a private method.

```suggestion
    /// <summary>
    /// Asking for the last task ID can throw an exception if the underlying value is not initialized yet. This method 
    /// catches the exception and returns null instead so it can be safely retried.
    /// </summary>
```

sarahelsaig (changes requested):

Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
DemeSzabolcs: Is a `Stopwatch` okay for timeout logic, or what would you recommend?

sarahelsaig: I figure you can just store `IClock.UtcNow` at the start and then check the difference inside the loop.

DemeSzabolcs: In the end I went with stopwatch, because `IClock` would require me to add it as a DE then pass it down through `UITestExecutionSession`. Plus I would have to subtract a value from the start time, then compare it to the timeout. Now I can just say stopWatch.Elapsed > timeout

Piedone: Use the test cancellation token if possible.

Piedone: Yeah, it should be.

PR #520, opened 2025-04-26
LMBQ-462: Updating ElasticsearchRunningContext with cancellation token 

DemeSzabolcs:
[LMBQ-462](https://lombiq.atlassian.net/browse/LMBQ-462)

[LMBQ-462]: https://lombiq.atlassian.net/browse/LMBQ-462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
+++ b/Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
@@ -80,6 +80,7 @@ public record ElasticsearchRunningContext(Guid Id, string Prefix)
                 // complete, since "GetLastTaskId()" returns only completed tasks.
                 while (lastTaskId > lastFinishedTaskId || lastFinishedTaskId == null)
                 {
+                    cancellation.ThrowIfCancellationRequested();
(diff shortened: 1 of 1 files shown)


Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
```diff
@@ -80,6 +80,7 @@ public Task BeforeTestAsync(UITestContext context) =>
                 // complete, since "GetLastTaskId()" returns only completed tasks.
                 while (lastTaskId > lastFinishedTaskId || lastFinishedTaskId == null)
                 {
+                    cancellation.ThrowIfCancellationRequested();
```
sarahelsaig: Don't you need to stop the `stopWatch` before throwing from cancellation? I'd put the whole `while` block into a `try..finally` and only call `stopWatch.Stop()` in the `finally` once.

sarahelsaig (changes requested):

PR #523, opened 2025-04-27
OSOE-1105: Moving Elasticsearch sample test to here from OSOCE and small Elasticsearch improvements

Piedone:
[OSOE-1105](https://lombiq.atlassian.net/browse/OSOE-1105)

[OSOE-1105]: https://lombiq.atlassian.net/browse/OSOE-1105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Extensions/TestContextExtensions.cs
```diff
@@ -1,30 +1,32 @@
 using System;
-using System.Diagnostics.CodeAnalysis;
 using Xunit;
 
 namespace Lombiq.Tests.UI.Extensions;
 
 public static class TestContextExtensions
 {
+    [Obsolete("Use GetElasticsearchSafeIndexName instead. This method will be removed in a future version.")]
+    public static string GetElasticserachSafeIndexName(this ITestContext context, Guid id) =>
```
sarahelsaig, line 9: :joy: :melting_face:

Lombiq.Tests.UI/Extensions/TestContextExtensions.cs
```diff
@@ -1,30 +1,32 @@
...
     {
+        // Elasticsearch indexes are lowercase only.
+#pragma warning disable CA1308 // Normalize strings to uppercase
         var name = context?
             .Test?
             .TestDisplayName?
             .ToLowerInvariant()
             .RegexReplace("[^a-z0-9]+", "-")
             .Trim('-');
+#pragma warning restore CA1308 // Normalize strings to uppercase
```
sarahelsaig, line 29: What's the reason of this change? In my opinion `#pragma warning` directives are annoying and make readability significantly worse than `SuppressMessage`. Also this isn't a huge method where boxing in just a smaller segment would be meaningful.

Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
```diff
@@ -17,93 +16,109 @@ namespace Lombiq.Tests.UI.Models;
...
     /// <summary>
-    /// Gets the expression that refers to all indexes that starts with <see cref="Prefix"/>. This should only be used
+    /// Gets the expression that refers to all indexes that start with <see cref="Prefix"/>. This should only be used
     /// with <see cref="IElasticClient"/>, because the OrchardCore-specific services automatically apply the prefix from
     /// configuration so it would result in double prefixing.
     /// </summary>
     private IndexName LowLevelIndexName => Indices.Index($"{Prefix}_*");
 
+    // Elasticsearch indexing sometimes takes longer, and the testing starts before it finishes. To prevent that, we are
+    // checking if all of the indexing tasks are finished.
```
sarahelsaig: The phrasing of "the testing starts before it finishes" is unclear. The "it" reads like we are still talking about the testing, as in "the testing starts before the testing finishes".
```suggestion
    // Elasticsearch indexing sometimes takes longer, and the testing starts before indexing finishes. To prevent that,
    // we are checking if all indexing tasks are finished.
```

Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
```diff
@@ -17,93 +16,109 @@ namespace Lombiq.Tests.UI.Models;
...
-                long? lastFinishedTaskId = null;
+            long? lastFinishedTaskId = null;
 
-                var timeout = TimeSpan.FromSeconds(60);
-                var stopWatch = Stopwatch.StartNew();
+            try
+            {
+                var lastTaskId = await GetLastTaskIdAsync(provider.GetRequiredService<IIndexingTaskManager>());
 
                 // We have the id of the last indexing task that should happen, so we are waiting here for that task to
```
sarahelsaig:
```suggestion
                // We have the ID of the last indexing task that should happen, so we are waiting here for that task to
```

sarahelsaig (changes requested):

Lombiq.Tests.UI/Extensions/TestContextExtensions.cs
Piedone: 🫠

Lombiq.Tests.UI/Extensions/TestContextExtensions.cs
Piedone: This is about boxing in a smaller segment, though. It's indeed not a huge method, but still, this minimizes the scope of the suppression. So if anybody adds another, but this time mistaken, `ToLowerInvariant()` two lines below, they'll get a warning. I don't trust anyone to keep an eye on method-level attributes while doing that.

PR #567, opened 2025-07-07
NEST-566: Create an abstract TeamsNotifierCloudflareRemoteUITestBase and other DRYing

wAsnk:
[NEST-566](https://lombiq.atlassian.net/browse/NEST-566)

[NEST-566]: https://lombiq.atlassian.net/browse/NEST-566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Extensions/RemoteUITestContextExtensions.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Lombiq.Tests.UI.Services;
+using System;
+
+namespace Lombiq.Tests.UI.Extensions;
+
+public static class RemoteUITestContextExtensions
+{
+    public static bool IsLocaleUITest(this UITestContext context) => context.TestStartUri.Host.ContainsOrdinalIgnoreCase("localhost");
```
DemeSzabolcs:
```suggestion
    public static bool IsLocalUITest(this UITestContext context) => context.TestStartUri.Host.ContainsOrdinalIgnoreCase("localhost");
```

Lombiq.Tests.UI/Helpers/TeamsHelper.cs
```diff
@@ -0,0 +1,106 @@
...
+    /// <remarks>
+    /// <para>
+    /// This needs a workflow in Teams. Add one starting with the "When a Teams webhook request is received"
+    /// trigger ("Who can trigger the flow?" -> Anyone). Then add the "Post card in a chat or channel" step with
+    /// `@triggerBody()` (the @ is important) as the "Adaptive Card" value. Just paste this into the text box,
+    /// trying to insert it as a dynamic content or expression value won't work.
+    /// </para>
+    /// </remarks>
+    /// </summary>
+    public static Task<HttpResponseMessage> SendFailedUiTestTeamsMessageAsync(string webhookUrl, string siteName)
```
DemeSzabolcs:
```suggestion
    public static Task<HttpResponseMessage> SendFailedUITestTeamsMessageAsync(string webhookUrl, string siteName)
```

Lombiq.Tests.UI/Helpers/TeamsHelper.cs
```diff
@@ -0,0 +1,106 @@
...
+    {
+        // This needs a workflow in Teams. Add one starting with the "When a Teams webhook request is received"
+        // trigger ("Who can trigger the flow?" -> Anyone). Then add the "Post card in a chat or channel" step with
+        // `@triggerBody()` (the @ is important) as the "Adaptive Card" value. Just paste this into the text box,
+        // trying to insert it as a dynamic content or expression value won't work.
```
DemeSzabolcs: The same text is also in the `summary`.

Lombiq.Tests.UI/TeamsNotifierCloudflareRemoteUITestBase.cs
```diff
@@ -0,0 +1,58 @@
...
+            TeamsMessageWasSent = true;
+
+            var isSuccessful = await TeamsHelper.SendFailedUiTestTeamsMessageAsync(TestFailedTeamsWebhookUrl, SiteName);
+
+            if (!isSuccessful.IsSuccessStatusCode)
+            {
+                _testOutputHelper.WriteLine($"Failed to send message to Teams with the following response: {isSuccessful}.");
+            }
+
+            throw;
```
DemeSzabolcs: Can't this also be part of `SendFailedUiTestTeamsMessageAsync`?

DemeSzabolcs (changes requested):

Lombiq.Tests.UI/TeamsNotifierCloudflareRemoteUITestBase.cs
wAsnk: I considered that. But then the logic separated into `SendFailedUiTestTeamsMessageAsync` would not be only sending the teams message, but it would also consider the response, which is something you would want to do when you are a caller of this function. If we would do this then we could add a new function called `SendAndCheckResponseForUITestTeamsMessageAsync`, which would also need a `ITestOutputHelper` as an input. Or I should then create a service instead of a helper, so it can DI the `ITestOutputHelper` itself. But I don't think it makes sense.

DemeSzabolcs (approved):

DemeSzabolcs (approved):

PR #582, opened 2025-07-31
OSOE-1069: Remove page classes in Lombiq.UITestingToolbox

### LINKED ISSUE #481: Remove page classes (OSOE-1069)
We have several page classes in the `Pages` namespace. While in the beginning it seemed like a good idea to write tests with page classes, we haven't done that in the end, and using extension methods for common operations turned out to be a better approach. So, now these classes are a parallel infrastructure that we have to maintain with awkward workarounds like manual `RefreshCurrentAtataContext()` and `TriggerAfterPageChangeEventAsync()` calls. Let's remove these classes and substitute their features with standard extension methods.

- [x] All page class behavior is reimplemented with browser operations and extension methods.
- [x] Nothing in the UI Testing Toolbox and OSOCE uses page classes anymore.
- [x] All page classes and the extension methods using them (like `GoToPageAsync<T>()`) are marked with `[Obsolete]`, pointing people to what to use instead.

This is necessary to do before Atata v4 ([see what's coming](https://atata.io/blog/2025/04/16/what-is-coming-to-atata-4/)), so we can deprecate everything remaining, release a UITT version, and then update Atata while removing the deprecated methods, not having to update them to Atata v4. At that point, we can remove the `Atata.Bootstrap` reference too.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-1069)

sarahelsaig:
[OSOE-1069](https://lombiq.atlassian.net/browse/OSOE-1069)
Fixes #481

[OSOE-1069]: https://lombiq.atlassian.net/browse/OSOE-1069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs
@@ -441,7 +441,9 @@ public static class BasicFeaturesTestingUITestContextExtensions
                 await context.GoToLoginAsync();
                 await context.ClickReliablyOnAsync(By.CssSelector("a[href*='/Register']"));
                 await parameters.RegisterAsync(context, navigate: false);
-                context.Driver.Url.ShouldNotBe(context.GetAbsoluteUri(UserRegistrationParameters.DefaultUrl).AbsoluteUri);
+
+                var registerUrl = context.GetAbsoluteUri(UserRegistrationParameters.DefaultUrl).AbsoluteUri;
+                context.DoWithRetriesOrFail(() => !registerUrl.EqualsOrdinalIgnoreCase(context.Driver.Url));
 
                 (await context.GetCurrentUserNameAsync()).ShouldBe(parameters.UserName);
                 await context.SignOutDirectlyAsync();
--- a/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs
@@ -569,7 +569,6 @@ public static class BasicFeaturesTestingUITestContextExtensions
                 };
 
                 await customPageHeaderCheckAsync(context);
-                if (context.Driver.WindowHandles.Count > 1) context.Driver.Close();
             });
 
     /// <summary>
--- a/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs
@@ -263,9 +263,11 @@ public static class BasicFeaturesTestingUITestContextExtensions
     }
 
     /// <summary>
-    /// <para>Tests the site setup with the recipe with the specified <paramref name="setupRecipeId"/> and tests the
-    /// site setup negatively. Negative test uses new <see cref="OrchardCoreSetupParameters"/> instance with empty
-    /// values of properties: </para>
+    /// <para>
+    /// Tests the site setup with the recipe with the specified <paramref name="setupRecipeId"/> and tests the site
+    /// setup negatively. Negative test uses new <see cref="OrchardCoreSetupParameters"/> instance with empty values of
+    /// the following properties:
+    /// </para>
     /// <list type="bullet">
     ///     <item><description><see cref="OrchardCoreSetupParameters.SiteName"/></description></item>
     ///     <item><description><see cref="OrchardCoreSetupParameters.UserName"/></description></item>
--- a/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs
@@ -2,12 +2,10 @@ using Atata;
 using Lombiq.Tests.UI.Constants;
 using Lombiq.Tests.UI.Extensions;
 using Lombiq.Tests.UI.Models;
-using Lombiq.Tests.UI.Pages;
 using Lombiq.Tests.UI.Services;
 using OpenQA.Selenium;
 using Shouldly;
 using System;
-using System.Linq;
 using System.Threading.Tasks;
 
 namespace Lombiq.Tests.UI.BasicOrchardFeaturesTesting;
@@ -337,7 +335,7 @@ public static class BasicFeaturesTestingUITestContextExtensions
             {
                 if (signOut) await context.SignOutDirectlyAsync();
 
-                await new UserLoginParameters(userName, password, loginButtonText).LogInAsync(context);
+                await context.LogInAsync(new(userName, password, loginButtonText));
 
                 var currentUser = await context.GetCurrentUserNameAsync();
                 if (shouldBeSuccess)
@@ -434,7 +432,7 @@ public static class BasicFeaturesTestingUITestContextExtensions
                 (await context.GetCurrentUserNameAsync()).ShouldBe(parameters.UserName);
                 await context.SignOutDirectlyAsync();
 
-                await new UserLoginParameters(parameters).LogInAsync(context);
+                await context.LogInAsync(parameters);
                 await context.TriggerAfterPageChangeEventAsync();
                 (await context.GetCurrentUserNameAsync()).ShouldBe(parameters.UserName);
                 await context.SignOutDirectlyAsync();
--- a/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs
@@ -4,7 +4,6 @@ using Lombiq.Tests.UI.Extensions;
 using Lombiq.Tests.UI.Models;
 using Lombiq.Tests.UI.Pages;
 using Lombiq.Tests.UI.Services;
-using Microsoft.SqlServer.TransactSql.ScriptDom;
 using OpenQA.Selenium;
 using Shouldly;
 using System;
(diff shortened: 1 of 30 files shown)


Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -204,26 +213,51 @@ public static Task<Uri> GoToSetupPageAndSetupOrchardCoreAsync(this UITestContext
 
     public static async Task<Uri> GoToSetupPageAndSetupOrchardCoreAsync(
```
Piedone, line 228: Keep this method, because it's Page class-specific, and add a new alternative instead (like `GoToSetupAndSetupOrchardCoreAsync`).

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -204,26 +213,51 @@ public static Task<Uri> GoToSetupPageAndSetupOrchardCoreAsync(this UITestContext
...
+    /// Navigate to the Features admin configuration page.
+    /// </summary>
+    public static Task GoToFeaturesAsync(this UITestContext context) =>
+        context.GoToAdminRelativeUrlAsync("/Features");
+
+    /// <summary>
+    /// Navigate to the Features admin configuration page and search for the provided text.
+    /// </summary>
+    /// <returns>A collection of visible checkbox elements.</returns>
+    public static async Task<IEnumerable<IWebElement>> GoToFeaturesAsync(this UITestContext context, string search)
```
Piedone: Use a more specific name, since this is a very specific method (and it's not just going to the Feature screen). But do we really need it here in public? It seems only useful (and is only used) in `TestTurningFeatureOnAndOffAsync`.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -204,26 +213,51 @@ public static Task<Uri> GoToSetupPageAndSetupOrchardCoreAsync(this UITestContext
...
 
-        return setupPage.PageUri.Value;
+        return new(context.Driver.Url);
     }
 
+    [Obsolete("Methods using Page<> classes will be removed in the next version.")]
     public static Task<OrchardCoreRegistrationPage> GoToRegistrationPageAsync(this UITestContext context) =>
         context.GoToPageAsync<OrchardCoreRegistrationPage>();
 
+    [Obsolete($"Use {nameof(GoToAdminAsync)} method instead.")]
```
Piedone: Add new versions for the other methods that have "dashboard" in the name too and obsolete the other ones.

But. There are a lot of places where we refer to the admin area as dashboard, starting with the OC UI itself, of course. But therre's also `OrchardCoreDashboardUITestContextExtensions` and a lot of other UI testing-related stuff with the same terminology. So, it would better to keep that here too.

Lombiq.Tests.UI/BasicOrchardFeaturesTesting/BasicFeaturesTestingUITestContextExtensions.cs
```diff
@@ -17,18 +18,16 @@ namespace Lombiq.Tests.UI.BasicOrchardFeaturesTesting;
...
-    /// </para>
-    /// <para>The test method assumes that the site is not set up.</para>
     /// </summary>
     /// <param name="setupRecipeId">The ID of the recipe to be used to set up the site.</param>
     /// <param name="customPageHeaderCheckAsync">
     /// The custom page header check logic to locate and/or check the header's text. This ultimately gets passed to
     /// TestContentOperationsAsync().
     /// </param>
-    /// <returns>The same <see cref="UITestContext"/> instance.</returns>
+    /// <returns>The instance passed to <paramref name="context"/>.</returns>
```
Piedone: I guess this is some leftover, because neither of the methods return the context now.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -189,9 +196,11 @@ await context.DoWithRetriesUntilNavigationHasOccurredOrFailAsync(
         return page;
     }
 
+    [Obsolete("Methods using Page<> classes will be removed in the next version.")]
     public static Task<OrchardCoreSetupPage> GoToSetupPageAsync(this UITestContext context, bool navigate = true) =>
         context.GoToPageAsync<OrchardCoreSetupPage>(navigate);
 
+    [Obsolete("Methods using Page<> classes will be removed in the next version.")]
```
Piedone: Point people to the new method corresponding to each method too, like "Use GoToLoginAsync instead."

Lombiq.Tests.UI/Models/UserLoginParameters.cs
```diff
@@ -0,0 +1,56 @@
...
+    string LoginButtonText = UserLoginParameters.DefaultLoginButtonText)
+{
+    public const string DefaultLoginButtonText = "Log in";
+
+    public UserLoginParameters(UserRegistrationParameters parameters)
+        : this(parameters.UserName, parameters.Password, parameters.LoginButtonText)
+    {
+    }
+
+    public async Task LogInAsync(UITestContext context, bool navigate = true)
```
Piedone, line 23: Having such an operation in a model class doesn't look appropriate. Add a `UITestContext` extension instead that expects this class as the input. Same for `UserRegistrationParameters.RegisterAsync` and `OrchardCoreSetupParameters.SetupOrchardCoreAsync`.

Lombiq.Tests.UI/Extensions/FormWebDriverExtensions.cs
```diff
@@ -10,6 +10,8 @@ public static IWebElement TryFillElement(this IWebDriver driver, IWebElement ele
     {
         element.ClearWithLogging();
 
+        if (string.IsNullOrEmpty(value)) return element;
```
Piedone, line 13: It's valid to empty out an input element, i.e. fill it with an empty string. Why this shortcut?

Lombiq.Tests.UI/Extensions/OrchardCoreDashboardUITestContextExtensions.cs
```diff
@@ -117,4 +143,22 @@ public static Task GoToContentItemEditorByIdAsync(this UITestContext context, st
...
+    /// Navigate to the Registration page.
+    /// </summary>
+    public static Task GoToRegistrationAsync(this UITestContext context) =>
+        context.GoToRelativeUrlAsync('/' + UserRegistrationParameters.DefaultUrl);
+
+    /// <summary>
+    /// Navigate to the Login page.
+    /// </summary>
+    public static Task GoToLoginAsync(this UITestContext context) =>
+        context.GoToRelativeUrlAsync("/Login");
```
Piedone: These two aren't really for the dashboard and would be more appropriate in `NavigationUITestContextExtensions`.

Lombiq.Tests.UI/Pages/OrchardCoreSetupParameters.cs
Piedone, line 1: Since we'll also remove the `Pages` namespace, this should rather be in `Models`.

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
sarahelsaig: I needed a different name, because for `Task<OrchardCoreDashboardPage> GoToDashboardAsync(this UITestContext context)` the only Page thing is the return value. So if we keep it as `[Obsolete]` than they can't clash. What should it be called?

Lombiq.Tests.UI/Models/UserLoginParameters.cs
sarahelsaig: Why is it inappropriate? It's a UI test model, using its own data in UI test seems like valid OOP to me. Makes the code more discoverable too, as opposed to having to find the logic in one of the million `*UITestContextExtensions` classes.

Lombiq.Tests.UI/Extensions/FormWebDriverExtensions.cs
sarahelsaig: The method has already cleared out the input by this point using the `element.ClearWithLogging();` in the line above. So if the value is `null` or empty, there is nothing further to do. Also, if the value was actually `null` we'd get an NRE. on the next line.

Lombiq.Tests.UI/Pages/OrchardCoreSetupParameters.cs
sarahelsaig: Shouldn't that be done when we actually remove them?

Lombiq.Tests.UI/Models/UserLoginParameters.cs
Piedone: Models ought to really only have methods that operate on the model's data, if at all (this is what we follow with MVC models and services too), but this here operates on the surrounding world. It's not really discoverable, because you have to know about the class. The extension methods are out there in a usually imported namespace, so typing `context.login` would bring it up.

Lombiq.Tests.UI/Pages/OrchardCoreSetupParameters.cs
Piedone: Ideally yes, but we shouldn't have it here until then, because code using the new approach will rely on it. The best would be to move this class to `Models`, restore the original version here and obsolete it, pointing to the other one.

Lombiq.Tests.UI/Extensions/FormWebDriverExtensions.cs
Piedone: Ah OK.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Piedone: I don't know. My best suggestion is break this by removing the Page return type, and adding a temporary, obsolete `GoToDashboardPageAsync` for those using this.

Lombiq.Tests.UI/Models/UserLoginParameters.cs
sarahelsaig: Here we send the data of this object to the driver, so this is categorically identical to a common WriteTo type method that takes a stream or serializer as argument. Also UI tests aren't MVC, so that feels like forcing the square peg into the circle hole ([when it should be the other way around](https://www.youtube.com/watch?v=6pDH66X3ClA)).

I agree with your point about discoverability, but to that effect a shortcut extension method like `public Task LogInAsync(this UITestContext context, UserLoginParameters parameters, bool navigate = true) => parameters.LogInAsync(context, navigate)` would be better.

Piedone: Do you know where the UI testing shortcut goes? That's right: it goes into a `UITestContext` extension method.

I'm OK with a shortcut.

Piedone (approved):

PR #607, opened 2025-09-16
NEST-627: Calculate absolute path respecting RequestUrlPrefix

wAsnk:
[NEST-627](https://lombiq.atlassian.net/browse/NEST-627)

[NEST-627]: https://lombiq.atlassian.net/browse/NEST-627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
@@ -74,6 +74,8 @@ public static class NavigationUITestContextExtensions
 
     public static string GetCurrentAbsolutePath(this UITestContext context) => context.GetCurrentUri().AbsolutePath;
 
+    // A simple new(context.Scope.BaseUri, relativeUrl.TrimStart('/')) would work for most cases but not when using
+    // tenants with a RequestUrlPrefix, because relativeUrl would then be relative to the host.
     public static Uri GetAbsoluteUri(this UITestContext context, string relativeUrl) =>
         new(context.Scope.BaseUri.OriginalString.TrimEnd('/') + "/" + relativeUrl.TrimStart('/'));
 
--- a/Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
@@ -87,7 +87,7 @@ public static class NavigationUITestContextExtensions
             Port = context.Scope.BaseUri.Port,
             Path = string.IsNullOrEmpty(localPath) || localPath == "/"
                 ? path
-                : localPath.TrimStart('/') + "/" + path,
+                : localPath.TrimStart('/').TrimEnd('/') + "/" + path,
             Query = query,
         };
         return uriBuilder.Uri;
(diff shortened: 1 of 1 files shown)


Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -74,8 +74,24 @@ await context.Configuration.Events.AfterNavigation
...
+        {
+            Scheme = context.Scope.BaseUri.Scheme,
+            Host = context.Scope.BaseUri.Host,
+            Port = context.Scope.BaseUri.Port,
+            Path = string.IsNullOrEmpty(localPath) || localPath == "/"
+                ? path
+                : localPath.TrimStart('/').TrimEnd('/') + "/" + path,
+            Query = query,
+        };
+        return uriBuilder.Uri;
```
Piedone: Why is this logic necessary? Explain it in a comment.

But is it really necessary? `RequestUrlPrefix` is only for tenants, and when calling `UITestContext.SwitchCurrentTenant()`, which should happen when switching between tenants, the `BaseUri` will be suffixed with it.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -74,8 +74,24 @@ await context.Configuration.Events.AfterNavigation
...
     public static string GetCurrentAbsolutePath(this UITestContext context) => context.GetCurrentUri().AbsolutePath;
 
-    public static Uri GetAbsoluteUri(this UITestContext context, string relativeUrl) =>
-        new(context.Scope.BaseUri, relativeUrl.TrimStart('/'));
+    public static Uri GetAbsoluteUri(this UITestContext context, string relativeUrl)
+    {
+        var localPath = context.Scope.BaseUri.LocalPath;
+        var splitRelativeUrl = relativeUrl.TrimStart('/').Split('?');
+        var query = splitRelativeUrl.Length > 1 ? splitRelativeUrl[1] : string.Empty;
+        var path = splitRelativeUrl[0];
```
Piedone: The query and path are also available in `context.Scope.BaseUri.AbsolutePath` and `context.Scope.BaseUri.Query`. Why is this logic, and using `LocalPath`, necessary?

Piedone (changes requested):

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
wAsnk: Yes, it will be suffixed with it but `new Uri(Uri, string)` does the following:
Base uri: `https://tenants-staging.dotnest.net/editeddotnesttenantscoretenantmanagementremoteuitest`
relativeUrl: `/login`
Output: `https://tenants-staging.dotnest.net/login`

So the path will be deleted from the base URI.

Let me know if you would prefer 
`new Uri(context.Scope.BaseUri.OriginalString + "/" + relativeUrl.TrimStart('/'))` 
instead.

Piedone: Then rather the base URI should be suffixed with `/` if that's still accepted by OC for `RequestUrlPrefix`es (when it's just a host, it's always suffixed), or if not, then what you write.

wAsnk: In OC setup that's not accepted. In the `SwitchCurrentTenant` it is specifically instructed not to use trailing slashes. So I believe there is a reason for that. E.g.other parts of UI testing break if we do that.

image

Also I don't see why we shouldn't just append the relativeUrl to the baseUri. So I will go with `new Uri(context.Scope.BaseUri.OriginalString + "/" + relativeUrl.TrimStart('/'))`

PR #691, opened 2026-01-19
OSOE-925: Upgrade OC preview

sarahelsaig:
[OSOE-925](https://lombiq.atlassian.net/browse/OSOE-925)

[OSOE-925]: https://lombiq.atlassian.net/browse/OSOE-925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
```diff
@@ -268,6 +268,12 @@ public static Task EnableFeatureDirectlyAsync(
             tenant,
             activateShell);
 
+    /// <summary>
+    /// Enables the Tenants feature for the default tenant.
+    /// </summary>
+    public static Task EnableTenantsFeatureAsync(this UITestContext context) =>
```
Piedone:
```suggestion
    public static Task EnableTenantsFeatureDirectlyAsync(this UITestContext context) =>
```

Lombiq.Tests.UI/Extensions/ShortcutsUITestContextExtensions.cs
```diff
@@ -493,14 +499,20 @@ public static async Task CreateAndSwitchToTenantAsync(
         string name,
         string urlPrefix,
         OrchardCoreSetupParameters setupParameters,
-        string featureProfile = null)
+        string featureProfile = null,
+        bool enableFeature = true)
```
Piedone: Same in Lombiq.Tests.UI/Extensions/TenantsUITestContextExtensions.cs.

```suggestion
        bool enableTenantsFeature = true)
```

Lombiq.Tests.UI/Models/ElasticsearchRunningContext.cs
Piedone, line 1: I love when a lot of code is ripped out and the result is still better.

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
```diff
@@ -152,7 +152,7 @@ await context.RunAndAssertSecurityScanAsync(
                             "No job named \"openapi\" found in the Automation Framework Plan. We can only run the " +
                             "OpenAPI scan if the job exists.");
 
-                    apiDefinitionUri ??= context.GetAbsoluteUri("/swagger/v1/swagger.json");
+                    apiDefinitionUri ??= context.GetAbsoluteUri("/openapi/v1.json");
```
Piedone, line 155: Did you check if this still works with ZAP?

Piedone (changes requested):

Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs
sarahelsaig: Sure, but I'm not sure what kind of difference should I see if was or wasn't working. But both returns a standard OpenAPI JSON so it should be the same just under a different endpoint URL.

Piedone: If an OpenAPI scan still works then that's it. I believe we have one, but not sure which app.

Lombiq.Tests.UI/Services/HtmlValidationConfiguration.cs
```diff
@@ -52,11 +56,42 @@ public class HtmlValidationConfiguration
     /// </summary>
     public Action<HtmlValidationOptions> HtmlValidationOptionsAdjuster { get; set; }
 
+    /// <summary>
+    /// Gets a dictionary of filters. If not empty, it's used in <see
+    /// cref="HtmlValidationUITestContextExtensions.AssertHtmlValidityAsync"/> instead of  <see
+    /// cref="AssertHtmlValidationOutputIsEmptyAsync"/>.
+    /// </summary>
+    public IDictionary<string, Func<JsonHtmlValidationError, bool>> HtmlValidationFilters { get; } =
```
Piedone: Why do we need this approach? All assertion configs use a Func, and this works well for such assertions:

- Any one assertion failing means the test fails, so throwing exceptions is appropriate.
- That way, exception messages can be customized.
- Existing assertions can be overridden by setting the property, or composed by concatenating with it (i.e. no advantage for a dictionary).

We do have instances of having both a filter and assertions (see e.g. `BrowserLogFilter` and `AssertBrowserLog`), and if you find filtering necessary here, I'd also do that.

Piedone (changes requested):

Lombiq.Tests.UI/Services/HtmlValidationConfiguration.cs
sarahelsaig: The point of this dictionary is to include multiple _separate_ filters for culling the HTML validation logs before asserting emptiness, instead of a single `Func<HtmlValidationResult, Task>`. The filters are automatically applied and asserted [here](https://github.com/Lombiq/UI-Testing-Toolbox/commit/d448bb15124f826c7468ff82272cd234120771cf#diff-544802917b7b8361053ce8bda641cb756a5bf308811ddbd443e4ca8733921c26R63-R78). The previous approach had several problems:
- DRY: every time you add a custom log handler, it has to call `validationResult.GetParsedErrors()` and `errors.ShouldBeEmpty(HtmlValidationResultExtensions.GetParsedErrorMessageString(errors));`. This is evil.
- Re-usability: now we can confidently add filtering logic in an extension method, because it won't override the outer configuration.
- Composition: previously, you had to make a single compound expression and annotate the sources as comments. Not super maintainable. Now you can compose them. For example:

```csharp
            configuration => configuration.HtmlValidationConfiguration.AssertHtmlValidationResultAsync =
                validationResult =>
                {
                    configuration.HtmlValidationConfiguration.WithRelativeConfigPath("NoUniqueLandmark.htmlvalidate.json");

                    // The first three rule exclusions due to https://github.com/OrchardCMS/OrchardCore/issues/15222,
                    // can be removed once it is resolved.
                    // "aria-label-misuse" due to https://github.com/OrchardCMS/OrchardCore/issues/18510.
                    var errors = validationResult.GetParsedErrors()
                        .Where(error =>
                            error.RuleId is not "prefer-native-element" and
                                not "text-content" and
                                not "no-redundant-role" and
                                not "aria-label-misuse")
                        .ToList();
                    errors.ShouldBeEmpty(HtmlValidationResultExtensions.GetParsedErrorMessageString(errors));
                    return Task.CompletedTask;
                });
```
becomes
```csharp
            configuration => configuration.HtmlValidationConfiguration
                .WithRelativeConfigPath("NoUniqueLandmark.htmlvalidate.json")
                .WithOC15222Filter()
                // This is necessary until https://github.com/OrchardCMS/OrchardCore/pull/18730 is merged.
                .WithFilters("OC-18730", error => error.RuleId is not "aria-label-misuse"));
```

Isn't that a million times nicer? (also note the code-reuse in the `WithOC15222Filter`)

Additional notes:
> That way, exception messages can be customized.

I have never seen this actually done. Why would you want to customize the error message? In practice it seems like you just end up copy-pasting what already works, making the code base more wet.

> Existing assertions can be overridden by setting the property, or composed by concatenating with it (i.e. no advantage for a dictionary).

Hard disagree. The ability to overwrite is actually a hidden trap, because you don't necessarily know what is being overwritten and it's super hard to find where a `Func` is set after the fact so it's not nice debugging experience either. With the dictionary the used filter keys are automatically listed in the exception message, so diagnostics is cleaner.
Also as you can see from my above example, the ability to compose is not semantic and readability suffers.

> We do have instances of having both a filter and assertions

This is not the case here. As you can see [here](https://github.com/Lombiq/UI-Testing-Toolbox/commit/d448bb15124f826c7468ff82272cd234120771cf#diff-544802917b7b8361053ce8bda641cb756a5bf308811ddbd443e4ca8733921c26L33-R55) and as mentioned in the xmldoc above, `HtmlValidationFilters` supersedes `AssertHtmlValidationOutputIsEmptyAsync`. You use one or the other.

Piedone: OK, fair enough, thanks for explaining, agree . But I still think that having the ability to both add filters and customize assertions (this is what I'm talking about with the browser log example) brings value.

Piedone: Please address this.

sarahelsaig: I can't really see a reason for custom assertion logic besides testing the logger itself. (that's literally the only use-case for a custom `AssertBrowserLog` function in our code base) Not a huge value imo.

Piedone: This keeps options open, including customizing the assertion logic or using a different assertion library than Shouldly. All I'm asking is to keep that func as-is.

sarahelsaig: You don't have to ask, [it's already done](https://github.com/Lombiq/UI-Testing-Toolbox/pull/691/changes/ee47f6c00f32d8ccd03d3904731e93735bf9c5b8). I just don't see why.

Piedone: OK then.

Lombiq.Tests.UI/Services/HtmlValidationConfiguration.cs
```diff
@@ -52,11 +52,22 @@ public class HtmlValidationConfiguration
     /// </summary>
     public Action<HtmlValidationOptions> HtmlValidationOptionsAdjuster { get; set; }
 
+    /// <summary>
+    /// Gets a dictionary of filters. If <see cref="AssertHtmlValidationResultAsync"/> is <see langword="null"/>, then
+    /// these are automatically used instead. The errors from the <see cref="HtmlValidationResult"/> are filtered by
+    /// each entry and validation only fails if there is still any errors left over. If you use a custom <see
+    /// cref="AssertHtmlValidationResultAsync"/> value, these will only apply if you explicitly call them.
```
Piedone: The filtering should happen every time, with the default or custom assertion happening on the filtered values (again how it's for the browser log).

Piedone (changes requested):

Lombiq.Tests.UI/Services/HtmlValidationConfiguration.cs
sarahelsaig: It's only possible if I change
```csharp
Func<HtmlValidationResult, Task> AssertHtmlValidationResultAsync
```
to be
```csharp
Func<IList<JsonHtmlValidationError>, Task> AssertHtmlValidationResultAsync
```
instead. Same with every single `Func<HtmlValidationResult, Task> assertHtmlValidationResultAsync = null` optional parameter (there are a bunch on various methods). Should I do that?

Piedone: An alternative would be to re-serialize the filtered errors into a new `HtmlValidationResult`. However, at one point I think we should just bite the bullet, so yes.

Also, please rename `JsonHtmlValidationError` to `HtmlValidationError`; it coming from JSON is an implementation detail that won't matter.

Piedone (approved):

Lombiq.Tests.UI/Models/AccessibilityCheckingResult.cs
Piedone, line 1: Why exactly do we need this? I understand that some properties being `IList`s is more convenient than arrays (for item removals), but does that really warrant having a new type that otherwise duplicates `AxeResult`?

Piedone (changes requested):

Lombiq.Tests.UI/Models/AccessibilityCheckingResult.cs
sarahelsaig: You can only update AxeResult by converting it back into JSON and then de-serializing using Newtonsoft.Json. Except it doesn't work, because the serialization code doesn't omit the null-valued properties but when those are deserialzied it throws an exception. (i.e. this literally doesn't work: `axeResult = new AxeResult(JObject.Parse(axeResult.ToString))`, but it definitely should) This is presumably a bug in the library, but fiddling with the JSON data is quite hackish anyway (to me it's on the same level as editing a private field using reflection) so it wouldn't have been a great solution even if it worked.

So I've created this secondary class where the properties are converted into mutable `List<T>` (e.g. `Violations = [.. axeResult.Violations]`) so they are actually editable in place.

Piedone: I see. Let's call it `AccessibilityCheckingResult` and then we have declared freedom to be independent of what's in `AxeResult`.

Piedone (approved):

Piedone (approved):

Piedone (approved):

.github/workflows/publish-nuget.yml
Piedone, line 1: smtp4dev 3.15.0 switched from net8.0 to net10.0, so disabled that update in `dev`. But here please remove the disable from the Renovate config and update the tool to the latest version.

Lombiq.Tests.UI/Services/OrchardCoreInstance.cs
Piedone, line 1: Set YesSql `EnableThreadSafetyChecks` (see https://docs.orchardcore.net/en/latest/reference/modules/Data/#configuring-yessql) so we can catch DB concurrency issues.

Lombiq.Tests.UI/Lombiq.Tests.UI.csproj
```diff
@@ -100,19 +102,24 @@
...
-  </ItemGroup>
-
-  <ItemGroup Condition="'$(NuGetBuild)' == 'true'">
-    <PackageReference Include="Lombiq.Tests" Version="5.0.1-alpha.0.osoe-925" />
-    <PackageReference Include="Lombiq.HelpfulLibraries.Cli" Version="12.6.1-alpha.5.osoe-925" />
-    <PackageReference Include="Lombiq.HelpfulLibraries.OrchardCore" Version="12.6.1-alpha.5.osoe-925" />
-    <PackageReference Include="Lombiq.HelpfulLibraries.Refit" Version="12.6.1-alpha.5.osoe-925" />
-  </ItemGroup>
+  <Choose>
+    <When Condition="'$(NuGetBuild)' == 'true'">
```
Piedone, line 64: Why is this better than the `Condition` on the `ItemGroup`?

Piedone (changes requested):

Lombiq.Tests.UI/Lombiq.Tests.UI.csproj
sarahelsaig: It makes the two blocks connected by syntax. Also no condition text on the "otherwise" side, which is less room for error. It's not super important here, I was just feeling out the syntax.

Piedone: OK, that's good.

Piedone (approved):

Lombiq.Tests.UI/Models/NonResetElasticsearchIndexRebuildStep.cs
```diff
@@ -0,0 +1,89 @@
...
+
+namespace Lombiq.Tests.UI.Models;
+
+/// <summary>
+/// This recipe step rebuilds an Elasticsearch index. It's identical to <see cref="ElasticsearchIndexRebuildStep"/>,
+/// except the index isn't reset first.
+/// </summary>
+/// <remarks><para>
+/// This variant is necessary to avoid errors like this:
+/// [Error] OrchardCore.Elasticsearch.Core.Services.ElasticsearchDocumentIndexManager: There were issues updating
```
Piedone: Should these errors indeed happen? Isn't there something to fix in OC?

Piedone (changes requested):

Piedone (changes requested):
https://github.com/Lombiq/UI-Testing-Toolbox/pull/691/changes/BASE..79ab09ad1ec927a853e5a674cbe8159ffda2f152#r3486990963 isn't addressed.

Lombiq.Tests.UI/Models/NonResetElasticsearchIndexRebuildStep.cs
sarahelsaig: As far as I was able to assess this only happens during UI testing, so I doubt it can be addressed in OC.

Piedone: OK, then please add this to the comment.

However, do we need a recipe step? It seems to me that all the related code starting from `ElasticsearchRunningContext` should rather be direct Elasticsearch OC service class calls. I don't see why we go through recipe steps.

sarahelsaig: It made more sense back when it worked with the original recipe step as-is. Now I've turned it into an extension method instead.

Piedone (approved):

Lombiq.Tests.UI/Models/NonResetElasticsearchIndexRebuildStep.cs
Piedone: Great, thanks!

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -87,6 +87,11 @@ static UITestExecutionSession()
 
     public async Task<bool> ExecuteAsync(int retryCount, string dumpRootPath)
     {
+        new GitHubAnnotationWriter(_testOutputHelper).Annotate(
+            LogLevel.Warning,
+            "UI test may be flaky",
+            $"The {_testManifest.Name} test failed X time(s) and will be retried. This may indicate it being flaky.");
+
```
Piedone: Why add this placeholder message here?

Piedone (changes requested):

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
sarahelsaig: I was testing something and forgot to remove it.

Piedone (approved):

Piedone (changes requested):
NuGet publish fails.

Piedone (approved):

PR #696, opened 2026-02-03
OSOE-353: Add duplicate SQL query detector

### LINKED ISSUE #93: Add duplicate SQL query detector (OSOE-353)

wAsnk:
[OSOE-353](https://lombiq.atlassian.net/browse/OSOE-353)
Fixes #93

[OSOE-353]: https://lombiq.atlassian.net/browse/OSOE-353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/SqlQueryMonitoring/SqlQueryMonitoringMiddleware.cs
```diff
@@ -0,0 +1,56 @@
...
+
+            if (monitoringContext != null && store != null)
+            {
+                var tenantName = services.GetService<ShellSettings>()?.Name ?? ShellSettings.DefaultShellName;
+                var summary = new SqlQueryMonitoringSummary(
+                    tenantName: tenantName,
+                    requestPath: $"{context.Request.PathBase}{context.Request.Path}{context.Request.QueryString}",
+                    requestMethod: context.Request.Method,
+                    traceIdentifier: context.TraceIdentifier,
+                    completedUtc: DateTimeOffset.UtcNow,
```
Piedone: Use `IClock`.

Lombiq.Tests.UI/Docs/SqlQueryMonitoring.md
Piedone, line 1: Remove what's fluff. Include applicable example configs and assertions in the samples instead of documenting it here, where applicable.

Lombiq.Tests.UI/Docs/SqlQueryMonitoring.md
```diff
@@ -0,0 +1,165 @@
...
+configuration.SqlQueryMonitoringConfiguration.ResultSetRowCountThreshold = 150;
+```
+
+Timing override:
+
+```csharp
+configuration.SqlQueryMonitoringConfiguration.SummaryLookupTimeout = TimeSpan.FromSeconds(5);
+configuration.SqlQueryMonitoringConfiguration.SummaryLookupInterval = TimeSpan.FromMilliseconds(200);
+configuration.SqlQueryMonitoringConfiguration.FollowUpSummaryQuietPeriod = TimeSpan.FromMilliseconds(500);
+```
```
Piedone: What makes these typical? If they're typical, why aren't they the default?

Lombiq.Tests.UI/SqlQueryMonitoring/SqlQueryMonitoringStore.cs
```diff
@@ -0,0 +1,115 @@
...
+    private const int MaxEntries = 50;
+
+    private readonly Queue<SqlQueryMonitoringSummary> _summaries = new();
+    private readonly object _lock = new();
+
+    public void AddSummary(SqlQueryMonitoringSummary summary)
+    {
+        if (summary == null) return;
+
+        lock (_lock)
```
Piedone: Have you tried  `ConcurrentQueue` instead of locking, with trimming right here?

Lombiq.Tests.UI/SqlQueryMonitoring/SqlQueryMonitoringStore.cs
```diff
@@ -0,0 +1,115 @@
...
+            items.Enqueue(candidate);
+        }
+
+        while (items.Count > 0)
+        {
+            _summaries.Enqueue(items.Dequeue());
+        }
+
+        return removed;
+    }
```
Piedone: Why does this need a predicate and a copy to a second queue? Why not just dequeue until the max capacity is reached?

Lombiq.Tests.UI.TestingModule/Controllers/SqlQueryMonitoringScenarioController.cs
Piedone, line 1: This project is shipped to consumers, it mustn't include code that we use to test UITT itself.

Lombiq.Tests.UI.Shortcuts/Controllers/SqlQueryMonitoringScenarioController.cs
```diff
@@ -0,0 +1,109 @@
...
+
+    /// <summary>
+    /// Returns a response without executing SQL, used to verify that noisy non-SQL requests do not evict actionable
+    /// SQL summaries from the monitoring store.
+    /// </summary>
+    [ResponseCache(Duration = 0, Location = ResponseCacheLocation.None, NoStore = true)]
+    public IActionResult NoSql() => Ok("No SQL executed.");
+
+    /// <summary>
+    /// Executes a raw SQL read query for source coverage tests.
```
Piedone: What "source coverage tests"?

Lombiq.Tests.UI.Tests.UI/TestCases/SqlQueryMonitoringTestCases.cs
Piedone, line 1: Multiple tests cases here don't assert on anything.

Lombiq.Tests.UI/SqlQueryMonitoring/Services/SqlQueryMonitoringConfiguration.cs
```diff
@@ -0,0 +1,418 @@
...
+
+        return entry => !regexes.Any(regex => regex.IsMatch(entry.CommandText));
+    }
+
+    public static readonly Predicate<UITestContext> EnableOnValidatablePagesSqlQueryMonitoringAndAssertionOnPageChangeRule =
+        UrlCheckHelper.IsValidatablePage;
+
+    private Task AssertSqlQueryMonitoringSummaryAgainstThresholdsAsync(SqlQueryMonitoringSummary summary)
+    {
+        if (summary == null) throw new InvalidOperationException("SQL query monitoring summary was not available.");
```
Piedone:
```suggestion
        ArgumentNullException.ThrowIfNull(summary);
```

Lombiq.Tests.UI/SqlQueryMonitoring/Services/SqlQueryMonitoringConfiguration.cs
```diff
@@ -0,0 +1,418 @@
...
+            hasDuplicateCommandFailures,
+            hasDuplicateCommandWithParametersFailures,
+            hasResultSetFailures);
+
+        failures.ShouldBeEmpty(failureMessage);
+
+        return Task.CompletedTask;
+    }
+
+    public void WriteSqlQueryMonitoringCounters(ITestOutputHelper testOutputHelper, SqlQueryMonitoringSummary summary)
```
Piedone: This looks out of place here, since it's not about configuration, more like a helper. Same for `BuildIgnoreCommandTextPatternFilter`.

Lombiq.Tests.UI.Tests.UI/TestCases/SqlQueryMonitoringTestCases.cs
```diff
@@ -0,0 +1,730 @@
...
+    private static async Task AssertExceptionIsThrownAsync<TException>(
+        Func<Task> assertionAsync,
+        Action<TException> assertException,
+        string failureMessage)
+        where TException : Exception
+    {
+        try
+        {
+            await assertionAsync();
+            throw new ConfigurationErrorsException(failureMessage);
```
Piedone: `ConfigurationErrorsException` isn't really applicable here, since it's for the ASP.NET Core configuration provider. Also, isn't `Should.Throw` suitable instead?

Lombiq.Tests.UI/SqlQueryMonitoring/Services/SqlQueryMonitoringConfiguration.cs
```diff
@@ -0,0 +1,418 @@
...
+    public sealed record SqlQueryMonitoringThresholds(
+        int? DuplicateCommandThreshold,
+        int? DuplicateCommandWithParametersThreshold,
+        int? ResultSetRowCountThreshold);
+
+    /// <summary>
+    /// Gets or sets a value indicating whether to automatically run SQL query monitoring assertions every time a page
+    /// changes (either due to explicit navigation or clicks). Disabled by default.
+    /// </summary>
+    public bool RunSqlQueryMonitoringAssertionOnAllPageChanges { get; set; }
```
Piedone, line 32: As a test, to make sure it doesn't interfere with anything, run all tests with this enabled and with large thresholds.

Piedone (changes requested):
Nice work!

Lombiq.Tests.UI/Docs/SqlQueryMonitoring.md
wAsnk: Typical wording of codex, doesn't mean much.

Piedone: You need to clean these up.

wAsnk: Yup, on it.

Lombiq.Tests.UI/SqlQueryMonitoring/SqlQueryMonitoringStore.cs
wAsnk: Yeah, so.... First things first I changed the data structure to a list instead of a queue, because then it is much easier to trim it and do other actions without iterating through the queue all the time. The initial idea was a queue because of how requests should be ordered, like a queue. 

The lock would be needed even if we would use some sort of thread safe collection, because of the other logic around them. Like here, the trimming or in TryRemoveMostRecentCore.

Lombiq.Tests.UI/SqlQueryMonitoring/SqlQueryMonitoringStore.cs
wAsnk: You are right, because of this I started to think about other data structures, and I've decided to use a list instead, because it fits the current logic better.

Lombiq.Tests.UI.TestingModule/Controllers/SqlQueryMonitoringScenarioController.cs
wAsnk: Hmmm, I don't know where to put it so that it will also work.

Lombiq.Tests.UI/SqlQueryMonitoring/Services/SqlQueryMonitoringConfiguration.cs
wAsnk: Yup, I did that previously. And now again without an issue.

Lombiq.Tests.UI.TestingModule/Controllers/SqlQueryMonitoringScenarioController.cs
Piedone: You can create a test-only module project.

Lombiq.Tests.UI/SqlQueryMonitoring/SqlQueryMonitoringStore.cs
Piedone: What other logic, you mean? With a `ConcurrentQueue` you can simply `Dequeue()` until it's small enough.

wAsnk: But the logic that checks what needs to be removed must be locked.

Piedone: What logic are you referring to, isn't it only a simple deque? See https://github.com/Lombiq/UI-Testing-Toolbox/pull/696#discussion_r2902464546.

wAsnk: So after our meeting I checked again the "simple deque" below. [Which is this now.](https://github.com/Lombiq/UI-Testing-Toolbox/pull/696/changes#diff-f9ac1e39d6d6fe63db4b00ec8a5437170c7025490f3aad518f53950c71e40d1aR26) 

The lock cannot be removed here because it's not just a dequeue, we remove only the predicate matched summary.

Piedone: Yeah but as we also discussed, the filtering out the empty summaries is unnecessary and also confusing.

wAsnk: Yeah I removed that so it's now just adding the summary. I'm talking about when we get the summary we need.

Lombiq.Tests.UI.Samples/Tests/SqlQueryMonitoringTests.cs
Piedone, line 1: Insert this into the walkthrough, look for "NEXT STATION".

Lombiq.Tests.UI.Samples/Tests/SqlQueryMonitoringTests.cs
```diff
@@ -0,0 +1,225 @@
...
+// - Filtering out known noisy SQL commands that you explicitly don't care about.
+public class SqlQueryMonitoringTests : UITestBase
+{
+    public SqlQueryMonitoringTests(ITestOutputHelper testOutputHelper)
+        : base(testOutputHelper)
+    {
+    }
+
+    [Fact]
+    public Task SqlQueryMonitoringShouldWork()
```
Piedone: This looks more like four, gradually more complex/less frequent use cases:

1. Explicit assertions.
2. Automatic assertions on page change but filtered to specific URLs, with `ExecutionFilter`, and a custom `BeforeNavigation` handler.
3. What comes after `BeforeNavigation -= perPageThresholdsBeforeNavigation` and before async.
4. Async requests.

Because this is hard to follow all at once.

So, better to have these in separate tests. You don't need to focus on DRY-ing these, like every test can set `EnableSqlQueryMonitoringCollection = true`, because it's easier to understand that way.

Lombiq.Tests.UI/SqlQueryMonitoring/SqlQueryMonitoringSummary.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    /// <summary>
+    /// Gets the request trace identifier associated with this summary.
+    /// </summary>
+    public string TraceIdentifier { get; }
+
+    /// <summary>
+    /// Gets when summary capture finished.
+    /// </summary>
+    public DateTimeOffset CompletedUtc { get; }
```
Piedone: Since this is already called UTC, why not use `DateTime`?

Lombiq.Tests.UI/Docs/SqlQueryMonitoring.md
```diff
@@ -1,165 +1,56 @@
...
+| `ResultSetRowCountThreshold` | `200` | Fails if a single SQL command returns more than 200 rows. |
+| `SummaryLookupTimeout` | `00:00:02` | How long assertions wait for the expected summary to show up. |
+| `SummaryLookupInterval` | `00:00:00.100` | How often assertions poll while waiting for a summary. |
+| `FollowUpSummaryQuietPeriod` | `00:00:00.300` | How long follow-up assertions keep waiting after the last newly captured summary. |
 
 ## Scenario Catalog
 
-For complete, runnable scenarios, see [SQL query monitoring scenario catalog.](../../../Lombiq.OSOCE.Tests.UI/Tests/SqlMonitoringTests/README.md)
+For complete scenarios, see
+[`test/Lombiq.OSOCE.Tests.UI/Tests/SqlMonitoringTests/Readme.md`](../../../Lombiq.OSOCE.Tests.UI/Tests/SqlMonitoringTests/Readme.md).
```
Piedone: This won't work when viewed on GitHub.

Lombiq.Tests.UI/SqlQueryMonitoring/Extensions/SqlQueryMonitoringUITestContextExtensions.cs
```diff
@@ -0,0 +1,367 @@
...
+
+        var assertTask = (assertSummaryAsync ?? configuration.AssertSqlQueryMonitoringSummaryAsync)?
+            .Invoke(summary);
+        return assertTask ?? Task.CompletedTask;
+    }
+
+    private static async Task<SqlQueryMonitoringSummary> GetCurrentRequestSqlQueryMonitoringSummaryAsync(this UITestContext context)
+    {
+        var currentUri = context.GetCurrentUri();
+        var expectedTenantName = NormalizeTenantName(context.TenantName);
```
Piedone: `SqlQueryMonitoringStore` is registered as a singleton for each tenant, not for the whole app, so it'll contain summaries only for a given tenant. So, while in these extension methods it's important to know the tenant, since you need to access its scope, down the call stack there's no need for filtering, nor to keep track of the `TenantName` in `SqlQueryMonitoringSummary`.

Lombiq.Tests.UI/SqlQueryMonitoring/SqlQueryMonitoringStore.cs
Piedone, line 1: There's still a pattern of copying the whole collection twice for every retrieval, which isn't ideal.

Also, `AddSummary` needs to be lock-free, because it's called the most. However, having locking only in `TryRemoveMostRecentMatching` isn't actually reliable, since while it locks two assertions (that shouldn't really be parallel anyway) it lets `AddSummary` add new items (like from async requests), making the copied new `_summaries` potentially incomplete.

Can you think of something better?

Since I see that there's still a need for filtering, I'd suggest a `ConcurrentBag`. It's not ordered, but you don't need it anyway: in `TryRemoveMostRecentMatching`, filter with the predicate, then `OrderByDescending` on `CompletedUtc` (and then `TryTake`). No locking needed anywhere.

Lombiq.Tests.UI.TestingModule/Manifest.cs
```diff
@@ -0,0 +1,16 @@
+using OrchardCore.Modules.Manifest;
+
+[assembly: Module(
+    Name = "TestingModule - Lombiq UI Testing Toolbox",
```
Piedone:
```suggestion
    Name =  "Lombiq UI Testing Toolbox - Testing Module",
```

Lombiq.Tests.UI.TestingModule/Manifest.cs
```diff
@@ -0,0 +1,16 @@
...
+    Author = "Lombiq Technologies",
+    Website = "https://github.com/Lombiq/UI-Testing-Toolbox",
+    Version = "0.0.1"
+)]
+
+[assembly: Feature(
+    Id = "Lombiq.Tests.UI.TestingModule",
+    Name = "Lombiq UI Testing Toolbox - TestingModules",
```
Piedone:
```suggestion
    Name = "Lombiq UI Testing Toolbox - Testing Module",
```

Lombiq.Tests.UI.TestingModule/Manifest.cs
```diff
@@ -0,0 +1,16 @@
...
+    Category = "Development",
+    Description = "WARNING: Only enable this feature in the UI testing environment. Provides shortcuts for common " +
+        "operations that UI tests might want to do or check."
```
Piedone: Leftover.

Lombiq.Tests.UI.Tests.UI/Lombiq.Tests.UI.Tests.UI.csproj
```diff
@@ -18,6 +18,8 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\..\src\Libraries\Lombiq.HelpfulLibraries\Lombiq.HelpfulLibraries.Samples\Lombiq.HelpfulLibraries.Samples.csproj" />
```
Piedone, line 21: Why is this necessary?

Lombiq.Tests.UI.TestingModule/Lombiq.Tests.UI.TestingModule.csproj
```diff
@@ -0,0 +1,27 @@
...
+    <PackageReference Include="OrchardCore.Abstractions" Version="2.1.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(NuGetBuild)' != 'true'">
+    <ProjectReference Include="..\..\..\src\Libraries\Lombiq.HelpfulLibraries\Lombiq.HelpfulLibraries.OrchardCore\Lombiq.HelpfulLibraries.OrchardCore.csproj" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(NuGetBuild)' == 'true'">
+    <PackageReference Include="Lombiq.HelpfulLibraries.OrchardCore" Version="12.5.0" />
+  </ItemGroup>
```
Piedone: This isn't shipped to NuGet, no?

Lombiq.Tests.UI/Extensions/OrchardCoreBuilderExtensions.cs
```diff
@@ -47,6 +49,7 @@ public static OrchardCoreBuilder ConfigureUITesting(
                 services.AddSingleton<ISqlQueryMonitoringStore, SqlQueryMonitoringStore>();
                 services.AddScoped<ISqlQueryMonitoringContext, SqlQueryMonitoringContext>();
                 services.AddSingleton<IStartup, SqlQueryMonitoringStartup>();
+                services.AddTransient(_ => new ShellFeature("Lombiq.Tests.UI.TestingModule", alwaysEnabled: true));
```
Piedone: You can use `builder.AddTenantFeatures()` to enable the feature in a standard way.

Piedone (changes requested):

Lombiq.Tests.UI/SqlQueryMonitoring/SqlQueryMonitoringStore.cs
wAsnk: You are right, this is currently not entirely perfect.

If we don't care about removing items from the  `ConcurrentBag` that's fine, otherwise `ConcurrentBag` can't be used.

If you try to filter and order the `ConcurrentBag` it will be an ordered enumerable, which you can use the summary from, but you can't remove that specific summary from the `ConcurrentBag` with `TryTake`.

Lombiq.Tests.UI.Tests.UI/Lombiq.Tests.UI.Tests.UI.csproj
wAsnk: Because of `LinqToDbSamplesController` used in `LinqToDbSamplesShouldBeCapturedBySqlMonitoringAsync`.

Lombiq.Tests.UI/Extensions/OrchardCoreBuilderExtensions.cs
wAsnk: We can only use it after the `builder.ConfigureServices` which we are inside of. It would look like this, which I think is meh, because of the double `if` for the same boolean and the variable moved outside of `ConfigureServives`. What do you think?
image

Lombiq.Tests.UI/SqlQueryMonitoring/SqlQueryMonitoringStore.cs
Piedone: Ah right, I thought you can give `TryTake` an object to remove, but no. It's useless in this case.

Is removal here needed for any logic (i.e. not memory saving) reason here? Because if not, then yes, let's use `ConcurrentBag`.

Lombiq.Tests.UI/Extensions/OrchardCoreBuilderExtensions.cs
Piedone: That's fine. Less than ideal but better than a custom feature registration.

Lombiq.Tests.UI/SqlQueryMonitoring/SqlQueryMonitoringStore.cs
wAsnk: We were dependent on removing the summaries for async requests, but I changed the logic for that as well.

Piedone (approved):

PR #770, opened 2026-06-20
OSOE-1286: Updating Atata packages and recreating the now unmaintained AxeHtmlReport

Piedone:
[OSOE-1286](https://lombiq.atlassian.net/browse/OSOE-1286)

[OSOE-1286]: https://lombiq.atlassian.net/browse/OSOE-1286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/AccessibilityChecking/AxeHtmlReport.cs
sarahelsaig, line 1: Is this meant to be just a patched copy or a divergent version?

If this is a patch, please copy the "Taken from https://github.com/TroyWalshProf/SeleniumAxeHtmlDotnet" as a regular comment at the top of the file, because currently it's kind of hidden in the middle of the `AxeHtmlReport` documentation xml. Also clearly indicate with comments or regions which lines changed meaningfully, because "and updated" will not be helpful if we ever have to update this from the original.

If this is a divergent report file that doesn't have to be kept in sync with the original then I'd mention that explicitly. Also please extract the CSS and JS code at the bottom into their own standalone files and reference them as `<EmbeddedResource>`.

sarahelsaig (changes requested):

sarahelsaig (approved):

PR #771, opened 2026-06-22
Introducing custom BrowserLogEntry type to prevent future breaking changes with Selenium BiDi types (⚠️ breaking changes)

Piedone:
[OSOE-1286](https://lombiq.atlassian.net/browse/OSOE-1286)

[OSOE-1286]: https://lombiq.atlassian.net/browse/OSOE-1286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Tests.UI/Models/BrowserLogEntry.cs
```diff
@@ -0,0 +1,55 @@
...
+    public StackTrace StackTrace { get; set; }
+
+    internal BrowserLogEntry(LogEntry entry)
+    {
+        Level = entry.Level;
+        Source = entry.Source;
+        Text = entry.Text;
+        Timestamp = entry.Timestamp;
+        StackTrace = entry.StackTrace;
+    }
```
sarahelsaig: Is there a reason why this isn't `record`? It's cleaner, works with serialization out of the box, and the auto generated `Equals` and `Deconstruct` could come in handy in custom assertions.

```suggestion
public record BrowserLogEntry(
    Level Level,
    Source Source,
    string Text,
    DateTimeOffset Timestamp,
    StackTrace StackTrace)
{
    internal BrowserLogEntry(LogEntry entry)
        : this(entry.Level, entry.Source, entry.Text, entry.Timestamp, entry.StackTrace)
    {
    }
```

Lombiq.Tests.UI/Models/BrowserLogEntry.cs
```diff
@@ -0,0 +1,55 @@
...
+        OrchardCoreUITestExecutorConfiguration.IsNonSuccessBrowserLogEntry(this);
+
+    private static string FormatStackTrace(StackTrace stackTrace)
+    {
+        if (stackTrace == null) return string.Empty;
+
+        return
+            Environment.NewLine +
+            "Stack trace: " +
+            Environment.NewLine +
```
sarahelsaig: Unnecessary space-newline combo.
```suggestion
            "Stack trace:" +
            Environment.NewLine +
```

Lombiq.Tests.UI/Models/BrowserLogEntry.cs
```diff
@@ -0,0 +1,55 @@
...
+            string.Join(
+                Environment.NewLine,
+                stackTrace.CallFrames.Select(frame =>
+                    "- " +
+                    (string.IsNullOrEmpty(frame.FunctionName) ? string.Empty : frame.FunctionName + " at ") +
```
sarahelsaig: Unparenthesized operators inside a ternary look creepy, better to use string interpolation.
```suggestion
                    (string.IsNullOrEmpty(frame.FunctionName) ? string.Empty : $"{frame.FunctionName} at ") +
```

sarahelsaig (changes requested):

Lombiq.Tests.UI/Models/BrowserLogEntry.cs
Piedone: Nope, just never occurred to me, thanks!

Lombiq.Tests.UI/Models/BrowserLogEntry.cs
Piedone: Image

PR #816, opened 2026-08-28
NEST-608: Disable attribute-empty-style and update html-validate NPM package

sarahelsaig:
because it will never work. When passing the page source to html-validate using `await new HtmlValidator(options).ValidateAsync(context.Driver.PageSource)` we will find that Chrome (also Firefox) converts the raw HTML into `<div hidden>` into `<div hidden=\"\"></div>` which violates the rule with that empty value. This can be confirmed in the browser's developer console by typing `document.body.innerHTML = '<div hidden>'; alert(document.body.innerHTML)`.

[NEST-608](https://lombiq.atlassian.net/browse/NEST-608)

[NEST-608]: https://lombiq.atlassian.net/browse/NEST-608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI.Shortcuts/Lombiq.Tests.UI.Shortcuts.csproj
+++ b/Lombiq.Tests.UI.Shortcuts/Lombiq.Tests.UI.Shortcuts.csproj
@@ -31,12 +31,6 @@
     <PackageReference Include="OrchardCore.Users" Version="$(OrchardCoreVersion)" />
   </ItemGroup>
 
-  <ItemGroup>
-    <EmbeddedResource Update="Recipes\Lombiq.Tenants.Core.Features.recipe.json">
-      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
-    </EmbeddedResource>
-  </ItemGroup>
-
   <Import Condition="'$(NuGetBuild)' != 'true'" Project="../../../src/Utilities/Lombiq.MSBuild.Targets/Lombiq.MSBuild.OrchardCore.Module.Sdk/Sdk/Import.targets" />
   <Import Condition="'$(NuGetBuild)' == 'true'" Project="Sdk.targets" Sdk="Lombiq.MSBuild.OrchardCore.Module.Sdk" Version="2.0.1" />
 </Project>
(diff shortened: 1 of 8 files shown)


Piedone (approved):

Lombiq.Tests.UI/Services/UITestExecutionSession.cs
```diff
@@ -789,6 +799,20 @@ private string GetSetupHashCode() =>
...
+        var isCloudflareErrorPage = context.ExecuteScript(CloudflareErrorPageDetectionJavaScript);
+        if (isCloudflareErrorPage is true)
+        {
+            throw new PageChangeAssertionException(
+                $"Cloudflare error page \"{context.Driver.Title}\" encountered. Check the \"PageSource.mhtml\" file " +
+                "in the test dump for more details.");
+        }
+
+        return Task.CompletedTask;
+    }
```
Piedone: Since this is remote test-specific for CF, it should rather be in `CloudflareRemoteUITestBase`.

Lombiq.Tests.UI.Shortcuts/Lombiq.Tests.UI.Shortcuts.csproj
```diff
@@ -31,6 +31,12 @@
     <PackageReference Include="OrchardCore.Users" Version="$(OrchardCoreVersion)" />
   </ItemGroup>
 
+  <ItemGroup>
+    <EmbeddedResource Update="Recipes\Lombiq.Tenants.Core.Features.recipe.json">
+      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
+    </EmbeddedResource>
+  </ItemGroup>
```
Piedone: Is this actually needed, also for proper NuGet publishing? E.g. `Lombiq.Walkthroughs` and `Lombiq.UIKit.Widgets` don't have this for their recipes.

Piedone (changes requested):

Piedone (approved):

PR #828, opened 2026-09-12
NEST-671: Added the ability to carry the By into ClickReliablyAsync and refresh the element

sarahelsaig:
[NEST-671](https://lombiq.atlassian.net/browse/NEST-671)

[NEST-671]: https://lombiq.atlassian.net/browse/NEST-671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
@@ -492,7 +492,7 @@ public static class NavigationUITestContextExtensions
     /// </summary>
     /// <param name="maxTries">The maximum number of clicks attempted altogether, if retries are needed.</param>
     public static Task ClickReliablyOnAsync(this UITestContext context, By by, int maxTries = 3) =>
-        context.Get(by).ClickReliablyAsync(context, maxTries);
+        context.Get(by).ClickReliablyAsync(context, by, maxTries);
 
     /// <summary>
     /// Reliably clicks on the link identified by the given text with <see
@@ -500,7 +500,7 @@ public static class NavigationUITestContextExtensions
     /// </summary>
     /// <param name="maxTries">The maximum number of clicks attempted altogether, if retries are needed.</param>
     public static Task ClickReliablyOnByLinkTextAsync(this UITestContext context, string linkText, int maxTries = 3) =>
-        context.Get(By.LinkText(linkText)).ClickReliablyAsync(context, maxTries);
+        context.ClickReliablyOnAsync(By.LinkText(linkText));
 
     /// <inheritdoc cref="ClickReliablyOnUntilNavigationHasOccurredAsync(UITestContext, By, TimeSpan?, TimeSpan?)"/>
     [Obsolete($"Use {nameof(ClickReliablyOnUntilNavigationHasOccurredAsync)} instead.")]
--- a/Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
@@ -552,7 +552,20 @@ public static class NavigationUITestContextExtensions
         TimeSpan? interval = null)
     {
         var currentUrl = context.Driver.Url;
-        await context.ClickReliablyOnAsync(by);
+
+        // If selected HTML element is <a> and it points to the current page, then this method will time out. To get
+        // around this problem, we modify the current URL using JavaScript.
+        var element = context.Get(by);
+        if (element.TagName == TagNames.A)
+        {
+            context.ExecuteScript(
+                "if (arguments[0].href === location.href) window.history.replaceState(null, '', `${location.href}#ts=${Date.now()}`);",
+                element);
+            currentUrl = context.Driver.Url;
+        }
+
+        await element.ClickReliablyAsync(context, by);
+
         ReliabilityHelper.DoWithRetriesOrFail(
             () => context.Driver.Url != currentUrl,
             timeout,
--- a/Lombiq.Tests.UI/Extensions/NavigationWebElementExtensions.cs
+++ b/Lombiq.Tests.UI/Extensions/NavigationWebElementExtensions.cs
@@ -1,4 +1,5 @@
 using Atata;
+using Lombiq.Tests.UI.Models;
 using Lombiq.Tests.UI.Services;
 using OpenQA.Selenium;
 using System;
@@ -9,6 +10,9 @@ namespace Lombiq.Tests.UI.Extensions;
 
 public static class NavigationWebElementExtensions
 {
+    public static Task ClickReliablyAsync(this IWebElement element, UITestContext context, int maxTries = 3) =>
+        element.ClickReliablyAsync(context, originalSelector: null, maxTries);
+
     /// <summary>
     /// Clicks an element even if the default Click() will sometimes fail to do so. It's more reliable than Click() but
     /// still not perfect. If you're doing a Get() before then use <see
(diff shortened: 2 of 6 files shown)


Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -552,7 +552,20 @@ public static async Task ClickReliablyOnAndWaitUntilUrlChangeAsync(
...
+        var element = context.Get(by);
+        if (element.TagName == TagNames.A)
+        {
+            context.ExecuteScript(
+                "if (arguments[0].href === location.href) window.history.replaceState(null, '', `${location.href}#ts=${Date.now()}`);",
+                element);
+            currentUrl = context.Driver.Url;
+        }
+
+        await element.ClickReliablyAsync(context, by);
```
sarahelsaig: This is the main fix to the test errors we've seen in prod. I use JS to [change the current URL](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState) without reloading the page, before clicking on the link. This way there is URL change even when you click on a link that points to the current page and don't have to rely on the URL's trailing `/`.

Lombiq.Tests.UI/Extensions/NavigationWebElementExtensions.cs
```diff
@@ -23,7 +27,7 @@ public static class NavigationWebElementExtensions
     /// </para>
     /// </remarks>
     /// <param name="maxTries">The maximum number of clicks attempted altogether, if retries are needed.</param>
-    public static Task ClickReliablyAsync(this IWebElement element, UITestContext context, int maxTries = 3) =>
+    public static Task ClickReliablyAsync(this IWebElement element, UITestContext context, By originalSelector, int maxTries = 3) =>
```
sarahelsaig, line 37: This change allows us to automatically re-query the element again if it goes stale, making the method much more reliable everywhere.

Lombiq.Tests.UI/Extensions/NavigationWebElementExtensions.cs
```diff
@@ -85,6 +63,50 @@ await context.Configuration.Events.AfterClick
...
+                return errorContext;
+
+            case WebDriverException webDriverException when webDriverException.IsStaleElementLikeException():
+                context.Configuration.TestOutputHelper.WriteLineTimestampedAndDebug(
+                    "Stale element exception with the message \"{0}\", retrying the click.",
+                    exception.Message);
+
+                return errorContext.OriginalSelector is { } by && context.Get(by.Safely()) is { } element
+                        ? errorContext with { Element = element }
+                        : errorContext;
```
sarahelsaig, line 99: This is the only substantive change, for this `switch` block. We just try to get a fresh version of the element if it's stale. This fixes some flakiness I've experienced, unrelated to the prod test failures.

The rest of this method is just refactoring. Originally I did this check with an `if` so the code analyzer complained that there are too many nested controls. So I had to extract this into a method. I kept the change because I think it's more readable this way. But we can revert it if you think it's unnecessary.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
Piedone: This feels a bit hackish. Instead of running JS to trick the logic below, why not change the logic? Like if it's an `a` that points to the current change, fall back to use `DoWithRetriesUntilNavigationHasOccurredOrFailAsync` instead of checking the URL.

Lombiq.Tests.UI/Extensions/NavigationWebElementExtensions.cs
```diff
@@ -9,6 +10,9 @@ namespace Lombiq.Tests.UI.Extensions;
 
 public static class NavigationWebElementExtensions
 {
+    public static Task ClickReliablyAsync(this IWebElement element, UITestContext context, int maxTries = 3) =>
```
Piedone: Add docs.

Lombiq.Tests.UI/Extensions/NavigationWebElementExtensions.cs
Piedone: Add docs about this.

Lombiq.Tests.UI/Extensions/NavigationWebElementExtensions.cs
Piedone: Looks good.

Lombiq.Tests.UI/Models/ClickReliablyErrorContext.cs
```diff
@@ -0,0 +1,14 @@
+#nullable enable
+
+using Lombiq.Tests.UI.Services;
+using OpenQA.Selenium;
+using System;
+
+namespace Lombiq.Tests.UI.Models;
+
+public record ClickReliablyErrorContext(
```
Piedone: Or it can be even a nested type in the extensions class.

```suggestion
internal record ClickReliablyErrorContext(
```

Piedone (changes requested):
Appreciate the comments!

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
sarahelsaig: The `<a>` elements are precisely where we want to avoid `DoWithRetriesUntilNavigationHasOccurredOrFailAsync`, because it tries to click on the link multiple times until the navigation occurs. Selenium under Chrome has this debounced under the hood, but relying on that is even more of a hack.

Piedone: That method is only a retry logic. I didn't mean we should repeatedly click, rather, use that, to check for the navigation state having changed (like this method checks for the URL below).

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
```diff
@@ -551,13 +551,32 @@ public static async Task ClickReliablyOnAndWaitUntilUrlChangeAsync(
...
+        }
+
+        if (isElementLinkToCurrentPage)
+        {
+            await context.DoWithRetriesUntilNavigationHasOccurredOrFailAsync(ProcessAsync, timeout, interval);
+        }
+        else
+        {
+            await context.DoWithRetriesUntilUrlChangeOrFailAsync(ProcessAsync, timeout, interval);
+        }
```
Piedone:
```suggestion
        await element.ClickReliablyAsync(context, by);

        if (isElementLinkToCurrentPage)
        {
            await context.DoWithRetriesUntilNavigationHasOccurredOrFailAsync(() => Task.CompletedTask, timeout, interval);
        }
        else
        {
            await context.DoWithRetriesUntilUrlChangeOrFailAsync(() => Task.CompletedTask, timeout, interval);
        }
```

Piedone (approved):

Piedone (changes requested):
Or rather, one proposed change.

Lombiq.Tests.UI/Extensions/NavigationUITestContextExtensions.cs
sarahelsaig: That's not good. You have to click _inside_ the reliability helper. It first stores the URL or navigation state. Then clicks. Then keeps checking for a change against the state it stored when you invoked the method. If you click first, it will think the already-altered state is the initial and you get a timeout error.

The `ProcessAsync` ensures that you only click once inside the polling loop of `DoWithRetries*`. It's a necessary complication.

sarahelsaig: I've renamed `ProcessAsync` to `ClickOnlyOnceAsync` and elaborated the comment above, to explain why this function is necessary.

Piedone (approved):

